PR #18, opened 2023-08-14
Big Pipeline Update

MatKuhr:
### Context Information

BLI: SAP/cloud-sdk-java-backlog#121

This PR does a couple of things:

- [x] Add OpenAPI modules
- [x] ~Change the pipeline to be a single job (to enable PMD + Spotbugs and improve performance)~ 
  - [x] Implement multi-stage parallel pipeline for better performance
  - [x] Add an equivalent, simplified i.e. linear version of the pipeline as daily cron job to serve as a sanity check 
- [x] Actually make use of `modules-bom`
- [x] Upgrade to version `4.20.0`
- [x] Clean up some unnecessary stuff from the parent pom
  - [x] This has been [done on the internal repo](https://github.wdf.sap.corp/MA/sdk/pull/8441) and copied into this PR to minimal less divergence to internal repo
- [x] Add file change detection
- [x] Fix PMD scripts to work with suppressions
- [x] Enhance scripts to allow for unlimited findings  
- [x] Merge latest state from main/v5


.github/workflows/build.yaml
```diff
@@ -12,114 +12,57 @@ env:
...
+        run: "mvn $MVN_CLI_ARGS com.github.ekryd.sortpom:sortpom-maven-plugin:verify"
+      # Check style is currently lacking the required file, so disable it for now
+      # - name: Check Code Style
+      #   run: |
+      #     mvn $MVN_CLI_ARGS org.apache.maven.plugins:maven-checkstyle-plugin:checkstyle
+      #     python .pipeline/scripts/print-checkstyle.py
+      - name: Compile
+        # we need to use install here because otherwise the code checks below don't work
+        # they seem to require "mvn dependency:resolve" to work, why is unclear honestly...
+        # Problem: With caching the maven repo this might? maybe? produce race conditions / side effects between different executions
```
Johannes-Schneider, line 51: I'm also worried about this... 

As by the documentation:
> The format of the used cache key is `setup-java-${{ platform }}-${{ packageManager }}-${{ fileHash }}`, where the hash is based on the following files [...]
> * maven: **/pom.xml

Therefore, it seems like the cache key would always be the exact same for different execution instances (which is bad) 😔

I'm wondering about the entire setup in general: The path `$HOME/.m2/repository` doesn't seem to be different for different runs... So we would always share the same `.m2`.
Therefore, I feel like we might need to come up with our own "caching" strategy to avoid leaking state from one job run into an other one?

.pipeline/scripts/print-spotbugs.py
```diff
@@ -55,5 +56,5 @@
 print(f"warnings normal: {findings['2']}, allowed is {threshold_normal}")
 print(f"warnings high:   {findings['1']}, allowed is {threshold_high}")
 
-if findings['3'] > threshold_low or findings['2'] > threshold_normal or findings['1'] > threshold_high:
+if findings['1'] > threshold_high: # or findings['2'] > threshold_normal or findings['3'] > threshold_low:
```
Johannes-Schneider, line 59: I assume you commented the rest of the condition for testing, right?
If so, we should probably revert this before merging.

pom.xml
```diff
@@ -503,7 +455,7 @@
 				<groupId>com.sap.cloud.sdk.testutil</groupId>
 				<artifactId>testutil-core</artifactId>
 				<version>${sdk.version}</version>
-				<scope>test</scope>
+				<scope>compile</scope>
```
Johannes-Schneider, line 458: Why do we need this in `compile` scope?

Johannes-Schneider (commented):
Just from the looks of this PR, it seems like you spent significant manual effort for getting the pipeline and entire maven setup to somehow work...
Additionally, there seems to be a lot of clean up contained in the added modules.

IMO, this is not a good approach.
If we start investing so much clean up / setup effort for just a few modules, merging the rest of the code base in here is going to be a nightmare.
So I'd suggest to stick with the approach we aligned on:

* Remove all the internal modules in our existing repo first
* (We can still update the pipeline as suggested in this PR)
* Once that is done, simply force push everything into this repo
* THEN start cleaning up and fixing things
* (Crossing all fingers that we don't have to migrate changes coming from `v4`)

.github/workflows/build.yaml
MatKuhr: yeah, this is [first in the ToDo list](https://github.com/SAP/cloud-sdk-java-backlog/issues/121#issuecomment-1678577931) to come up with a solution for. 

For now it's kinda fine, because the cache doesn't properly work anyway. The cache key is indeed the same, as I could tell from the logs, but the content is still lost between jobs (even in the same workflow). Part of the reason why I had to merge everything into one job...

.pipeline/scripts/print-spotbugs.py
MatKuhr: no, we have around 80 normal findings which we would all have to fix or increase the threshold otherwise. to me it looked like minor stuff not worth failing over.

pom.xml
MatKuhr: leftover, fixed

.github/workflows/build.yaml
```diff
@@ -2,123 +2,172 @@ name: CI
...
+          path: ~/.m2/repository
+          key: maven
+      - name: Compile
+        # we need to use install here because otherwise the code checks below don't work
+        # they require the JARs of all modules to work, so those need to be installed
+        run: |
+          mvn $MVN_CLI_ARGS install -DskipTests -Dskip-formatting
+      - name: Local Changes Check
+        run: |
+          CHANGED_FILES="$(git status --porcelain | rev | cut -f1 -d " " | rev)"
```
Johannes-Schneider, line 82: 😉

```suggestion
          CHANGED_FILES="$(git diff --name-only)"
```

.github/workflows/build.yaml
```diff
@@ -2,123 +2,172 @@ name: CI
...
+          if [ ! -z $CHANGED_FILES ]; then
+              echo "There are local changes in the following files:"
+              echo $CHANGED_FILES
+              echo "Printing the git diff:"
+              git --no-pager diff
```
Johannes-Schneider, line 87: Ufff... this might be getting HUGE if we decided to change some service specification (in one of the sample modules, for example), right.
So it this really a good idea? 😅

.github/workflows/build.yaml
```diff
@@ -2,123 +2,172 @@ name: CI
...
         uses: actions/upload-artifact@v3
         with:
-          name: compiled-classes-jdk-${{ matrix.Java }}
-          path: |
-            **/target/**/*.class
+          name: m2-sdk
+          path: ~/.m2/repository/com/sap/cloud/sdk/m2-sdk.zip
           retention-days: 1
-      - name: Archive target
+      - name: Purge Maven Repository from SDK Artifacts
```
Johannes-Schneider, line 102: Sounds better...?

```suggestion
      - name: Purge SDK Artifacts from Maven Repository 
```

Johannes-Schneider (commented):
I just looked at the pipeline changes, which looked pretty nice 👍

.github/workflows/build.yaml
```diff
@@ -2,123 +2,172 @@ name: CI
...
-      - name: Publish Test Report
+          name: target-dirs
+          path: .
+      - name: UNZIP Target Dirs
+        run: |
+          unzip target.zip
+
+      - name: Test
+        run: |
+          mvn $MVN_CLI_ARGS test jacoco:report
```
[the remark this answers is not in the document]
Johannes-Schneider: Actually, when checking the output of the [last successful run](https://github.com/SAP/cloud-sdk-java/actions/runs/5889200162/job/15971941582?pr=18) (at this point in time), I can see that the code is indeed still getting compiled:

```
[INFO] --- maven-compiler-plugin:3.11.0:compile (default-compile) @ fluent-result ---
[INFO] Changes detected - recompiling the module! :source
[INFO] Compiling 35 source files with javac [debug deprecation release 8] to target/classes
```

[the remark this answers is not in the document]
Johannes-Schneider: It looks like this can be fixed rather easily by directly invoking the `surefire` plugin (`mvn surefire:test ...`) instead of just `mvn test` - verified manually on my local machine 👍

.github/workflows/build.yaml
MatKuhr: sure, but why not? I don't think you will get flamed for too many log output lines that quickly.. and even if we do, worst case the run is aborted, which is fine since it already failed anyway..

.github/workflows/build.yaml
MatKuhr: Almost, remember `git diff` prints in interactive mode, but adding `--no-pager` does the trick 😉
Unfortunately, I can't give `--porcelain` (which is supposed to make the output more script/CI friendly in the case of git status) here, but I can't think of something we would need it for on the file names.. so should be fine..

.github/workflows/build.yaml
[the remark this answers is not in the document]
MatKuhr: uff, good that you checked that. It was working at some point, I swear 😄 
My best guess is that it stopped working when I switched to the zip file approach. It can also be fixed by simply touching all files after unzipping, which I added. 

However, I now see that `man test` runs a bunch of other plugins again as well eg. the enforcer plugin. Those take some time and already run during install, so I would go with `surefire:test`

Remaining question is: Is there any plugin/goal that we don't execute, if we run `install -DskipTests` and `surefire:test` compared to `install`? 
I think I'm just gonna compare the logs to see if I can find any diff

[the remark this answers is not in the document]
Johannes-Schneider: > Remaining question is: Is there any plugin/goal that we don't execute, if we run `install -DskipTests` and `surefire:test` compared to install?

Good question.
How about being more precise in the compilation step by actually just skipping the `surefire:test` phase?
(I saw that there seems to be an option in the plugin config to skip the execution, so that could be done using our regular property approach - but not sure if really achieves the desired result)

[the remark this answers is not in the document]
Johannes-Schneider: 🚨🚨🚨🚨

It looks like the most recent approach (`mvn jacoco:prepare-agent surefire:test jacoco:report`) produces a different test coverage result compared the "regular" approach (`mvn test jcoco:report`).

I compared [run #168](https://github.com/SAP/cloud-sdk-java/actions/runs/5891683113) vs. [run #161](https://github.com/SAP/cloud-sdk-java/actions/runs/5890550367/job/15975933873):

```
(run 168)
coverage = 0.668060439038297
branch coverage = 0.5957746478873239

 vs. 

(run 161)
coverage = 0.5919857235561323
branch coverage = 0.5228365384615384
```

@MatKuhr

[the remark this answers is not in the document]
MatKuhr: We are already doing that. The source for the surefire option is:
```
<parameter>
          <name>skipTests</name>
          <type>boolean</type>
          <since>2.4</since>
          <required>false</required>
          <editable>true</editable>
          <description>Set this to &quot;true&quot; to skip running tests, but still compile them. Its use is NOT RECOMMENDED, but quite convenient on occasion.
Failsafe plugin deprecated the parameter skipTests and the parameter will be removed in Failsafe 3.0.0 as it is a source of conflicts between Failsafe and Surefire plugin.</description>
</parameter>
```

I was only concerned about other plugins that bind to the `test` maven lifecycle (and may or may not also use the same property as flag). But I checked and other than the `prepare-agent` from Jacoco there is nothing running in addition when comparing `surefire:test` with `test` that didn't already run during `install -DskipTests`, so we should be safe.

And the benefit is quite substantial, as we don't unnecessarily re-run enforcer and dependency plugins.

.github/workflows/build-sanity.yaml
```diff
@@ -0,0 +1,66 @@
...
+          fi
+      - name: Test Report
+        if: success() || failure()
+        uses: scacap/action-surefire-report@v1
+      - name: Generate JaCoCo Badge
+        id: jacoco
+        uses: cicirello/jacoco-badge-generator@v2
+        with:
+          generate-branches-badge: true
+          jacoco-csv-file: "**/jacoco.csv"
```
Johannes-Schneider, line 54: I'd say let's not generate the JaCoCo Badge from this pipeline (we do that from our regular pipeline, which runs on every `main` build anyways, right?).

.github/workflows/build-sanity.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: 17
+      - name: Load Maven Repository
+        uses: actions/cache/restore@v3
+        with:
+          path: ~/.m2/repository
+          key: maven
```
Johannes-Schneider, line 32: As we are aiming for a sanity check, let's remove external factors (such as this) to be extra sure everything is indeed working as intended (maybe we messed up something in the caching in our other pipeline and we don't notice because we made the same mistake here?).

Ultimately, since this runs only once a day, performance doesn't really matter here.

.github/workflows/build.yaml
```diff
@@ -2,123 +2,171 @@ name: CI
 
 on:
   push:
-    branches: ["main", "poc_pipeline"]
+    branches: ["main"]
```
Johannes-Schneider, line 5: [comment - no action needed in this PR]

Do we plan on sticking with our aligned branch naming strategy?
(I.e. rename `main` to `main/v5`?)

.github/workflows/build.yaml
```diff
@@ -2,123 +2,171 @@ name: CI
...
-    strategy:
-      fail-fast: false
-      matrix:
-        # jdk 20 currently does not work because of a warning:
-        # Warning:  source/target value 8 is obsolete and will be removed in a future release
-        java: [ '17', '19' ]
-    name: Build, Test and Verify JDK ${{ matrix.Java }}
+    name: Build
+    env:
+      CHANGED_FILES: ""
```
Johannes-Schneider, line 61: Why is this needed?
This variable should be defined in the `Local Changes Check` no matter what, right?

Johannes-Schneider (commented):
Just a few more (minor) things. Otherwise LGTM

.github/workflows/build-sanity.yaml
MatKuhr: I'm not sure if the badge is used anywhere. Instead, we use the outputs of the step to determine the coverage %, [which we will enforce via thresholds in the future](https://github.com/SAP/cloud-sdk-java-backlog/issues/282). I would leave it to the ticket to figure out the details on this..

.github/workflows/build.yaml
MatKuhr: I was thinking about this too. But with the current timeline I figured we should leave it out for now and just use `main`, until there is a need for a `v6`.

PR #25, opened 2023-08-28
Update Blackduck Action

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/280


.pipeline/config.yml
```diff
@@ -4,6 +4,15 @@ steps:
     groups:
       - 'CLOUD SDK 4.0'
     serverUrl: 'https://sap.blackducksoftware.com/'
+    mavenExcludedScopes: ["provided", "test"]
+    failOn: ['BLOCKER', 'CRITICAL', 'MAJOR']
+    versioningModel: "major"
+    scanProperties:
+      # --detect.maven.excluded.modules=s4hana-all,s4hana-api-odata,s4hana-api-odata-onpremise,s4hana-api-odata-v4,s4hana-api-odata-v4-onpremise,s4hana-api-parent,end-to-end-tests,performance-tests,tests-parent,unit-tests,vdm-integration-tests,testutil-core,testutil-parent
+      - --detect.maven.excluded.modules=tests-parent,testutil-core,testutil-parent
+      # - --detect.maven.build.command="'-pl !:s4hana-all,!:s4hana-api-odata,!:s4hana-api-odata-onpremise,!:s4hana-api-odata-v4,!:s4hana-api-odata-v4-onpremise,!:s4hana-api-parent,!:end-to-end-tests,!:performance-tests,!:tests-parent,!:unit-tests,!:vdm-integration-tests,!:testutil-core,!:testutil-parent'"
```
MatKuhr, line 13: in the internal repo we currently read this from `<x-sap-exclude-from-blackduck-scan>true</x-sap-exclude-from-blackduck-scan>`

I'm not sure it is worthwhile continuing with an automated approach, as we have a lot less modules, and no internal modules at all anymore.

Maybe you can add a note [here](https://github.com/SAP/cloud-sdk-java-backlog/issues/281) to remove the maven properties when copying the sources over?

.pipeline/config.yml
```diff
@@ -4,6 +4,15 @@ steps:
     groups:
       - 'CLOUD SDK 4.0'
     serverUrl: 'https://sap.blackducksoftware.com/'
+    mavenExcludedScopes: ["provided", "test"]
+    failOn: ['BLOCKER', 'CRITICAL', 'MAJOR']
+    versioningModel: "major"
+    scanProperties:
+      # --detect.maven.excluded.modules=s4hana-all,s4hana-api-odata,s4hana-api-odata-onpremise,s4hana-api-odata-v4,s4hana-api-odata-v4-onpremise,s4hana-api-parent,end-to-end-tests,performance-tests,tests-parent,unit-tests,vdm-integration-tests,testutil-core,testutil-parent
```
MatKuhr, line 11: I think we can safely remove the commented out stuff, all of the modules are going to be deleted, except maybe `unit-tests`, not sure on that yet

MatKuhr (approved):
I guess we have to merge this to test it?

.pipeline/config.yml
KevinHudemann: Ok will delete it :)

KevinHudemann:
> I guess we have to merge this to test it?

No there is a workflow dispatch trigger so you can trigger it manually: https://github.com/SAP/cloud-sdk-java/actions/runs/5999268530

There are findings at the moment though which let the run fail. It was quite difficult to judge whether the results are equivalent or would be irrelevant once the repo is complete which is why we only tested functionality and that the config parameter are the same.

.pipeline/config.yml
KevinHudemann: Done

MatKuhr:
> There are findings at the moment though which let the run fail

Ah okay, the scan is currently overriding our scan results from the internal repository. This is somewhat a problem... but also somewhat surprising to me, because with the versioning model I had assumed it would create a new version 0 instead of taking the existing v5...

Regarding the findings, chances are most are false positives coming from the CLI module which we are intending to remove.

KevinHudemann:
> Ah okay, the scan is currently overriding our scan results from the internal repository. This is somewhat a problem... but also somewhat surprising to me, because with the versioning model I had assumed it would create a new version 0 instead of taking the existing v5...

@MatKuhr Should we change it to use the version 0 or should we keep it as v5?

MatKuhr:
> Should we change it to use the version 0 or should we keep it as v5?

I think it's safe to keep at 5 already 👍🏻

MatKuhr (approved):

.pipeline/config.yml
```diff
@@ -4,6 +4,13 @@ steps:
     groups:
       - 'CLOUD SDK 4.0'
     serverUrl: 'https://sap.blackducksoftware.com/'
+    mavenExcludedScopes: ["provided", "test"]
+    failOn: ['BLOCKER', 'CRITICAL', 'MAJOR']
+    versioningModel: "major"
+    scanProperties:
+      - --detect.maven.excluded.modules=tests-parent,testutil-core,testutil-parent,openapi-generator-cli
+      - --detect.maven.build.command="'-pl !:tests-parent,!:testutil-core,!:testutil-parent,!:openapi-generator-cli'"
```
Johannes-Schneider, line 12: FYI (@daniel-kurzynski and @KevinHudemann):

Since we talked about our `module-inventory.json` today, there is also a property `"excludeFromBlackDuckScan": true/false`, which could be automatically read and used here.

KevinHudemann: Yeah that's right, now I unfortunately just merged the PR. I'll create a follow-up PR for that.

Johannes-Schneider: I think it is not super important / pressing, but could still be considered if time allows.

PR #26, opened 2023-08-29
Add step for snapshot deployments

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/252
Build with deployment: https://github.com/SAP/cloud-sdk-java/actions/runs/6011750052/job/16305849354


.github/workflows/build.yaml
```diff
@@ -302,4 +310,61 @@ jobs:
...
+          name: target-dirs
+          path: .
+      - name: UNZIP Target Dirs
+        run: |
+          unzip -D target.zip
+      - name: Publish Snapshot
+        run: >
+          mvn $MVN_CLI_ARGS
+          -DaltSnapshotDeploymentRepository=artifactory-snapshots::https://common.repositories.cloud.sap/artifactory/build-snapshots-cloudsdk
+          -Dmaven.test.skip -Dmaven.compiler.showCompilationChanges -Dhttp.keepAlive=false-Dformatter.skip
```
Johannes-Schneider, line 365: [typo]

```suggestion
          -Dmaven.test.skip -Dmaven.compiler.showCompilationChanges -Dhttp.keepAlive=false -Dformatter.skip
```

.pipeline/scripts/get-maven-excludes.py
```diff
@@ -0,0 +1,39 @@
...
+def write_maven_excludes_github_output(filter_key, filter_value):
+    with open("module-inventory.json", "r") as file:
+        module_inventory = json.load(file)
+
+        filtered_inventory = filter(lambda x: x[filter_key] == filter_value, module_inventory)
+        artifact_ids = list(map(lambda x: x["artifactId"], filtered_inventory))
+        artifact_ids_for_maven_excludes = list(map(lambda x: "!:" + x, artifact_ids))
+
+        excludes = ','.join(artifact_ids)
+        print("Excludes: ",excludes)
```
Johannes-Schneider, line 14: [minor - code style]

```suggestion
        print("Excludes: ", excludes)
```

.github/workflows/build.yaml
```diff
@@ -302,4 +310,62 @@ jobs:
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
         with:
-          category: "/language:java"
\ No newline at end of file
+          category: "/language:java"
+
+  deploy-snapshot:
+    name: Deploy Snapshot
+    if: ${{ github.ref == 'refs/heads/main' }}
```
Johannes-Schneider, line 317: [minor]

This condition seems to be fine for the moment.
However, in the future, there might actually be multiple branches we are deploying snapshots from.
(Right now, we are using the scheme `main/v*` for our major version branches, which all need to be able to deploy snapshots)

.pipeline/scripts/get-maven-excludes.py
```diff
@@ -0,0 +1,39 @@
+import os
+import json
+import argparse
+
+def write_maven_excludes_github_output(filter_key, filter_value):
+    with open("module-inventory.json", "r") as file:
+        module_inventory = json.load(file)
+
+        filtered_inventory = filter(lambda x: x[filter_key] == filter_value, module_inventory)
```
Johannes-Schneider, line 9: This comparison might become a problem when filtering for non-string properties (e.g. boolean values).

```suggestion
        filtered_inventory = filter(lambda x: x[filter_key] == str(filter_value), module_inventory)
```

KevinHudemann: Good point, but I guess you meant it like this right?
```suggestion
        filtered_inventory = filter(lambda x: str(x[filter_key]) == filter_value, module_inventory)
```
I mean `filter_values` should always be a string as it is a parameter and argparse should parse it as string by default, but the value in the modules-inventory might be a non string property like the `excludeFromBlackDuckScan` property.

Johannes-Schneider: Yes, you're right!

KevinHudemann:
> I just had a look at the pipeline logs and it seems like the `mvn deploy` is actually re-compiling the modules:
> 
> Logs
> I'm not sure whether this can be mitigated (sources might simply be flagged as _"stale"_ if their timestamp is "too" old) but maybe using the "unzip and touch" workaround (basically `touch` every file in the unzipped "target" folders so that the last modification timestamp is new enough for maven to not re-compile everything) could be considered.

It is not only recompiling but also rebuilding. In addition, it is actually doing the same in the old pipeline.
The question is I guess what you'd like to have. For the Cloud SDK Pipeline we have built a step (nexusUpload) to avoid rebuilding the artifacts and we could also switch to this step if you'd prefer that.

.github/workflows/build.yaml
KevinHudemann: @Johannes-Schneider Sure that's doable. You could use functions like [`contains` or `startsWith`](https://docs.github.com/en/actions/learn-github-actions/expressions#startswith) or simply add an `||` and list the branches for which you want to execute the step.
I'd say this can be changed once there is the need to deploy snapshots for other branches, as well.
What do you think?

Johannes-Schneider (approved):
LGTM

.github/workflows/build.yaml
```diff
@@ -302,4 +310,62 @@ jobs:
...
+
+      # Target
+      - name: Download Target Dirs
+        uses: actions/download-artifact@v3
+        with:
+          name: target-dirs
+          path: .
+      - name: UNZIP Target Dirs
+        run: |
+          unzip -DD target.zip
```
Johannes-Schneider, line 360: [minor]

We could revert the change as it doesn't change anything I guess.
That way, the pipeline looks a bit more consistent.

```suggestion
          unzip -D target.zip
```

KevinHudemann: It at least avoids the recompile.

.github/workflows/build.yaml
```diff
@@ -302,4 +310,62 @@ jobs:
...
+      - name: Publish Snapshot
+        run: >
+          mvn $MVN_CLI_ARGS
+          -DaltSnapshotDeploymentRepository=artifactory-snapshots::https://common.repositories.cloud.sap/artifactory/build-snapshots-cloudsdk
```
MatKuhr, line 364: I think you could replace `-Dformatter.skip -Dimpsort.skip` with `-DskipFormatting", this will also skip the maven POM sorter 😉

PR #28, opened 2023-08-30
Test archetypes

daniel-kurzynski:


.github/workflows/build.yaml
```diff
@@ -312,10 +317,127 @@ jobs:
...
+          -Dmaven.wagon.http.pool=false
+
+      - name: Start ${{ matrix.task.archetype }}
+        working-directory: ./example-${{ matrix.task.archetype }}/application
+        run: |
+          logFilePath=log.txt
+          ${{ matrix.task.startCommand }} > $logFilePath 2>&1 &
+          
+          if ! curl -s --retry 60 --retry-delay 3 --retry-all-errors http://127.0.0.1:8080/hello ; then
+            echo "Project generated from archetype ${archetype} failed to start locally."
```
KevinHudemann, line 399:
```suggestion
            echo "Project generated from archetype ${{ matrix.task.archetype }} failed to start locally."
```

.github/workflows/build.yaml
```diff
@@ -312,10 +317,127 @@ jobs:
...
+            cat $logFilePath
+            exit 1
+          fi
+          
+          if ! grep -q "I am running!" $logFilePath; then
+            echo "Project generated from archetype ${archetype} started locally, but did not contain the expected log output."
```
KevinHudemann, line 405:
```suggestion
            echo "Project generated from archetype ${{ matrix.task.archetype }} started locally, but did not contain the expected log output."
```

.github/workflows/build.yaml
```diff
@@ -312,10 +317,127 @@ jobs:
...
+            cat $logFilePath
+            exit 1
+          fi
+          
+          if grep -iq "caused by" $logFilePath; then
+            echo "Error in log file:"
```
KevinHudemann, line 411:
```suggestion
            echo "Project generated from archetype ${{ matrix.task.archetype }} started locally, but an unexpected error occurred."
```

.github/workflows/build-sanity.yaml
```diff
@@ -59,6 +76,124 @@ jobs:
...
+          -Dmaven.wagon.http.pool=false
+
+      - name: Start ${{ matrix.task.archetype }}
+        working-directory: ./example-${{ matrix.task.archetype }}/application
+        run: |
+          logFilePath=log.txt
+          ${{ matrix.task.startCommand }} > $logFilePath 2>&1 &
+
+          if ! curl -s --retry 60 --retry-delay 3 --retry-all-errors http://127.0.0.1:8080/hello ; then
+            echo "Project generated from archetype ${archetype} failed to start locally."
```
KevinHudemann, line 159:
```suggestion
            echo "Project generated from archetype ${{ matrix.task.archetype }} failed to start locally."
```

.github/workflows/build-sanity.yaml
```diff
@@ -59,6 +76,124 @@ jobs:
...
+            cat $logFilePath
+            exit 1
+          fi
+
+          if ! grep -q "I am running!" $logFilePath; then
+            echo "Project generated from archetype ${archetype} started locally, but did not contain the expected log output."
```
KevinHudemann, line 165:
```suggestion
            echo "Project generated from archetype ${{ matrix.task.archetype }} started locally, but did not contain the expected log output."
```

archetypes/scp-cf-spring/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,83 @@
...
+        <fileSet filtered="true">
+            <directory>integration-tests/src/test/resources</directory>
+            <includes>
+                <include>**</include>
+            </includes>
+        </fileSet>
+
+        <!-- cx-server folder -->
+        <fileSet filtered="false">
+            <directory>cx-server</directory>
```
KevinHudemann, line 77: I guess this one needs to be removed as well.

archetypes/scp-cf-spring-jakarta/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,83 @@
...
+        <fileSet filtered="true">
+            <directory>integration-tests/src/test/resources</directory>
+            <includes>
+                <include>**</include>
+            </includes>
+        </fileSet>
+
+        <!-- cx-server folder -->
+        <fileSet filtered="false">
+            <directory>cx-server</directory>
```
KevinHudemann, line 77: Also this one

archetypes/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
+	<!--
+    !!! Add all modules that should never be released here. If left empty this can be removed. !!!
+    -->
+	<profiles>
+		<profile>
+			<id>build-sap-internal-modules</id>
```
KevinHudemann, line 89: This one can be removed I guess?

.github/workflows/build-sanity.yaml
```diff
@@ -59,6 +76,124 @@ jobs:
...
+            cat $logFilePath
+            exit 1
+          fi
+
+          if grep -iq "caused by" $logFilePath; then
+            echo "Error in log file:"
```
KevinHudemann, line 171:
```suggestion
            echo "Project generated from archetype ${{ matrix.task.archetype }} started locally, but an unexpected error occurred."
```

.github/workflows/build.yaml
```diff
@@ -312,10 +317,126 @@ jobs:
...
+      - name: Purge SDK Artifacts from Maven Repository
+        run: |
+          rm -rf $HOME/.m2/repository/com/sap/cloud/sdk/
+      - name: Save Maven Repository
+        uses: actions/cache/save@v3
+        if: steps.restore-cache.outputs.cache-hit != 'true'
+        id: save-cache
+        with:
+          path: ~/.m2/repository
+          key: maven-central
```
Johannes-Schneider, line 434: [question]

Why are these two steps needed?
As far as I'm aware, `steps.restore-cache.outputs.cache-hit` should always be `true` as we are creating the cache latest in the `build` step (so there will always be a cache once we are reaching this step of the pipeline).

.github/workflows/build.yaml
```diff
@@ -312,10 +317,126 @@ jobs:
...
+              "archetype": "scp-cf-spring",
+              "javaVersion": 8,
+              "startCommand": "mvn spring-boot:run -B",
+            }
+          - {
+              "archetype": "scp-cf-spring-jakarta",
+              "javaVersion": 17,
+              "startCommand": "mvn spring-boot:run -B",
+            }
+    name: ${{ matrix.task.archetype }}
```
Johannes-Schneider, line 339: Let's maybe add a prefix here so that it is more obvious what is actually happening in this step.

```suggestion
    name: Test Archetype ${{ matrix.task.archetype }}
```

.github/workflows/build.yaml
daniel-kurzynski: That one is a different maven cache than the one in build. It is because maybe we decide to use a different repository than maven central for build, e.g. if there are internal dependencies (e.g. auditlog). For testing the archetypes we have to make sure that the cache contains only what a user could download ...

Johannes-Schneider: Ahh... I see.
The crucial part here is ofc the `key` of the cache (which is `maven-central` in this step, but we use just `maven` everywhere else).
Makes sense.

archetypes/pom.xml
daniel-kurzynski: @MatKuhr @Johannes-Schneider ?

.github/workflows/build-sanity.yaml
```diff
@@ -59,6 +76,123 @@ jobs:
...
+          mvn archetype:generate -B
+          -DarchetypeCatalog=local
+          -DarchetypeGroupId=com.sap.cloud.sdk.archetypes
+          -DarchetypeArtifactId=${{ matrix.task.archetype }}
+          -DarchetypeVersion=${{ needs.build.outputs.version }}
+          -DgroupId=com.sap.test
+          -DartifactId=example-${{ matrix.task.archetype }}
+          -Dversion=1.0-SNAPSHOT
+          -Dpackage=com.sap.test
+          -Dhttp.keepAlive=false
```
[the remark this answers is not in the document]
daniel-kurzynski: We found the issue. The archetype-catalog.xml is stored in .m2/repository/archetype-catalog.xml and thus was not included in the m2 stash. We added the catalog to the artefacts and the warning is gone.

archetypes/pom.xml
MatKuhr: should be safe to be removed, yes. Relates to https://github.com/SAP/cloud-sdk-java-backlog/issues/90

As this module will be publicly released with SDK 5 we probably have to update our docs. I created a small reminder note for us to d

Johannes-Schneider:
It seems like your last commit ([d2b4620](https://github.com/SAP/cloud-sdk-java/pull/28/commits/d2b4620c7c97e93935fc8bc42caa85eb07af5002)) broke the pipeline.
This is because the `jakarta` archetype is no longer declared as a module in the `archetype-parent`, so it won't be installed locally during the `build` step.

I see that we are currently actually not releasing that archetype... But I'm not exactly sure why.
I'll reach out to the team via Slack.

Johannes-Schneider (approved):
I didn't pay too much attention to the actual content of the archeytpes, but since our tests apparently succeed I assume everything is fine.

So LGTM.

PR #31, opened 2023-09-01
Notify via slack in case CI workflow on main fails

daniel-kurzynski:
BLI https://github.com/SAP/cloud-sdk-java-backlog/issues/282

### DIFF
--- a/.github/workflows/blackduck.yaml
+++ b/.github/workflows/blackduck.yaml
@@ -42,3 +42,17 @@ jobs:
           DETECT_MAVEN_BUILD_COMMAND: -pl ${{ steps.get-maven-excludes-for-blackduck.outputs.PREFIXED_EXCLUDES }},!:openapi-generator-cli
           DETECT_TIMEOUT: 7200
           PROJECT_VERSION: "5"
+#  notify-job:
+#    runs-on: ubuntu-latest
+#    needs: [ scan ]
+#    if: ${{ always() && github.ref == 'refs/heads/main' && contains(needs.*.result, 'failure') }}
+#    steps:
+#      - name: Checkout
+#        uses: actions/checkout@v3
+#      - name: Notify
+#        run: python .pipeline/scripts/notify.py
+#        env:
+#          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+#          WORKFLOW: ${{ github.workflow }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -284,6 +284,7 @@ jobs:
 
   codeql:
     name: Analyze with CodeQL
+    needs: formatting
```
Johannes-Schneider, line 287: Good catch!

.github/workflows/build.yaml
```diff
@@ -496,3 +497,18 @@ jobs:
         env:
           ARTIFACTORY_USER: ${{ secrets.ARTIFACTORY_USER }}
           ARTIFACTORY_TOKEN: ${{ secrets.ARTIFACTORY_TOKEN }}
+  notify-job:
+    runs-on: ubuntu-latest
+    needs: [formatting, build, test, pmd, spotbugs, codeql, archetypes, deploy-snapshot]
+    if: ${{ always() && github.ref == 'refs/heads/main' && contains(needs.*.result, 'failure') }}
```
Johannes-Schneider, line 503: > `if always() && ... `

sounds a bit suspicious.
Is `always()` used to prevent this step from not being executed if one of the prerequisites fails?

daniel-kurzynski: Yes, but failure() seems to be sufficient. Updated the if condition.

.github/workflows/blackduck.yaml
```diff
@@ -42,3 +42,17 @@ jobs:
...
+#    steps:
+#      - name: Checkout
+#        uses: actions/checkout@v3
+#      - name: Notify
+#        run: python .pipeline/scripts/notify.py
+#        env:
+#          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+#          WORKFLOW: ${{ github.workflow }}
+#          WORKFLOW_RUN_URL: https://github.com/SAP/cloud-sdk-java/actions/runs/${{ github.run_id }}
+#          BRANCH_NAME: ${{ github.ref_name }}
```
Johannes-Schneider, line 58: Let's uncomment the slack notification and instead comment the cron trigger in this job as we currently don't really want to run BlackDuck on the OSS repo anyways.

daniel-kurzynski: Done

stippi2 (approved):
LGTM.

Johannes-Schneider (approved):
LGTM

PR #39, opened 2023-09-05
Improve release workflow

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/282

Release workflow:
- trigger the [prepare release workflow](https://github.com/SAP/cloud-sdk-java/actions/workflows/prepare_release.yml)
- Check/approve the automatically created PR and publish the release
- perform release workflow will be triggered once the release is published

Example draft release (should be deleted after PR review): https://github.com/SAP/cloud-sdk-java/releases/tag/untagged-df758d5dc3827df5f1d8
Example PR that gets created when triggering the `prepare_release` workflow (although from the wrong branch, would usually be main): https://github.com/SAP/cloud-sdk-java/pull/38


.github/workflows/prepare_release.yml
```diff
@@ -61,25 +65,300 @@ jobs:
...
 
           git push origin $BRANCH_NAME:$BRANCH_NAME
 
-          gh release create "rel/${{ steps.get-release-versions.outputs.RELEASE_VERSION }}" \
-            --target $RELEASE_COMMIT_ID \
-            --title "Release ${{ steps.get-release-versions.outputs.RELEASE_VERSION }}" \
-            --draft --generate-notes sap-cloud-sdk-java-${{ steps.get-release-versions.outputs.RELEASE_VERSION }}.zip
 
-          rm -rf sap-cloud-sdk-java-${{ steps.get-release-versions.outputs.RELEASE_VERSION }}.zip rel-${{ steps.get-release-versions.outputs.RELEASE_VERSION }}
+  build:
```
Johannes-Schneider, line 73: I'm wondering whether we can somehow circumvent to (more or less) copy & paste the entire CI pipeline into this pipeline.

Idea:
What if we would use the [wait on check](https://github.com/marketplace/actions/wait-on-check) action after we have updated the version to our release version and created the PR.
That way, we could wait with the second commit (setting the version to the next snapshot version) until the CI pipeline on our newly created PR went through successfully.

Problem could be that we cannot access the built `*.jar` files (maybe?).

daniel-kurzynski: > Problem could be that we cannot access the built *.jar files (maybe?).

We also hat a long discussion about that. But essentially the reason was that we want to assure to build only once and would somehow get the artifacts copied over. 
Alternative would have been to merge o master and trigger the release by that event, using the main pipeline to create the release draft ...
But finally we thought, the pipeline structure is less complex if we just have different pipelines, although we have to duplicate some parts.

daniel-kurzynski: I guess we could discuss to extracts some pipeline parts into reusable actions.

daniel-kurzynski: Btw. I guess the the release pipeline might be different from the main. It is rather similar to the sanity build, but might include more, e.g. blackduck, etc ...

.github/workflows/prepare_release.yml
```diff
@@ -61,25 +65,314 @@ jobs:
...
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
+        with:
+          category: "/language:java"
+
+
+  create-release:
+    runs-on: ubuntu-latest
+    needs: [archetypes, codeql, prepare-release]
```
Johannes-Schneider, line 316: I think we should include Blackduck and Fosstar before actually performing the release.

.github/workflows/prepare_release.yml
```diff
@@ -61,25 +65,314 @@ jobs:
...
+        run: echo "VERSION=$(cat latest.json | jq -r .version)" >> $GITHUB_OUTPUT
+        id: get-version
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: 17
+      - name: Test/Install/Package
+        run: |
+          mvn $MVN_CLI_ARGS install jacoco:report org.apache.maven.plugins:maven-deploy-plugin:3.1.1:deploy -DaltReleaseDeploymentRepository=local::file:./rel-${{ needs.prepare-release.outputs.release_version }}
```
Johannes-Schneider, line 98: This command doesn't exclude our _not to be released_ modules yet, right?

Johannes-Schneider (changes requested):
Please make sure to not release any of our _not to be released_ modules.

.github/workflows/prepare_release.yml
KevinHudemann: That's true but also part of another BLI (https://github.com/SAP/cloud-sdk-java-backlog/issues/90) 🙂

KevinHudemann: And we don't release anything yet.

Johannes-Schneider: > That's true but also part of another BLI (https://github.com/SAP/cloud-sdk-java-backlog/issues/90) 🙂

Ah okay, wasn't aware.

> And we don't release anything yet.

Yes, but the modules will be part of the draft release (attached as a `zip` file), right?
So technically consumers could download compiled `.jar` files from our GH release, if we were to publish it.

Johannes-Schneider: But if you'd like to handle those concerns as part of the other BLI, I'm fine with that.

KevinHudemann: Yes we already picked the other BLI up so I think its ok to handle it with a follow-up.

.github/workflows/prepare_release.yml
KevinHudemann: Sure will do.

.github/workflows/prepare_release.yml
Johannes-Schneider: Do we also handle the inclusion of BlackDuck and Fosstar as part of an other ticket?

stippi2: My guess is that the deploy target needs to be called in a separate mvn invokation using excludes that are generated just like for the snapshot release.

KevinHudemann: Nope I'd include it in this PR. Already on it.

.github/workflows/prepare_release.yml
```diff
@@ -14,15 +14,20 @@ env:
...
+      release_version: ${{ steps.get-release-versions.outputs.RELEASE_VERSION }}
+      new_snapshot: ${{ steps.get-release-versions.outputs.NEW_SNAPSHOT }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-        with:
-          ref: 'main'
-          fetch-depth: '0'
+        # with:
+        #   ref: 'main'
```
Johannes-Schneider, line 30: This should be uncommented before merging this PR, right?

```suggestion
        with:
          ref: 'main'
```

.github/workflows/prepare_release.yml
```diff
@@ -61,25 +66,370 @@ jobs:
...
+      #      - name: Purge SDK Artifacts from Maven Repository
+      #        run: |
+      #          rm -rf $HOME/.m2/repository/com/sap/cloud/sdk/
+      #      - name: Save Maven Repository
+      #        uses: actions/cache/save@v3
+      #        if: steps.restore-cache.outputs.cache-hit != 'true'
+      #        id: save-cache
+      #        with:
+      #          path: ~/.m2/repository
+      #          key: maven-central
```
Johannes-Schneider, line 268: Is this supposed to be uncommented before merging as well?

Johannes-Schneider (commented):
Generally LGTM - just uncomment the commented blocks (maybe?)

.github/workflows/prepare_release.yml
KevinHudemann: This cannot be uncommented as the verify of the archetypes doesn't work yet, since not all modules are part of this repo yet.
Thats why we hardcoded the sdk version in the archetypes for now to 4.20.

Johannes-Schneider (approved):

PR #40, opened 2023-09-05
Add checkstyle rule for TODO and FIXME comments

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/282

### DIFF
--- a/.pipeline/config.yml
+++ b/.pipeline/config.yml
@@ -26,7 +26,7 @@ stages:
       class: -1
   codeCheck:
     checkstyle:
-      high: '0'
+      high: '3'
       normal: '-1'
       low: '-1'
     findbugs:
--- a/codestyle/checkstyle.xml
+++ b/codestyle/checkstyle.xml
@@ -88,7 +88,7 @@
 			<property name="severity" value="warning"/>
 		</module>
 		<module name="TodoComment">
-      <property name="format" value="(TODO)|(FIXME)"/>
-    </module>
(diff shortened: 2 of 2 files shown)


stippi2 (commented):
From the old [Jenkinsfile](https://github.wdf.sap.corp/MA/sdk/blob/b0776cc8dbb0efe458ffe8d25629f36d10705eb2/Jenkinsfile#L723), I understand it was case insensitive. The [examples](https://checkstyle.sourceforge.io/checks/misc/todocomment.html#Example2-code) in checkstyle suggest the check would no longer be case-insensitve. Not sure if this was already discussed and deemed OK.

KevinHudemann:
> From the old [Jenkinsfile](https://github.wdf.sap.corp/MA/sdk/blob/b0776cc8dbb0efe458ffe8d25629f36d10705eb2/Jenkinsfile#L723), I understand it was case insensitive. The [examples](https://checkstyle.sourceforge.io/checks/misc/todocomment.html#Example2-code) in checkstyle suggest the check would no longer be case-insensitve. Not sure if this was already discussed and deemed OK.

Yeah good point. No it was not discussed.
Maybe adding both in lower case would be sufficient.

stippi2:
> Maybe adding both in lower case would be sufficient.

ChatGPT claims that regex like `(?i)` is not supported by this module and also suggests just listing (most common) permutations.

Johannes-Schneider (approved):
LGTM

PR #41, opened 2023-09-05
Update Fosstars workflow and trigger scan on every push

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/110

Report is generated in the fosstars-report branch: https://github.com/SAP/cloud-sdk-java/blob/fosstars-report/fosstars_report.md

Badge is already part of the Readme: https://github.com/SAP/cloud-sdk-java/tree/main#disclaimer-we-are-working-on-the-migration-of-this-repository-the-planned-timeline-for-the-os-migration-and-sdk-version-5-is-q4-2023-please-stay-tuned


.github/workflows/fosstars-report.yml
```diff
@@ -1,16 +1,18 @@
-name: 'Fosstars (Security)'
+name: "Fosstars (Security)"
 on:
   workflow_dispatch:
+  push:
+    branches: ["main"]
```
Johannes-Schneider, line 5: I'm wondering: If we are including the fosstar check on every push on `main`, why don't we make it part of the regular CI pipeline (that way, we wouldn't merge vulnerable code to `main` in the first place)?

Also: If we are checking on every push, do we really need the cron trigger in addition? 
(Of course, vulnerabilities in dependencies might always be found, so having the `cron` trigger provides some additional safety, but that is pretty neglectable IMO - especially since DependaBot will also check for vulnerabilities continiously)

KevinHudemann: One problem of adding it to the build pipeline is that the action also updates the report if it is running in a PR, i.e., overwriting it. To me it seemed like this doesn't make sense since multiple PRs would overwrite the report all the time.
Of course we can discuss if its required to run it on every push or if the cron trigger is enough or the other way round. The ticket stated it should run as part of the build pipeline and I thought the closest alternative is to run it on every push.

The JS SDK team only has a cron trigger for Fosstars.

KevinHudemann: I would even say it's alright to have it as cron triggered workflow, since the tool is rather checking how or if you use scan tools and it is not a replacement for the security scans itself.

Johannes-Schneider: [The Service Binding Library](https://github.com/SAP/btp-environment-variable-access/blob/main/.github/workflows/fosstars-project-report.yml) also just uses a `cron` trigger.

IMO, this is sufficient - but no strong opinion.

Johannes-Schneider (approved):

PR #53, opened 2023-09-11
Improve perform release

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/90

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -94,7 +94,7 @@ jobs:
           java-version: 17
       - name: Test and Package
         run: |
-          mvn $MVN_CLI_ARGS package jacoco:report
+          mvn $MVN_CLI_ARGS verify jacoco:report
           python .pipeline/scripts/generate-release-artifacts.py --version ${{ needs.prepare-release.outputs.release_version }} --path-prefix release
           zip -q -r sap-cloud-sdk-java-${{ needs.prepare-release.outputs.release_version }}.zip release
       - name: Local Changes Check
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -173,7 +173,7 @@ jobs:
           cd release
           rm -rf /home/runner/.m2/repository/archetype-catalog.xml
           mvn $MVN_CLI_ARGS install
-          mvn -f ${{ matrix.task.pomPath }} archetype:update-local-catalog
+          mvn $MVN_CLI_ARGS archetype:crawl
 
           cd ..
           rm -rf release
@@ -258,6 +258,7 @@ jobs:
       #      - name: Purge SDK Artifacts from Maven Repository
       #        run: |
       #          rm -rf $HOME/.m2/repository/com/sap/cloud/sdk/
+      #          rm -rf $HOME/.m2/repository/archetype-catalog.xml
       #      - name: Save Maven Repository
       #        uses: actions/cache/save@v3
       #        if: steps.restore-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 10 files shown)


.github/workflows/prepare_release.yml
```diff
@@ -93,10 +92,11 @@ jobs:
         with:
           distribution: "temurin"
           java-version: 17
-      - name: Test/Install/Package
+      - name: Test and Package
         run: |
-          mvn $MVN_CLI_ARGS install jacoco:report org.apache.maven.plugins:maven-deploy-plugin:3.1.1:deploy -DaltReleaseDeploymentRepository=local::file:./rel-${{ needs.prepare-release.outputs.release_version }}
-          zip -q -r sap-cloud-sdk-java-${{ needs.prepare-release.outputs.release_version }}.zip rel-${{ needs.prepare-release.outputs.release_version }}
+          mvn $MVN_CLI_ARGS install jacoco:report
+          python .pipeline/scripts/generate-release-artifacts.py --version ${{ needs.prepare-release.outputs.release_version }} --path-prefix release
```
Johannes-Schneider, line 98: This job still contains the `get-version` step, which seems to be superfluous as we already determine the release version in the `prepare-release` job, right?

Johannes-Schneider (commented):
It seems like the `set-next-snapshot-version` script doesn't correctly respect the given `release-version` parameter:
When I give a specific release version (e.g. `0.0.2` as done in [the example release](https://github.com/SAP/cloud-sdk-java/pull/56/files#)), I would expect the next snapshot to respect that value accordingly (i.e. the next snapshot version should have been `0.0.2-SNAPSHOT` in the example).

KevinHudemann:
> It seems like the `set-next-snapshot-version` script doesn't correctly respect the given `release-version` parameter: When I give a specific release version (e.g. `0.0.2` as done in [the example release](https://github.com/SAP/cloud-sdk-java/pull/56/files#)), I would expect the next snapshot to respect that value accordingly (i.e. the next snapshot version should have been `0.0.2-SNAPSHOT` in the example).

Yeah Florian wrote the script in a way that always the minor version will be incremented. We assumed there was a reason for that. 
Would you instead always just want to increment the patch version?

.github/workflows/prepare_release.yml
KevinHudemann: Yes that's right, I've removed it.

Johannes-Schneider:
> > It seems like the `set-next-snapshot-version` script doesn't correctly respect the given `release-version` parameter: When I give a specific release version (e.g. `0.0.2` as done in [the example release](https://github.com/SAP/cloud-sdk-java/pull/56/files#)), I would expect the next snapshot to respect that value accordingly (i.e. the next snapshot version should have been `0.0.2-SNAPSHOT` in the example).
> 
> Yeah Florian wrote the script in a way that always the minor version will be incremented. We assumed there was a reason for that. Would you instead rather want to increment the patch version?

In our old pipeline, we were able to set the `nextSnapshotVersion` manually (actually, it was mandatory to specify that parameter when running the job).
I'm wondering whether we also need / want that parameter here (but this time, we make it optional and by default the minor version is increased)?

I think this change is not super needed for now. So no immediate action required from my perspective.

Johannes-Schneider (approved):
I just found that the BlackDuck scan doesn't use the `needs.prepare-release.outputs.release_version` yet.
(Can be addressed in a follow-up PR)

---

Otherwise LGTM

KevinHudemann:
> In our old pipeline, we were able to set the `nextSnapshotVersion` manually (actually, it was mandatory to specify that parameter when running the job). I'm wondering whether we also need / want that parameter here (but this time, we make it optional and by default the minor version is increased)?
> 
> I think this change is not super needed for now. So no immediate action required from my perspective.

Sure that would also be an option. Should not be a lot of work, I guess we can quickly address this in a follow-up PR.

PR #57, opened 2023-09-11
Add create module inventory script

KevinHudemann:


.github/workflows/build.yaml
```diff
@@ -84,6 +84,9 @@ jobs:
         # they require the JARs of all modules to work, so those need to be installed
         run: |
           mvn $MVN_CLI_ARGS install -DskipTests -DskipFormatting
+      - name: Create module-inventory
+        run: |
+          python ./scripts/create_module_inventory_file.py --sdk-root-directory ./ --output-file ./module-inventory.json --script-config ./scripts/common/_maven_module/maven_module_reader_configuration.json
```
Johannes-Schneider, line 89: Should this be added to the `build-sanity` pipeline as well?

Johannes-Schneider (approved):
LGTM.

We might want to consider whether to add the module-inventory check to the sanity build pipeline as well, but not mandatory IMO.

PR #85, opened 2023-10-05
Use all runner cores for mvn executions

stippi2:
The number of available cores of the GitHub runner is added to the relevant mvn CLI args. For the one test run, I don't see any significant change for the build workflow. The reason could be that the runner only had 2 cores, and already there are jobs executed in parallel during the workflow. At least the time has not gotton worse and there is potential for a positive effect in the future, and thus I don't withdraw the PR. :-)


.github/workflows/build.yaml
```diff
@@ -88,6 +92,8 @@ jobs:
           python .pipeline/scripts/generate-release-artifacts.py --version ${{ steps.get-version.outputs.VERSION }} --path-prefix release
           zip -q -r sap-cloud-sdk-java-${{ steps.get-version.outputs.VERSION }}.zip release
           rm -rf release
+        env:
+          MVN_CLI_ARGS: ${{ env.MVN_CLI_ARGS }} -T ${{ steps.get-core-count.outputs.CORE_COUNT }}
```
CharlesDuboisSAP, line 96:
```suggestion
          MVN_CLI_ARGS: ${{ env.MVN_CLI_ARGS }} -T 1C # 1 thread per cpu core
```
You can also try 2C there might be an additional performance improvement.

stippi2: Thanks, I wasn't aware of the `1C` notation which let's maven itself figure out the core count. I would rework this PR and make it part of the global CLI args...

CharlesDuboisSAP (approved):

PR #94, opened 2023-10-17
Deploy snapshots to Maven central

KevinHudemann:
BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/333
This PR adds a workflow to manually trigger a snapshot deployment to Maven Central.
The workflow has the input parameter `skip_blackduck_scan` which can be set to true or false, depending on whether the check should be skipped.
Examples for both cases can be found here:
`skip = true`: https://github.com/SAP/cloud-sdk-java/actions/runs/6547414695
`skip = false`: https://github.com/SAP/cloud-sdk-java/actions/runs/6548084491
The snapshot artifacts can be found here: https://oss.sonatype.org/#view-repositories;snapshots~browsestorage~com/sap/cloud

image

The workflow is basically a copy of the build.yaml but has been extended to run the blackduck scan, if required.


.github/workflows/deploy_snapshot.yaml
```diff
@@ -0,0 +1,621 @@
+name: Deploy Snapshot to Maven Central
+
+on:
+  workflow_dispatch:
+    inputs:
+      skip_blackduck_scan:
+        description: "Skip Blackduck Scan (true or false), default is false."
+        default: "false"
+        required: false
```
stippi2, line 9: I think there are parameter types which would result in different UI (not sure, but it should result in a checkmark). See here: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#defining-inputs-for-manually-triggered-workflows

KevinHudemann: Good point I'll try it out. Also didn't see the `choice` type before. I just read in an issue that there were problems with boolean inputs in the if conditions, which is why I opted for string but I'll quickly try it out. 👍

stippi2: I also stumbled across that when trying to find a reference. But I think that's for (reusable) Actions, not Workflows.

KevinHudemann: Switched to boolean input type, now it looks like this:
image

stippi2 (dismissed):

CharlesDuboisSAP (commented):
The notify-job never ran.
Also could you add the tech debt to this item: [[Pipeline] Remove GitHub workflows code duplication](https://github.com/SAP/cloud-sdk-java-backlog/issues/318#top)

KevinHudemann:
> The notify-job never ran. Also could you add the tech debt to this item: [[Pipeline] Remove GitHub workflows code duplication](https://github.com/SAP/cloud-sdk-java-backlog/issues/318#top)

Good catch! The notify job was missing the `scan` job in the `needs` property. In general, the notify job only runs if there was a failure.

.github/workflows/deploy_snapshot.yaml
```diff
@@ -0,0 +1,623 @@
...
+  scan:
+    name: "Blackduck Scan"
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    if: ${{ !github.event.inputs.skip_blackduck_scan }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          ref: ${{ needs.prepare-release.outputs.branch_name }}
```
stippi2, line 388: I would say this cannot work.

stippi2 (changes requested):

.github/workflows/deploy_snapshot.yaml
KevinHudemann: Good catch. It worked actually, but I guess only because there is a reasonable fallback. I'll delete the ref

stippi2 (dismissed):
👍

stippi2 (approved):

PR #95, opened 2023-10-17
Remove unused class `OAuth2DestinationBuilder`

newtork:
Minor housekeeping

_(from old repo)_


MatKuhr:
This is currently used by customers: https://github.tools.sap/search?q=OAuth2DestinationBuilder.&type=code

If we remove it we probably need to offer an alternative. Maybe making the header provider public?

newtork:
> If we remove it we probably need to offer an alternative.

I went through some options but they didn't look convincing.

> Maybe making the header provider public?

Yeah.. before I'm doing that, I'll keep the current convenience API. Presumably it stays experimental.

PR #98, opened 2023-10-19
Fix: Add `providerTenantId` to `MegacliteServiceBinding`

CharlesDuboisSAP:
# Pull Request Description 🤖

This PR is part of the work on backlog item [#297](https://github.com/SAP/cloud-sdk-java-backlog/issues/297). This problem was uncovered after the use of ScpCf destination API on the v5 `DwC-test-app`.

## Changes

The changes in this PR include:

- A new `dwcConfiguration` field has been added to the `MegacliteServiceBinding` class, it can be (re)set after the binding is built.
- The `getCredentials()` method in the `MegacliteServiceBinding` class has been updated to return a map containing the `tenantid` and `providerTenantId`, or, throws the Exception created by `DwcConfiguration`.
- The `testPropertiesOfTheOpenSourceServiceBinding()` method in the `MegacliteServiceBindingTest` class has been updated to include a mock of the `DwcConfiguration` class and to check that the `tenantid` is correctly set in the credentials.
- `testGetCredentialsThrowsExceptionWithoutProviderTenantId()` has been added to check that the exception created by `DwcConfiguration` is propagated.

## Reviewer Instructions

Please review the changes in the context of the associated backlog item. Ensure that the changes align with the requirements and that the updated tests cover the new functionality.

### DIFF
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DwcConfiguration.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DwcConfiguration.java
@@ -51,18 +51,26 @@ final class DwcConfiguration
         this.providerTenant = Lazy.of(() -> providerTenant);
     }
 
+    @Nonnull
     URI megacliteUrl()
+        throws CloudPlatformException,
+            IllegalArgumentException
     {
         return megacliteUrl.get();
     }
 
+    @Nonnull
     String providerTenant()
+        throws CloudPlatformException,
+            IllegalArgumentException
     {
         return providerTenant.get();
     }
 
     @Nonnull
     private static URI loadMegacliteUri( @Nonnull final Function<String, String> environmentVariableReader )
+        throws CloudPlatformException,
+            IllegalArgumentException
     {
         final Option<String> dwcApplication = Option.of(environmentVariableReader.apply(DWC_APPLICATION));
         if( dwcApplication.isEmpty() ) {
@@ -75,6 +83,8 @@ final class DwcConfiguration
 
     @Nonnull
     private static String loadProviderTenantId( @Nonnull final Function<String, String> environmentVariableReader )
+        throws CloudPlatformException,
+            IllegalArgumentException
     {
         final Option<String> dwcApplication = Option.of(environmentVariableReader.apply(DWC_APPLICATION));
         if( dwcApplication.isEmpty() ) {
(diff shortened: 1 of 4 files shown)


cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBinding.java
```diff
@@ -368,7 +371,12 @@ public MegacliteServiceBinding build()
             } else {
                 subscriberConfiguration = currentConfiguration;
             }
-            return new MegacliteServiceBinding(service, providerConfiguration, subscriberConfiguration);
+            final String providerTenantId = DwcConfiguration.getInstance().providerTenant();
```
MatKuhr, line 374: this is not ideal for testing, as you have to mock the static method below. For an example how it can be improved check out the test setup on `MegacliteDestinationFactory` and `MegacliteDestinationFactoryTest`.

I think you could achieve the same here by moving `DwcConfiguration.getInstance()` into the `getCredentials()` instead. Added bonus of doing this: `.providerTenant()` is only computated once it is actually needed, not before.

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBinding.java
```diff
@@ -66,6 +66,9 @@ public final class MegacliteServiceBinding implements ServiceBinding
     @Nullable
     private final MandateConfiguration subscriberConfiguration;
 
+    @Nullable
+    static DwcConfiguration dwcConfiguration;
```
MatKuhr, line 70: while this approach works, it might be a bit easier to make this non-static and assign to `DwcConfiguration.getInstance()` by default. Then, with a package private setter, change it on the binding object in the test class.

The downside of having this being static right now is that you may have unintended side effects. We are setting it for one test case below, but actually it is set for all test cases globally now.

Finally, a small test for the case where the provider tenant ID is not defined might be interesting. Just to see that we have a proper error message that is understandable maybe..

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DwcConfiguration.java
```diff
@@ -51,18 +51,26 @@ private DwcConfiguration()
         this.providerTenant = Lazy.of(() -> providerTenant);
     }
 
+    @Nonnull
     URI megacliteUrl()
+        throws CloudPlatformException,
+            IllegalArgumentException
```
MatKuhr, line 57: why `IllegalArgumentException`?  It is a bit weird for a method that takes no arguments to throw an `IllegalArgumentException` :D 

I guess you added this because it may be thrown somewhere down the stack? If so it would be nice to wrap it into a `CloudPlatformException` wherever it is originally coming from then.. 

But not too important, optional for a follow up PR I'd say

MatKuhr (approved):

PR #104, opened 2023-10-23
Change ScpCf modules behaviour with DwC bindings

CharlesDuboisSAP:
## Context
DwC apps cannot call `ScpCfCloudPlatform` without the `DWC_APPLICATION` env var. This was introduced in #98.

## Solution
Here is the intended stacktrace from top to bottom
- `DwcConfiguration` throws
- then `MegacliteServiceBinding` catches and returns an empty map
- then `ScpCfCloudPlatform` works normally
-  then `ScpCfDestinationServiceAdapter` throws if the map is empty.

## Questions for the Reviewer
- What should the message thrown by `ScpCfDestinationServiceAdapter` be?
  - _Maybe it should be `"No DWC_APPLICATION environment variable defined. Cannot determine the provider account id."`. Not sure if there is a scenario in which the Adapter doesn't find `tenantid` while the binding is not DwC_
- Do the new test dependencies cause a problem?
- Is there still a need for `DwcConfiguration` to be public to help customer test their app locally without setting an env var?
- Can the Exception hidden  by `MegacliteServiceBinding.getCredentials()` cause problem for customers with broken configs?

### DIFF
--- a/cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
+++ b/cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
@@ -38,6 +38,7 @@ import com.sap.cloud.sdk.cloudplatform.exception.NoServiceBindingException;
 import com.sap.cloud.sdk.cloudplatform.security.AuthToken;
 import com.sap.cloud.sdk.cloudplatform.security.AuthTokenAccessor;
 
+import io.vavr.control.Try;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
@@ -83,19 +84,20 @@ class ScpCfDestinationServiceAdapter
         if( providerTenantId != null ) {
             return providerTenantId;
         }
-        try {
-            final ServiceBinding binding = serviceBindingSupplier.get();
 
-            final TypedMapView credentials = TypedMapView.ofCredentials(binding);
-            if( credentials.containsKey("tenantid") ) {
-                return credentials.getString("tenantid");
-            }
+        final TypedMapView credentials = Try.of(() -> {
+            final ServiceBinding binding = serviceBindingSupplier.get();
+            return TypedMapView.ofCredentials(binding);
+        })
+            .getOrElseThrow(
+                e -> new DestinationAccessException(
+                    "Could not resolve destination to Destination Service on behalf of provider.",
+                    e));
+
+        if( credentials.containsKey("tenantid") ) {
+            return credentials.getString("tenantid");
         }
-        catch( final Exception e ) {
-            throw new DestinationAccessException(
-                "Could not resolve destination to Destination Service on behalf of provider.",
-                e);
-        }
+
         throw new DestinationAccessException(
             "The provider tenant id is not defined in the service binding."
                 + " Please verify that the service binding contains the field 'tenantid' in the credentials list.");
--- a/cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
+++ b/cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
@@ -86,13 +86,19 @@ class ScpCfDestinationServiceAdapter
         try {
             final ServiceBinding binding = serviceBindingSupplier.get();
 
-            return TypedMapView.ofCredentials(binding).getString("tenantid");
+            final TypedMapView credentials = TypedMapView.ofCredentials(binding);
+            if( credentials.containsKey("tenantid") ) {
+                return credentials.getString("tenantid");
+            }
         }
         catch( final Exception e ) {
             throw new DestinationAccessException(
                 "Could not resolve destination to Destination Service on behalf of provider.",
                 e);
         }
+        throw new DestinationAccessException(
+            "The provider tenant id is not defined in the service binding."
+                + " Please verify that the service binding contains the field 'tenantid' in the credentials list.");
     }
 
     private Function<OnBehalfOf, HttpDestination> prepareServiceDestinationComputation()
--- a/cloudplatform/cloudplatform-core-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/ScpCfCloudPlatform.java
+++ b/cloudplatform/cloudplatform-core-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/ScpCfCloudPlatform.java
@@ -307,6 +307,8 @@ public class ScpCfCloudPlatform implements CloudPlatform
             .accept(
                 "tags",
                 () -> serviceBinding.getTags().stream().collect(JsonArray::new, JsonArray::add, JsonArray::addAll));
+        // DwC: The binding doesn't have credentials, MegacliteServiceBinding either reads the DWC_APPLICATION env var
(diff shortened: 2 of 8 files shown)


cloudplatform/cloudplatform-core-scp-cf/src/test/java/com/sap/cloud/sdk/cloudplatform/ScpCfCloudPlatformTest.java
```diff
@@ -645,4 +649,35 @@ public void underlyingPropertiesWillBeOverwrittenInServiceBinding()
...
+
+        // setup subject under test
+        final ScpCfCloudPlatform sut = new ScpCfCloudPlatform();
+        sut.setServiceBindingAccessor(() -> Collections.singletonList(serviceBinding));
+
+        // does not throw
+        final Map<String, JsonArray> parsedVcapServices = sut.getVcapServices();
+
+        // assert
+        assertThat(parsedVcapServices).containsOnlyKeys(String.valueOf(ServiceIdentifier.DESTINATION));
```
Johannes-Schneider, line 681: [minor]
`String.valueOf` seems odd here...

```suggestion
        assertThat(parsedVcapServices).containsOnlyKeys(ServiceIdentifier.DESTINATION.toString());
```

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBinding.java
```diff
@@ -123,10 +123,12 @@ public List<String> getTags()
...
     public Map<String, Object> getCredentials()
-        throws CloudPlatformException,
-            IllegalArgumentException
     {
-        return Collections.singletonMap("tenantid", dwcConfiguration.providerTenant());
+        final Try<String> tenantId = Try.of(dwcConfiguration::providerTenant);
+        if( tenantId.isSuccess() ) {
+            return Collections.singletonMap("tenantid", tenantId.get());
+        }
+        return Collections.emptyMap();
```
newtork, line 131: (Comment)

<details><summary>Personally, in this case, I would prefer regular try-catch</summary>

```suggestion
        try {
            return Collections.singletonMap("tenantid", dwcConfiguration.providerTenant());
        }
        catch ( Exception e ) {
            log.debug("Failed to get the provider tenant ID from the DWC configuration.", e);
            return Collections.emptyMap();
        }
```

(if you'd wanted to log, the class would require `@Slf4j`)
</details>

But I'm fine with the code as is.

cloudplatform/cloudplatform-core-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/ScpCfCloudPlatform.java
```diff
@@ -307,6 +307,8 @@ private JsonObject convert( @Nonnull final ServiceBinding serviceBinding )
             .accept(
                 "tags",
                 () -> serviceBinding.getTags().stream().collect(JsonArray::new, JsonArray::add, JsonArray::addAll));
+        // DwC: The binding doesn't have credentials, MegacliteServiceBinding either reads the DWC_APPLICATION env var
+        // and returns the map {tenantid: <ProviderTenantId>}, or, returns an empty map if DWC_APPLICATION is not found.
```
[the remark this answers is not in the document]
CharlesDuboisSAP: Matthias mentioned he wanted to delete/redo this API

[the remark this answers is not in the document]
Johannes-Schneider: Sounds good 👍

I'm still wondering about this comment... It doesn't make a lot of sense to me. Why are we mentioning the DwC case here explicitly?

cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
```diff
@@ -83,16 +84,23 @@ String getProviderTenantId()
...
-        try {
+
+        final TypedMapView credentials = Try.of(() -> {
             final ServiceBinding binding = serviceBindingSupplier.get();
+            return TypedMapView.ofCredentials(binding);
+        })
+            .getOrElseThrow(
+                e -> new DestinationAccessException(
+                    "Could not resolve destination to Destination Service on behalf of provider.",
+                    e));
```
Johannes-Schneider, line 95: [minor]
Formatting is pretty odd IMO.

How about:
```suggestion
        final Try<ServiceBinding> maybeServiceBinding = Try.ofSupplier(serviceBindingSupplier);
        if( maybeServiceBinding.isFailure() ) {
            final String msg = "Could not resolve destination to Destination Service on behalf of provider.";
            throw new DestinationAccessException(msg, maybeServiceBinding.getCause());
        }

        final TypedMapView credentials = TypedMapView.ofCredentials(maybeServiceBinding.get());
```

cloudplatform/cloudplatform-core-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/ScpCfCloudPlatform.java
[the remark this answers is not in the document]
CharlesDuboisSAP: I'm just helping the next developer that will maintain this code, I can remove it.

cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
CharlesDuboisSAP: TypedMapView.ofCredentials can throw many Exceptions, I thought it was safer this way.

cloudplatform/cloudplatform-core-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/ScpCfCloudPlatform.java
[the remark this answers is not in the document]
Johannes-Schneider: Personally, I find it more confusing than it helps... But maybe I'm the only one 😅

cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationServiceAdapter.java
Johannes-Schneider: > `TypedMapView.ofCredentials` can throw many Exceptions

True... didn't consider that.
We could change the code to something like that:

```suggestion
        final Try<ServiceBinding> maybeServiceBinding = Try.ofSupplier(serviceBindingSupplier);
        if( maybeServiceBinding.isFailure() ) {
            final String msg = "Could not resolve destination to Destination Service on behalf of provider.";
            throw new DestinationAccessException(msg, maybeServiceBinding.getCause());
        }
        
        final Try<TypedMapView> maybeCredentials = Try.of(() -> TypedMapView.ofCredentials(maybeServiceBinding.get()));
        if (maybeCredentials.isFailure()) {
            final String msg = "Could not extract credentials from destination to Destination Service.";
            throw new DestinationAccessException(msg, maybeCredentials.getCause());
        }

        final TypedMapView credentials = maybeCredentials.get();
```

But I'm not sure whether it makes it more readable (maybe a tiny bit).

Johannes-Schneider (approved):
Changes LGTM

PR #106, opened 2023-10-25
feature: Introduce OAuth2 caching

cschubertcs:
Related BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/209

This PR introduces a cache in our `OAuth2ServiceImpl`.
That way the the underlying `XsuaaTokenFlows` of the security library is not re-created with every destination, allowing for usage of the cache there (more specifically: in the `AbstractOAuth2TokenService`).

Overall we have the following instance hierarchy:

`DefaultOAuth2TokenService`/`AbstractOAuth2TokenService` (security lib, caches the http response, so the token received)
vv -- **Gets called by** -- vv
`XsuaaTokenFlows` (security lib)
vv -- **Gets called by** -- vv
`OAuth2ServiceImpl` (SDK, cache introduced by this PR)
vv -- **Gets called by** -- vv
`OAuth2DestinationBuilder` and `OAuth2ServiceBindingDestinationLoader` (both SDK, and both create a new `OAuth2ServiceImpl` for every destination they create, no caching here)

The newly introduced cache now allows for reusing of the underlying `XsuaaTokenFlows` and with that the even deeper response cache.

In addition this PR also takes care, that the `HttpClient` used to retrieve the token in `AbstractOAuth2TokenService` is not sending cookies between different tenants.
For that, the cache in `OAuth2ServiceImpl` also takes the zone id into consideration.


cschubertcs:
Still pending some documentation in form of javadoc as "ADR".

cschubertcs:
Can now be fully reviewed

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OnPremTest.java
```diff
@@ -0,0 +1,153 @@
...
+import com.github.tomakehurst.wiremock.stubbing.ServeEvent;
+import com.sap.cloud.sdk.cloudplatform.resilience.ResilienceConfiguration;
+import com.sap.cloud.sdk.cloudplatform.tenant.DefaultTenant;
+import com.sap.cloud.sdk.cloudplatform.tenant.DefaultTenantFacade;
+import com.sap.cloud.sdk.cloudplatform.tenant.TenantAccessor;
+import com.sap.cloud.security.config.ClientCredentials;
+import com.sap.cloud.security.config.ClientIdentity;
+
+public class OnPremTest
+{
```
CharlesDuboisSAP, line 30: This tests separation only from `/oauth/token` and not `authorizeEndpoint` and `jwksUri`.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceImpl.java
```diff
@@ -42,20 +45,58 @@
...
     {
-        this.tokenFlows = tokenFlows;
+        this.uri = uri;
+        this.identity = identity;
     }
 
     static OAuth2ServiceImpl fromCredentials( final String uri, final ClientIdentity identity )
+    {
+        return new OAuth2ServiceImpl(uri, identity);
+    }
```
CharlesDuboisSAP, line 78: A new `OAuth2ServiceImpl` is still created for every destination. Couldn't the cache be one level higher at `OAuth2ServiceBindingDestinationLoader.createHeaderProvider`?

cschubertcs: Hm, I think there would not be a lot of benefit of caching the `OAuth2ServiceImpl`, as there is not a lot of state connected to it or created (besides the now cached `XsuaaTokenFlows`). The only thing I would spontaneously see would be a minor memory efficiency win.
Also, the `OAuth2ServiceImpl` gets used by `OAuth2DestinationBuilder` and `OAuth2ServiceBindingDestinationLoader`, meaning we would need to cache it in both of those classes.

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OnPremTest.java
cschubertcs: Hm, true, but I don't have a spontaneous good idea on how to test those, as this is used under the hood by the XsuaaTokenFlows, not by us, to retrieve the actual token we mock in here.
The most important part is actually the `baseUri`, and that is the same for all three of them.

I will have a look tomorrow to see some reasonable way to check those as well.

cschubertcs: I don't see any way to test this:
- The `Endpoints` class itself is simple implementation of the `OAuth2ServiceEndpointsProvider` interface
- The evaluation of the interface methods happens inside the security library. The tests only mock the usual OAuth flow, which happens to use the `/oauth/token` endpoint.
- The only parameter of `Endpoints` is the `baseUri`, which we could use alternatively as a key. We just decided against it, in case that we change anything in the implementation of the `Endpoints` class.

All in all I don't see a way and also no benefit in testing these other properties.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceImpl.java
```diff
@@ -42,20 +45,58 @@
...
+     * The cache key is composed of the following parts:
+     * <ul>
+     * <li>{@code zoneId}, that way flow is tenant isolated, especially for the HttpClient used against the OAuth2
+     * service.</li>
+     * <li>{@code ClientIdentity}, to separate by the credentials used against the OAuth2 service.</li>
+     * <li>{@code Endpoints}, to separate by different OAuth2 services.</li>
+     * </ul>
+     */
+    private static final Cache<CacheKey, XsuaaTokenFlows> tokenFlowCache =
+        Caffeine.newBuilder().expireAfterAccess(1, TimeUnit.HOURS).build();
```
newtork, line 65: (Major)

<details>
<summary>I think we should leverage the caching capabilities from XSUAA security library</summary>

image

</details>

Instead of cached lookup:
```
String, OAuth2ServiceEndpointsProvider, ClientIdentity -> XsuaaTokenFlows
```
You could just cache-lookup:
```
ClientIdentity -> DefaultOAuth2TokenService
```

The rest should work automatically, As far as I understood, `XsuaaTokenFlows` can be instantiated with every call, it's immutable and low-footprint.

cschubertcs: Hm, that kind of makes sense, I will try it out!

We are currently already leveraging the response cache in the `AbstractOAuth2TokenService`, but the way you propose looks much more manageable and less error prone.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceImpl.java
```diff
@@ -42,20 +46,49 @@
...
+     * The {@code OAuth2ServiceImpl} is newly instantiated by {@code OAuth2DestinationBuilder} and
+     * {@code OAuth2ServiceBindingDestinationLoader} for each destination they build/load. This means, without the
+     * cache, also new {@code OAuth2TokenService} would be created for each destination, which in turns defeats the
+     * purpose of the response cache used therein.
+     * <p>
+     * The cache key is composed of the following parts:
+     * <ul>
+     * <li>{@code zoneId}, that way flow is tenant isolated, especially for the HttpClient used against the OAuth2
+     * service.</li>
+     * <li>{@code ClientIdentity}, to separate by the credentials used against the OAuth2 service.</li>
```
Johannes-Schneider, line 61: I'm wondering:
Since we are no longer using the actual token endpoint URI for our cache key, chances are (even if almost zero) that  we do have two different token services that do use the same `ClientIdentity` (i.e. same `clientid`, `clientsecret).

As far as I could see, this shouldn't be an issue as the XSUAA lib already uses the actual token endpoint uri for their internal cache.
Nevertheless, I think it would be nice to be rather sure by introducing a corresponding unit test.
That way, we will also notice in case the XSUAA lib ever decides to change their internal implementation for whatever reason.

cschubertcs: Makes sense, test added!

newtork (changes requested):
Let me check whether we can move "behalf" evaluation from lazy to eager.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceImpl.java
```diff
@@ -42,20 +46,49 @@
...
+        return new OAuth2ServiceImpl(uri, identity);
+    }
+
+    static void clearCache()
+    {
+        log.warn("Resetting the TokenService cache. This should not be done outside of testing.");
+        tokenServiceCache.invalidateAll();
+    }
+
+    XsuaaTokenFlows createTokenFlow( final String zoneId )
```
newtork, line 86: (Major)

To avoid future confusion, this method does not "create a token flow" but it "creates a token flow factory" 😁 

Maybe...
```suggestion
    XsuaaTokenFlows getTokenFlows( final String zoneId )
```

newtork (approved):

PR #112, opened 2023-11-01
feat: Add error handling for incorrect destination configuration in HttpClientWrapper

CharlesDuboisSAP:
# Pull Request Description 🤖

This PR addresses the backlog item [#320](https://github.com/SAP/cloud-sdk-java-backlog/issues/320) and introduces a new feature to the `HttpClientWrapper` class in the `connectivity-apache-httpclient4` module.

## Changes

The changes in this PR include:

- A new block of code in the `HttpClientWrapper` constructor that checks if the destination is configured correctly. If the destination is not configured correctly, it throws a `DestinationAccessException` with an appropriate error message.

The new block of code checks if the destination's `PROXY_TYPE` is `ON_PREMISE` and if the `PROXY_HOST` is empty. If these conditions are met, it further checks if the `TENANT_ID` is empty (indicating a provider) and if the `AUTH_TYPE` (or `AUTH_TYPE_FALLBACK`) is `BASIC_AUTHENTICATION`. If these conditions are met, it throws a `DestinationAccessException` with the message "Accessing an onpremise system on behalf of the provider tenant explicitly is currently not supported." Otherwise, it throws a `DestinationAccessException` with the message "Unable to resolve connectivity service binding."

## Context

This change is part of the effort to migrate the DwC Spring E2E Test, as described in the backlog item. The new feature ensures that the destination is configured correctly before attempting to establish a connection, which helps prevent errors during runtime.

## Reviewer Instructions

Please review the changes in the context of the associated backlog item. Pay particular attention to the new error handling code and ensure that it aligns with the expected behavior described in the backlog item.


cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
```diff
@@ -76,6 +77,23 @@ public String toString()
...
+                destination.get(DestinationProperty.TENANT_ID).filter(String::isEmpty).isDefined();
+            final boolean isBasicAuth =
+                destination
+                    .get(DestinationProperty.AUTH_TYPE)
+                    .orElse(() -> destination.get(DestinationProperty.AUTH_TYPE_FALLBACK))
+                    .contains(AuthenticationType.BASIC_AUTHENTICATION);
+            if( isProvider && isBasicAuth ) {
+                throw new DestinationAccessException(
+                    "Accessing an onpremise system on behalf of the provider tenant explicitly is currently not supported.");
+            }
```
newtork, line 94: Afaik this limitation is only applicable for DwC not in general.
```suggestion
```

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
```diff
@@ -76,6 +77,23 @@ public String toString()
...
+            throw new DestinationAccessException("Unable to resolve connectivity service binding.");
```
newtork, line 95: (Minor)

Maybe this is a good place to encourage looking into the logs?

```suggestion
            throw new DestinationAccessException("Unable to resolve connectivity service binding. Please check the logs.");
```

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
CharlesDuboisSAP: Removed.

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
CharlesDuboisSAP: I stored the failed proxy destination and re-throwing it in the HttpClientWrapper. Tell me what you think.

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
```diff
@@ -76,6 +76,14 @@ public String toString()
     HttpClientWrapper( final CloseableHttpClient httpClient, final HttpDestinationProperties destination )
     {
         this.httpClient = httpClient;
+
+        // check destination is configured correctly
+        if( destination.get(DestinationProperty.PROXY_TYPE).contains(ProxyType.ON_PREMISE)
+            && destination.get(DestinationProperty.PROXY_HOST).isEmpty() ) {
+
+            // throw the original error
+            new DefaultHttpDestinationBuilderProxyHandler().handle(DefaultHttpDestination.fromDestination(destination));
```
Johannes-Schneider, line 85: Personally, I feel like this change isn't a good solution.

Not only does it rely on internal implementation details that might change over time and shouldn't be used outside the `DefaultHttpDestinationBuilder`, but it also isn't guaranteed to actually re-produce the original exception (e.g. the Service Bindings might have changed since we initially created the destination, so this call here might very well succeed).

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -949,9 +949,9 @@ public DefaultHttpDestination build()
 
             // handle proxy type == OnPremise
             if( builder.get(DestinationProperty.PROXY_TYPE).contains(ProxyType.ON_PREMISE) ) {
-                final DefaultHttpDestination proxyDestination = proxyHandler.handle(this);
-                if( proxyDestination != null ) {
-                    return proxyDestination;
+                final Try<DefaultHttpDestination> proxyDestination = Try.of(() -> proxyHandler.handle(this));
+                if( proxyDestination.isSuccess() ) {
+                    return proxyDestination.get();
```
Johannes-Schneider, line 954: It feels pretty wrong to silently ignore any exception that might occur here... customers can't even refer to the application logs to see what might have gone wrong.

Johannes-Schneider (commented):
I do have some concerns about the suggested solution (see comments).

Furthermore, personally, I think it's a pretty big design flaw to not have the two concerns of **(1)** _I want to connect to a remote system_ and **(2)** _I want to see how my destination in the Destination Service is configured_ (i.e. I just need the properties) separated.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
CharlesDuboisSAP: I added a log here.

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
CharlesDuboisSAP: I added a throw just below just in case the service binding has changed. The destination shouldn't be used anyway.

CharlesDuboisSAP:
I agree, the two different use cases (reading properties and using the destination) for the same class is a design flaw. But with the timeframe that we have to release v4, I would accept it as tech debt and have usable exceptions messages.
@Johannes-Schneider Do you think the latest commit is better than storing the exception?

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -949,9 +949,16 @@ public DefaultHttpDestination build()
...
+                // Any change here must be copied to the HttpClientWrapper constructor
+                final Try<DefaultHttpDestination> proxyDestination = Try.of(() -> proxyHandler.handle(this));
+                if( proxyDestination.isSuccess() ) {
+                    return proxyDestination.get();
+                }
+                if( proxyDestination.isFailure() ) {
+                    log
+                        .error(
+                            "Unable to resolve destination to connectivity service binding. This destination cannot be used other than reading its properties.",
+                            proxyDestination.getCause());
```
Johannes-Schneider, line 961: I think the details should be included in the cause. So let's not jump to conclusions on this higher abstraction level.

```suggestion
                    log
                        .error(
                            "Unable to resolve proxy details for destination. This destination cannot be used for anything other than reading its properties.",
                            proxyDestination.getCause());
```

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java
```diff
@@ -76,6 +77,18 @@ public String toString()
...
+        // Any change here must be copied to DefaultHttpDestination.Builder.build()
+        if( destination.get(DestinationProperty.PROXY_TYPE).contains(ProxyType.ON_PREMISE)
+            && destination.get(DestinationProperty.PROXY_HOST).isEmpty() ) {
+
+            // throw the original error
+            new DefaultHttpDestinationBuilderProxyHandler().handle(DefaultHttpDestination.fromDestination(destination));
+            // just in case the service binding has changed and the exception does not occur anymore
+            throw new DestinationAccessException(
+                "Unable to resolve connectivity service binding. Please check the logs.");
+        }
```
Johannes-Schneider, line 91: Let's stick what we've aligned on:

```suggestion

        if( destination.getProxyType().contains(ProxyType.ON_PREMISE)
            && destination.getProxyConfiguration().isEmpty() ) {
            throw new DestinationAccessException(
                "Unable to create an HttpClient from the provided destination. "
                    + "The destination is supposed to target an on-premise system but lacks the correct proxy configuration. "
                    + "Please check the application logs for further details.");
        }
```

---

Also: Please apply the same change to the `HttpClient5Wrapper`.

Johannes-Schneider (changes requested):

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -949,9 +949,16 @@ public DefaultHttpDestination build()
...
+                // Any change here must be copied to the HttpClientWrapper constructor
+                final Try<DefaultHttpDestination> proxyDestination = Try.of(() -> proxyHandler.handle(this));
+                if( proxyDestination.isSuccess() ) {
+                    return proxyDestination.get();
+                }
+                if( proxyDestination.isFailure() ) {
+                    log
+                        .error(
+                            "Unable to resolve proxy configuration for destination. This destination cannot be used for anything other than reading its properties.",
+                            proxyDestination.getCause());
```
newtork, line 961: Unnecessary if condition:

```suggestion
                final String msg =
                    "Unable to resolve proxy configuration for destination. This destination cannot be used for anything other than reading its properties.";
                log.error(msg, proxyDestination.getCause());
```

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -949,9 +949,16 @@ public DefaultHttpDestination build()
...
                 }
```
newtork, line 962: Unnecessary if condition:

```suggestion
                final String msg =
                    "Unable to resolve proxy configuration for destination. This destination cannot be used for anything other than reading its properties.";
                log.error(msg, proxyDestination.getCause());
```

The rest looks good to me :+1:

newtork: (Comment)

<details>
<summary>The code block could've also just be using regular try-catch for potentially-better(?) readability 🤔</summary>

```suggestion
                try {
                    return proxyHandler.handle(this);
                }
                catch( final Exception e ) {
                    final String msg =
                        "Unable to resolve proxy configuration for destination. This destination cannot be used for anything other than reading its properties.";
                    log.error(msg, e);
                }
```

</details>

(No action required)

newtork (approved):

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
CharlesDuboisSAP: Applied suggestion

newtork (dismissed):

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -949,9 +949,14 @@ public DefaultHttpDestination build()
 
             // handle proxy type == OnPremise
             if( builder.get(DestinationProperty.PROXY_TYPE).contains(ProxyType.ON_PREMISE) ) {
-                final DefaultHttpDestination proxyDestination = proxyHandler.handle(this);
-                if( proxyDestination != null ) {
-                    return proxyDestination;
+                // Any change here must be copied to HttpClientWrapper(s) constructors
```
Johannes-Schneider, line 952: That comment is no longer true.

```suggestion
```

rnewbigging (approved):

PR #113, opened 2023-11-02
Fix: Silently Ignore Resilience for multiple `ResilienceDecorationStrategies`

Johannes-Schneider:
## Context

BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/337

This change is ported over from our `main/v4` fix: https://github.wdf.sap.corp/MA/sdk/pull/8633

**Note:** No release notes needed as the issue will also be fixed in V4 and, therefore, nothing has changed in our V5 release from the consumers' perspective.


Johannes-Schneider:
Continued in #115

PR #115, opened 2023-11-02
Fix: Silently Ignore Resilience for multiple ResilienceDecorationStrategies - Alternative

Johannes-Schneider:
## Context

BLI: https://github.com/SAP/cloud-sdk-java-backlog/issues/337
Alternative to #113


newtork (approved):
I think I like this solution more.

PR #124, opened 2023-11-08
chore: re-enable blackduck nightly job

jjtang1985:
This PR enables the nightly job of the blackduck.
Let's evaluate this tooling later.


CharlesDuboisSAP (commented):
A nightly scan has no purpose if unreliable. I would add this flag remove any copyright findings from the scan:
>            --excludedDirectories=src \

newtork:
> An unreliable nightly scan has no purpose.

Is this because of false-positive findings of SDK code being reported?

> this flag to remove all copyright findings

I would advice against doing that, since we mentioned (accidental) code theft as one of the main motivations for BlackDuck. Without it, we only have vulnerability- and license checks for the dependencies, for which we don't necessarily need BlackDuck.

CharlesDuboisSAP (approved):
I give up

MatKuhr (approved):

PR #127, opened 2023-11-08
Reduce scope of `com.sap.cloud.security:java-api` to `provided`

newtork:
## Context
 

The dependency doesn't necessarily need to be at runtime for all cases.
It's only required IF the `SecurityContext` class is available at all.

## Definition of Done

- [ ] E2E tested
- [x] ~Documentation updated~
- [x] ~Release notes updated~


cloudplatform/security/pom.xml
```diff
@@ -70,6 +66,11 @@
 			<artifactId>httpcore</artifactId>
 		</dependency>
 		<!-- scope "provided" -->
+		<dependency>
+			<groupId>com.sap.cloud.security</groupId>
+			<artifactId>java-api</artifactId>
+			<scope>provided</scope>
```
MatKuhr, line 72: (minor) alternatively we could use `optional` to indicate this module also works without this dependency

cloudplatform/security/src/main/java/com/sap/cloud/sdk/cloudplatform/security/SecurityContextThreadContextDecorator.java
```diff
@@ -31,9 +30,18 @@ public class SecurityContextThreadContextDecorator implements ThreadContextDecor
     @Override
     public <T> Callable<T> decorateCallable( @Nonnull final Callable<T> callable )
     {
+        try {
+            getClass().getClassLoader().loadClass("com.sap.cloud.security.token.SecurityContext");
+        }
+        catch( final Exception e ) {
```
MatKuhr, line 36:
```suggestion
        catch( final ClassNotFoundException e ) {
```

Also, apparently one can [skip instantiation of the class](https://www.baeldung.com/java-check-class-exists#telling-classforname-to-skip-initialization).

MatKuhr (approved):
pending e2e testing

newtork:
Blocked until E2E test runs again.

MatKuhr (approved):
Manually E2E tested with the JCo E2E test, works as expected 👍🏻

PR #128, opened 2023-11-08
chore: Test Additions and Improvements for Various Connectivity Classes

newtork:
Minor syntax and test improvements


cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ScpCfDestinationLoader.java
```diff
@@ -110,21 +106,6 @@ public ScpCfDestinationLoader()
...
 
-    /**
-     * Create instance with a specific TimeLimiter timeout {@link Duration}.
-     *
-     * @param timeLimiterTimeout
-     *            The maximum {@link Duration} a request may take before the TimeLimiter interrupts the request.
-     * @deprecated Please use {@link Builder#withTimeLimiterConfiguration} to supply a time limiter configuration.
-     */
-    @Deprecated
-    public ScpCfDestinationLoader( @Nonnull final Duration timeLimiterTimeout )
```
newtork, line 121: (Comment)

Removing unused deprecated constructor.

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoaderTest.java
```diff
@@ -94,6 +94,15 @@ public void testCustomServiceCanBeRegistered()
         assertThat(DEFAULT_SERVICE_RESOLVERS).anyMatch(f -> f.matches(OPTIONS_WITH_EMPTY_BINDING));
     }
 
+    @Test
+    public void testCustomServiceCanBeRegisteredLegacy()
+    {
+        OAuth2ServiceBindingDestinationLoader
+            .registerPropertySupplier(TEST_SERVICE, DefaultOAuth2PropertySupplier::new);
```
newtork, line 101: (Explanation)

Same test case as above, but start testing public `.registerPropertySupplier( ServiceIdentifier. Function)`

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultServiceBindingDestinationLoaderChain.java
```diff
@@ -46,18 +45,14 @@
...
-        DEFAULT_DELEGATE_LOADERS =
-            FacadeLocator.getFacades(ServiceBindingDestinationLoader.class).stream().collect(toList());
-        DEFAULT_INSTANCE = new DefaultServiceBindingDestinationLoaderChain(DEFAULT_DELEGATE_LOADERS);
-    }
+    static final DefaultServiceBindingDestinationLoaderChain DEFAULT_INSTANCE =
+        new DefaultServiceBindingDestinationLoaderChain(DEFAULT_DELEGATE_LOADERS);
 
     @Nonnull
-    private final List<ServiceBindingDestinationLoader> delegateLoaders;
+    private final Collection<ServiceBindingDestinationLoader> delegateLoaders;
```
Johannes-Schneider, line 55: I'm not exactly happy with these changes.
This is because `List` guarantees a specific order of the elements whereas `Collection` doesn't provide that guarantee.

In our case here, having an order is actually important as the first successful delegate will win.
Therefore, customers may want to prepend their custom `ServiceBindingDestinationLoader` to handle some specific edge case.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ServiceBindingDestinationOptions.java
```diff
@@ -204,6 +204,7 @@ public ServiceBindingDestinationOptions build()
      *            The type of content that this option will represent.
      * @since 4.20.0
      */
+    @FunctionalInterface
```
Johannes-Schneider, line 207: We explicitly DIDN'T add this annotation when we created this interface.
We even have a pretty weird check to make sure nobody passes a lambda into our API:

```java
        @Nonnull
        public Builder withOption( @Nonnull final OptionsEnhancer<?> option )
        {
            final Class<?> cls = option.getClass();
            if( cls.isAnonymousClass() || cls.isSynthetic() ) {
                throw new IllegalArgumentException("Lambdas or anonymous classes are not supported here.");
            }
            if( furtherOptions.containsKey(cls) ) {
                throw new IllegalArgumentException("The given option " + cls.getSimpleName() + " is already in use.");
            }
            furtherOptions.put(cls, option);
            return this;
        }
```

This is because the concrete `OptionEnhancer` instances are stored by their `class` in our internal `Map`.
Hereby, each lambda instance would get it's own anonymous class, so that getting the actual enhancer would only be possible if you stored the lambda reference to being with (which seems highly unlikely).
So to prevent customers from thinking that providing a lambda might be a good idea, we refrained from using the `@FunctionalInterface` annotation (even though it doesn't really make a difference for the actual usage, but still...)

newtork: Reverted productive code change.

cloudplatform/connectivity-apache-httpclient4/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientCacheTest.java
```diff
@@ -0,0 +1,31 @@
...
+package com.sap.cloud.sdk.cloudplatform.connectivity;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import org.junit.jupiter.api.Test;
+
+class HttpClientCacheTest
```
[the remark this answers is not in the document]
newtork: It's a unit test about the interface "HttpClientCache", the `DISABLED` constant respectively.

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoaderTest.java
```diff
@@ -94,6 +94,15 @@ void testCustomServiceCanBeRegistered()
         assertThat(DEFAULT_SERVICE_RESOLVERS).anyMatch(f -> f.matches(OPTIONS_WITH_EMPTY_BINDING));
     }
 
+    @Test
+    void testCustomServiceCanBeRegisteredLegacy()
```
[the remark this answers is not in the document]
newtork: _Legacy_ in the sense of "old" API usage, based on `ServiceIdentifier`.
Most likely users will prefer the `Predicate` method alternative, e.g. CAP.

PR #129, opened 2023-11-08
chore: Rename ScpCf prefixed classes

cschubertcs:
SAP/cloud-sdk-java-backlog#346.

This PR renames the classes prefixed with `ScpCf...`. All of them where in the context of the `ScpCfDestinationLoader` and got renamed to `DestinationService...`.


cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationRetrievalStrategyResolver.java
```diff
@@ -193,8 +193,8 @@ DestinationRetrieval prepareSupplier(
     }
 
     private void warnOrThrowOnDeprecatedOrUnsupportedCombinations(
-        @Nonnull final ScpCfDestinationRetrievalStrategy retrievalStrategy,
-        @Nullable final ScpCfDestinationTokenExchangeStrategy tokenExchangeStrategy )
+        @Nonnull final DestinationServiceStrategy retrievalStrategy,
```
newtork, line 196: (Major)

This name change appears inconsistent to me.

```
ScpCfDestinationRetrievalStrategy -> DestinationServiceStrategy
ScpCfDestinationTokenExchangeStrategy .> DestinationTokenExchangeStrategy
```

cschubertcs: Fair, must have missed those/renamed them wrong. Fixed!

newtork (approved):
LGTM
Please consider adding release notes to announce class renamings, either here or in a separate PR.

MatKuhr (commented):
I think we should also rename the Maven module 😉
Maybe `connectivity-destination-service`?

newtork:
> I think we should also rename the Maven module

You really want to put up with dependency conflicts on customer side accidentally mixing v4+v5?

MatKuhr:
> > I think we should also rename the Maven module
> 
> You really want to put up with dependency conflicts on customer side accidentally mixing v4+v5?

Yeah it is a tradeoff.. This is again a module that uses class loading so having it on the class path twice would actually be bad. We could explicitly check if the `ScpCfDestinationLoader` class exists and throw if it does, not sure where to add this check though..

On the other hand, it would also be some migration effort for customers. A quick search reveals around 200 usages in total. So maybe not worth it? But if we already rename the core classes then renaming the module accordingly kinda makes sense..

How about this: We create a new module `connectivity-destination-service` but also keep the old module, make it empty and simply reference the new module as dependency. Then we dodge the version conflict and customers can migrate slowly if they like.

cschubertcs:
Okay, I will have a look into the 2 module solution, sounds good to me!

cschubertcs:
For reference here as well:

> Current usages of `ScpCfDestinationLoader`:
> [github.wdf.sap.corp](https://github.wdf.sap.corp/search?l=Java&q=ScpCfDestinationLoader&type=Code): 182
> [github.tools.sap](https://github.tools.sap/search?l=Java&q=ScpCfDestinationLoader&type=Code): 176
> [github.com](https://github.com/search?q=ScpCfDestinationLoader+language%3AJava&type=code&l=Java): 14

cschubertcs:
@MatKuhr : Do we want to wait for the meeting on monday to merge this, or get it done now and "fix" (whatever we decide on) after the meeting?

MatKuhr:
I think we can merge now, for the meeting on Tuesday I'd rather discuss the "aggregator" module strategy and only revert this change if we come to drastically different conclusions there.

newtork (approved):

PR #130, opened 2023-11-09
chore: Update XSUAA Client Lib Major (and Transitive Dependencies) Version

Johannes-Schneider:
This PR updates following dependencies and build prerequisites:

* XSUAA Client Lib: `2.14.2` → `3.2.1`
* Java: `1.8` → `17`
* ~Maven: `3.5` → `3.9`~
* Spring Bom (internal dependency only): `5.3.29` → `6.0.13`
* Wiremock (tests only): `2.35.1` → `3.0.1`
* Jetty (tests only): `9.4.53.v20231009` → `11.0.18`


bom/pom.xml
```diff
@@ -51,7 +51,6 @@
 		<!-- Keep these two versions consistent with the one from: com.sap.cloud.security.xsuaa:parent -->
 		<!-- see https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/pom.xml -->
 		<scp-cf.xsuaa-client.version>2.14.2</scp-cf.xsuaa-client.version>
-		<spring.version>5.3.29</spring.version>
```
Johannes-Schneider, line 54: Property is never used.

pom.xml
```diff
@@ -93,9 +93,7 @@
 		<!--  XML used once in SOAP payload deserialisation  -->
 		<!--  End of non-essential dependencies  -->
 		<!--  Fully internal (test) dependencies  -->
-		<!-- Keep Spring version in sync with bom/pom.xml -->
-		<spring.version>5.3.29</spring.version>
-		<spring-security.version>5.8.6</spring-security.version>
```
Johannes-Schneider, line 98: Property is never used.

MatKuhr (commented):
We need to update XSUAA to v3 as well, right? 
Also, we need to adjust our `scp-cf` to include Jakarta servlet by default?

Or am I missing something here?

Johannes-Schneider:
> We need to update XSUAA to v3 as well, right?

At first I was hoping that we could be doing that update independently, but obviously that is not true as XSUAA v3 requires Spring 6... So yeah, I'll include the update in this PR as well.

> Also, we need to adjust our `scp-cf` to include Jakarta servlet by default?

Yep, good catch. Will do that as well.

Johannes-Schneider:
Continued in #142

PR #131, opened 2023-11-09
chore: update caffeine major version

Johannes-Schneider:
This PR updates the Caffeine version to `3.1.8` (from `2.9.3`).


newtork (approved):
Of course, there's a risk for transitive dependency conflicts. Let's see in E2E tests whether this works out.

PR #132, opened 2023-11-09
chore: `dwc-cf` includes `scp-cf`

CharlesDuboisSAP:
# Pull Request Description

This PR Closes SAP/cloud-sdk-java-backlog#349

## Context

- Investigation: should we keep `dwc-cf` and `scp-cf`?
  - We need to keep `scp-cf` and not rename it because everyone uses this dependency.
  - We also need to keep `dwc-cf` as DwC customers need many individual dependencies such as `security` and `tenant` and `cloudplatform-connectivity-scp-cf`.
    - In fact `dwc-cf` could even contain `scp-cf` to include all those aforementioned dependencies.

## Reviewer Notes

Please ensure that the removal of the `security` dependency and the replacement of the `tenant` dependency with `scp-cf` does not affect any existing functionality or dependencies.


newtork (approved):
lgtm

note: `dwc-cf` and `scp-cf` may be subject to change again

PR #135, opened 2023-11-09
feat: SDK-Core Dependency Bundle

MatKuhr:
## Context Information

Fixes SAP/cloud-sdk-java-backlog#349

This PR introduces `sdk-core` as a replacement for the `scp-cf` convenience module.

Note: This PR lists the new module in a separate directory `dependency-bundles` for better clarity. We could move the other artifacts `sdk-bom` and `sdk-modules-bom` also under this directory in a follow-up PR.


dependency-bundles/sdk-core/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+			<artifactId>connectivity-oauth</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+			<artifactId>connectivity-apache-httpclient4</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+			<artifactId>connectivity-apache-httpclient5</artifactId>
+		</dependency>
```
Johannes-Schneider, line 58: This specific dependency is somewhat questionable, right?
Personally, I'd say lets be a bit more conservative and do not include it here... If we see many customers that requires this dependency, we could always just add it later on.

Johannes-Schneider (commented):
Generally LGTM - just one "question" about the apache http client 5.

Johannes-Schneider (approved):
LGTM

PR #149, opened 2023-11-20
fix: DwC Connectivity Service Auto Configuration

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#209

This PR changes the connectivity service binding for megaclite to no longer be registered by default. This is necessary as otherwise there would be two bindings available in case shared orbits are used.

Users will have to register the connectivity binding explicitly, similar to the destination service binding.

### Feature scope:
 
- [x] Make `CONNECTIVITY_BINDING` public
- [ ] E2E Test update
- [x] Update DwC Documentation & Migration Guide

## Definition of Done

- [x] Documentation updated
- [x] ~Release notes updated~

### DIFF
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBindingAccessor.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBindingAccessor.java
@@ -34,23 +34,22 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class MegacliteServiceBindingAccessor implements ServiceBindingAccessor
 {
-    // this binding just exists so we can use ServiceIdentifier.CONNECTIVITY in our
-    // service binding -> destination APIs
-    // its content is currently not used -- see https://github.com/SAP/cloud-sdk-java-backlog/issues/209
-    static final MegacliteServiceBinding CONNECTIVITY_BINDING =
+    /**
+     * The {@link MegacliteServiceBinding} for the connectivity service. In case you want to access on-premise systems
+     * via Megaclite, you need to register this service binding manually using
+     * {@link #registerServiceBinding(MegacliteServiceBinding)}.
+     */
+    @Nonnull
+    public static final MegacliteServiceBinding CONNECTIVITY_BINDING =
         MegacliteServiceBinding
             .forService(ServiceIdentifier.CONNECTIVITY)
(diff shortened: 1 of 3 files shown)


cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBindingAccessor.java
```diff
@@ -34,24 +34,23 @@
...
-    // service binding -> destination APIs
-    // its content is currently not used -- see https://github.com/SAP/cloud-sdk-java-backlog/issues/209
-    static final MegacliteServiceBinding CONNECTIVITY_BINDING =
+    /**
+     * The {@link MegacliteServiceBinding} for the connectivity service. In case you want to access on-premise systems
+     * via Megaclite, you need to register this service binding manually using
+     * {@link #registerServiceBinding(MegacliteServiceBinding)}.
+     */
+    @Nonnull
+    public static final MegacliteServiceBinding CONNECTIVITY_BINDING =
```
newtork, line 43: (Major)

I think this behavior change requires a proper release, doesn't it?

MatKuhr: Do you mean release note? Since this only affects DwC users I was thinking to have this in the DwC docs only. We did it similarly for other DwC related API changes IIRC.

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteServiceBindingAccessor.java
```diff
@@ -34,24 +34,23 @@
...
+            .name("connectivity")
+            .version("v1")
             .build();
 
     @Nonnull
     private static final Set<MegacliteServiceBinding> serviceBindings = new HashSet<>();
 
-    static {
-        registerServiceBinding(CONNECTIVITY_BINDING);
-    }
```
Johannes-Schneider, line 53: We are using the same code again in the `clearServiceBindings` method (line `67`).
I'd say let's also remove it from there.

Johannes-Schneider (changes requested):

Johannes-Schneider (approved):
Changes LGTM - I think we should start something like a `release_notes_dwc_next_snapshot.md` to keep track of (future) changes.

PR #151, opened 2023-11-21
fix: CAP API Compatibility

MatKuhr:
## Context

This is effectively the same fix that we did to the `HttpClientAccessor`. The API is used by the CAP multi tenancy implementation, both in cds-services and in the DwC Spring Utils.

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultDestinationLoader.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultDestinationLoader.java
@@ -42,6 +42,23 @@ public final class DefaultDestinationLoader implements DestinationLoader
         return this;
     }
 
+    /**
+     * Register a {@link HttpDestination} to this destination loader. It must have a non-empty name to be identified by.
+     * If a destination with the same name was registered previously it will be replaced by this one.
+     *
+     * @param destination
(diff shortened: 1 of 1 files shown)


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultDestinationLoader.java
```diff
@@ -42,6 +42,23 @@ public DefaultDestinationLoader registerDestination( @Nonnull final Destination
...
+     *
+     * @param destination
+     *            A destination this loader should return when queried.
+     * @return This loader instance.
+     *
+     * @throws IllegalArgumentException
+     *             if the destination name is null or empty.
+     */
+    @Nonnull
+    public DefaultDestinationLoader registerDestination( @Nonnull final HttpDestination destination )
```
Johannes-Schneider, line 57: Should we declare these methods as `@Deprecated` (also affects the `HttpClientAccessor`)

Johannes-Schneider: Also: Is there any way we can check for further compatibility issues?
(Idea: What about migrating the DwC CALM Test App? Is that one using CAP by any chance?)

Johannes-Schneider (approved):

PR #157, opened 2023-11-23
PoC use Bulkhead with Size 1 for Double-Checked-Locking

MatKuhr:


cloudplatform/connectivity-dwc/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/MegacliteConnectivityProxyInformationResolverTest.java
```diff
@@ -136,48 +138,35 @@ void testProxyAuthHeader()
     @Timeout( value = 300_000, unit = java.util.concurrent.TimeUnit.MILLISECONDS )
     void testProxyUrlRetrievalIsLockedGlobally()
     {
-        doReturn(successResponse).when(sut).makeHttpRequest(any());
+        doAnswer((i) -> {
+            Thread.sleep(5_000);
```
MatKuhr, line 142: this is in place to make it likely the first thread will be paused and the second runs into the bulkhead. Not so easy to assert the same with latches, because we would n

Johannes-Schneider:
Implemented #162

PR #158, opened 2023-11-23
General Cleanup of Leftovers

MatKuhr:
## Context

This PR does some housekeeping across the project.

### Feature scope:
 
- [x] Remove any deprecated API 
- [x] Remove `@Beta` annotations on API that is unlikely to change and frequently used by users
- [x] Dependency improvements & cleanup
- [x] Minor renaming of OAuth related classes


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationProperty.java
```diff
@@ -23,7 +23,6 @@
  *
  * @since 4.3.0
  */
-@Beta
```
newtork, line 26: (Comment)

This API is a bit awkward: users are interested in instances of `DestinationPropertyKey`. For that they look into utility class `DestinationProperty`. Neither type has public constructors, making one wonder why it isn't an enum then. 

However I also don't see this API being changed for v5. So it's fine then.

cloudplatform/cloudplatform-core/src/main/java/com/sap/cloud/sdk/cloudplatform/requestheader/RequestHeaderAccessor.java
```diff
@@ -47,6 +46,7 @@ public final class RequestHeaderAccessor
      *            The {@link RequestHeaderFacade} to use, or {@code null} if the default {@link RequestHeaderFacade}
      *            should be used.
      */
+    @Beta
```
newtork, line 49: (Question)

What is your concern with `RequestHeaderFacade` for stabilizing?

cloudplatform/connectivity-dwc/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OutboundServiceTest.java
```diff
@@ -1,227 +0,0 @@
...
-import com.sap.cloud.sdk.datamodel.odatav4.core.GetAllRequestBuilder;
-import com.sap.cloud.sdk.datamodel.odatav4.core.VdmEntity;
-
-import io.vavr.control.Try;
-import lombok.Getter;
-import lombok.RequiredArgsConstructor;
-import lombok.SneakyThrows;
-import lombok.experimental.Delegate;
-
-class OutboundServiceTest
```
newtork, line 48: (Comment)

I compared coverage report and found no regression :+1:

newtork (approved):
Minor comment / question. Otherwise lgtm

cloudplatform/cloudplatform-core/src/main/java/com/sap/cloud/sdk/cloudplatform/requestheader/RequestHeaderAccessor.java
MatKuhr: I figured it's not frequently used (and probably shouldn't be frequently used :D ) so I saw no downside of keeping it beta for now. Maybe at some point we decide we are not happy if anybody overrides the facade and go for removing it..

CharlesDuboisSAP:
- Should `ScpCfTenantFacadeXsuaaTest` be deleted?
- There are still 85 classes that were annotated `@Beta` in v4 and that don't have any plans to be worked on.

MatKuhr:
> Should DefaultTenantFacadeXsuaaTest [(previously ScpCfTenantFacadeXsuaaTest)](https://github.wdf.sap.corp/MA/sdk/pull/8030/files#diff-454d9666302535c889813068710492438630a6bebeff778108cd2f33e1f31f1b) be deleted?

We still need to perform the JWT -> Tenant extraction correctly, so the test is still needed I'd say

> There are still 85 classes that were annotated @Beta in v4 and that don't have any plans to be worked on. Should they be stable?

For now I would keep them beta. If we find API problems later we then still have the option to fix it. Any specific classes where you think we should consider stabilizing?

CharlesDuboisSAP:
> We still need to perform the JWT -> Tenant extraction correctly, so the test is still needed I'd say

Can we undeprecte `DefaultTenantFacadeXsuaaTest` then?

> For now I would keep them beta. If we find API problems later we then still have the option to fix it. Any specific classes where you think we should consider stabilizing?

I would stabilize 
- Anything related to OnBehalfOf
- apache-http-client5
- connectivity-dwc
- Anything in OData modules and datamodel-metadata-generator
- Maybe more but I don't know much of the other classes

PR #161, opened 2023-11-24
fix: Default Tenant must not throw on euqlas, hashcode and toString

MattKuhr:
## Context

Rather nasty bug. Questionable if we want to stick to the "throw inside getter" approach. This still might happen e.g. during JSON serialization, depending on how the serialiser works.

But otherwise we would have to make changes to the API which may again break stuff...

### DIFF
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -96,35 +96,6 @@ public class DwcHeaderUtils
         return logonName;
     }
 
-    /**
-     * This method fetches the value of the {@link #DWC_CLIENT_HEADER} header or throws an
-     * {@link DwcHeaderNotFoundException} if the header was not found.
-     *
-     * @return The value of the {@link #DWC_CLIENT_HEADER} header.
-     * @throws DwcHeaderNotFoundException
-     *             if the header was not found.
-     */
-    @Nonnull
-    public static String getDwcClientOrThrow()
-    {
-        return getNonEmptyDwcHeaderValue(DWC_CLIENT_HEADER);
-    }
-
-    /**
-     * This method fetches the values of the {@link #DWC_SCOPES_HEADER} header throws an
-     * {@link DwcHeaderNotFoundException} if the header was not found.
-     *
-     * @return The values of the {@link #DWC_SCOPES_HEADER} header.
-     * @throws DwcHeaderNotFoundException
-     *             if the header was not found.
-     */
-    @Nonnull
-    public static Set<String> getDwcScopesOrThrow()
-    {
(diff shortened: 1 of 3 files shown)


cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
```diff
@@ -96,35 +93,6 @@ public static String getDwcPrincipalIdOrThrow()
...
-     * @throws DwcHeaderNotFoundException
-     *             if the header was not found.
-     */
-    @Nonnull
-    public static Set<String> getDwcScopesOrThrow()
-    {
-        final String scopes = getNonEmptyDwcHeaderValue(DWC_SCOPES_HEADER);
-        return Arrays.stream(scopes.split(" ")).collect(Collectors.toSet());
-    }
-
```
Johannes-Schneider, line 127: Drive-by breaking change? 🙈

Is there a specific reason to remove these methods other than (probably): "We are not using them"?

cloudplatform/tenant/src/main/java/com/sap/cloud/sdk/cloudplatform/tenant/DefaultTenant.java
```diff
@@ -17,8 +17,8 @@
 /**
  * Implementation of {@link Tenant} which can be used on SAP Business Technology Platform Cloud Foundry.
  */
-@EqualsAndHashCode
-@ToString
+@EqualsAndHashCode( doNotUseGetters = true )
+@ToString( doNotUseGetters = true )
```
Johannes-Schneider, line 21: Personally, I think this behavior is inconsistent with the rest of the API.
Why does the `toString` method reveal that that subdomain is indeed `null` but the getter indicates that it cannot be `null`?

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
MatKuhr: yeah, I think we just forgot to remove this API when we were removing the principal related XSUAA stuff

cloudplatform/tenant/src/main/java/com/sap/cloud/sdk/cloudplatform/tenant/DefaultTenant.java
MatKuhr: I think the ideation was that using the getter implies you actually need the subdomain in your logic. And it was already `@Nonnull`, so changing it to nullable is easy to overlook. But yeah, inconsistent ofc. Maybe better to just change the API contract to nullable, but no strong opinion..

Johannes-Schneider (commented):
Changes generally LGTM - assuming that the removed DwC utils are fine to be removed.

I'd like to get a second opinion from the team before merging this, though.

Johannes-Schneider (approved):
Lets do it!

PR #165, opened 2023-11-28
General Cleanup 2

CharlesDuboisSAP:
## Context

Follow-up of #158 

### Feature scope:

 
- [x] Deprecated `dwc-cf`
- [x] Stabilized `servlet-jakarta`
- [x] Removed any mention of `javax.inject`
- [x] Added missing release notes
- [x] Stabilized `DefaultTenantFacadeXsuaaTest`

## Definition of Done

- [x] [Documentation updated](https://github.com/SAP/cloud-sdk/pull/1601#discussion_r1407908497)
- [x] Release notes updated

### DIFF
--- a/archetypes/spring-boot3/src/main/resources/archetype-resources/application/pom.xml
+++ b/archetypes/spring-boot3/src/main/resources/archetype-resources/application/pom.xml
@@ -77,12 +77,6 @@
         </dependency>
         -->
 
-        <dependency>
-            <groupId>jakarta.inject</groupId>
-            <artifactId>jakarta.inject-api</artifactId>
-            <version>2.0.1</version>
-        </dependency>
-
         <!-- Dependencies for security setup -->
         <!--
         <dependency>
--- a/archetypes/spring-boot3/src/main/resources/archetype-resources/application/src/main/java/__packageInPathFormat__/Application.java
+++ b/archetypes/spring-boot3/src/main/resources/archetype-resources/application/src/main/java/__packageInPathFormat__/Application.java
@@ -12,7 +12,7 @@ import org.springframework.context.annotation.ComponentScan;
 
 @SpringBootApplication
 @ComponentScan({"com.sap.cloud.sdk", "${package}"})
-@ServletComponentScan({"com.sap.cloud.sdk.cloudplatform.servletjakarta", "${package}"})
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -607,14 +607,11 @@
 										<message>Use Java EE modules with scope provided or test.</message>
 										<excludes>
 											<exclude>javax:javaee-api</exclude>
-											<exclude>javax.inject:javax.inject</exclude>
```
newtork, line 610: (Question)
If I understand correctly, you would want to keep this line, such that `javax.inject` is banned for all scopes.

CharlesDuboisSAP: I added the exclude back, do you think transitive dependencies add javax.inject such that this safety is useful?

newtork: > such that this safety is useful?

Fair question, would we ever allow imperfect code contributions to the project? :)

My reasoning is:
* Who knows whether we'll allow for other team contributions in the future. They could make a mistake and add the dependency by accident.
* Maybe we make the mistake and include some RFC / JCo dependency back in, without knowing about `javax.inject` 🤔
* There are still other exclusions, so let's keep the list intact

CharlesDuboisSAP: **For clarity:** If I understand correctly, we allowed transitive use of javax.inject since this maven plugin modules are not included as dependency. Either we could:
- Remove `skipEnforceScopeJavaEE` on maven plugin modules
- Or keep the skipEnforce but not allow any transitive use of javax.inject in other modules, which would be safer and you are recommending

cloudplatform/dwc-cf/pom.xml
```diff
@@ -9,7 +9,8 @@
 	<artifactId>dwc-cf</artifactId>
 	<name>Cloud Platform - SAP Deploy with Confidence on Cloud Foundry (Beta)</name>
 	<description>Collection of all Cloud platform abstractions for SAP Deploy with Confidence stack hosted on Cloud
-        Foundry (Beta).</description>
+		Foundry (Deprecated, please replace with com.sap.cloud.sdk:sdk-core and
+		com.sap.cloud.sdk.cloudplatform:connectivity-dwc).</description>
```
Johannes-Schneider, line 13: The `<name>` (one line above) contains the _"(Beta)"_ suffix.

MatKuhr (approved):

PR #173, opened 2023-12-04
fix: PrincipalPropagationMode

MatKuhr:
## Context

This PR renames the `PrincipalPropagationMode` entries to more expressive terms. Intentionally breaking the API so that any users actively notice the change and switch to the recommended (default) `TOKEN_FORWARDING`

## Definition of Done

- [ ] Documentation updated
- [x] Release notes updated


Johannes-Schneider (approved):
As [discussed in Slack](https://sap-cpe.slack.com/archives/C01AMF777GA/p1701701801419559) lets break the API to make sure customers are aware of that change, as it is rather important.

PR #177, opened 2023-12-05
fix: error handling in OAuth2ServiceBindingDestinationLoader

Johannes-Schneider:
This PR fixes the error handling in our `OAuth2ServiceBindingDestinationLoader` in cases where a (customer-provided) `OAuth2PropertySupplier` throws an Exception in any of the getters.

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoader.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoader.java
@@ -148,6 +148,10 @@ public class OAuth2ServiceBindingDestinationLoader implements ServiceBindingDest
         catch( final DestinationAccessException e ) {
             return Try.failure(e);
         }
+        catch( final Exception e ) {
+            final String msg = "Failed to retrieve OAuth2 properties for service binding of service '%s'.";
+            return Try.failure(new DestinationAccessException(String.format(msg, identifier), e));
+        }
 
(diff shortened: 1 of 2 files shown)


MatKuhr (approved):

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoader.java
```diff
@@ -148,6 +148,10 @@ public Try<HttpDestination> tryGetDestination( @Nonnull final ServiceBindingDest
         catch( final DestinationAccessException e ) {
             return Try.failure(e);
         }
+        catch( final Exception e ) {
+            final String msg = "Failed to retrieve OAuth2 properties for service binding of service '%s'.";
```
CharlesDuboisSAP, line 152: This error should tell users that the binding doesn't have the three properties above and how to fix their binding

MatKuhr: Well we don't know if the binding doesn't have the properties. The exception could be thrown for any unknown reason. E.g. a property might be in the wrong format, or there may be a bug in the users implementation of `OAuth2PropertySupplier`

CharlesDuboisSAP: The error is telling the user that we failed to parse it when it's the service binding who failed to follow the spec.

PR #189, opened 2023-12-15
[Fix] Invalid Destination Domain Throws

CharlesDuboisSAP:
## Context
 

Closes SAP/cloud-sdk-java-backlog#382.

Invalid domain names break our `UriPathMerger`, which sends a broken URI to the HttpClient.

### Feature scope:

 
- [x] Add a meaningful exception message if `UriPathMerger` if `getHost` is null

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/UriPathMerger.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/UriPathMerger.java
@@ -14,7 +14,6 @@ import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.StringUtils;
 
-import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationAccessException;
 import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationPathsNotMergeableException;
 import com.sap.cloud.sdk.cloudplatform.exception.ShouldNotHappenException;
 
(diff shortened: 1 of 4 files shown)


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/UriPathMerger.java
```diff
@@ -59,6 +60,11 @@ public URI merge( @Nonnull final URI primaryUri, @Nullable final URI secondaryUr
             throw new DestinationPathsNotMergeableException("The destination URI must be absolute OR empty.");
         }
 
+        if( primaryUri.getHost() == null ) {
```
MatKuhr, line 63: arguably we should also apply this check for `secondaryUri`, probably in line 56?

release_notes.md
```diff
@@ -25,6 +25,6 @@ docs: https://sap.github.io/cloud-sdk/docs/java/release-notes
 
 ## 🐛 Fixed Issues
 
-- 
+- Invalid destination domain names now throw a DestinationAccessException when executing a request.
```
MatKuhr, line 28:
```suggestion
- Fix an issue where an invalid URL in a destination would lead to an empty hostname.
  Now a `DestinationAccessException` will be thrown upon executing a request, if the URL domain contains a hostname that violates [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt).
```

newtork, line 28: I don't know, I'm still of the opinion: maybe we are doing it wrong:

<details><summary>Instead of extracting "host" and "user info", we could just extract "authority"</summary>

image

image


</details>

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/UriPathMerger.java
CharlesDuboisSAP: I don't mind if the domain is _technically_ invalid on the secondary URL, it would never return a broken URL

newtork (approved):
. o O (acceptable breaking change)

PR #190, opened 2023-12-15
fix: Destination Caching and Error Handling

MatKuhr:
## Context

This PR fixes an issue which alters the intended error handling and caching for retrieving individual destinations.

For context also see this PR on [Cloud SDK 4](https://github.wdf.sap.corp/MA/sdk/pull/8336).

### Feature scope:

- [x] Implement behaviour of eagerly throwing in case of token errors
  - [x] Implicitly: Prevent token errors from getting cached
- [x] Move & update explicit tests
- [x] Refactor & adjust the big 🧠  test covering all combinations

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java
```diff
@@ -215,509 +222,433 @@ static Collection<TestParameters> createTestParameters()
...
-                        "No test parameters found for tenant '" + tenant + "' and principal '" + principal + "'."));
+            return DestinationServiceOptionsAugmenter.getTokenExchangeStrategy(options).get();
         }
 
-        @Override
-        public String toString()
+        void stubDestinationRetriever( BiFunction<String, DestinationOptions, Destination> destinationRetriever )
         {
-            return authenticationType + " + " + tokenExchangeStrategy;
+            if( DestinationUtility.requiresUserTokenExchange(getAuthenticationType(), null)
```
MatKuhr, line 433: this is not super great, it needs to be in line with what the other classes actually do. We could replace this with some of the actual logic, e.g. involving `DestinationRetrievalStrategyResolver`. But then we blow up the test further...

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java
```diff
@@ -215,509 +222,433 @@ static Collection<TestParameters> createTestParameters()
...
-        }
+            @Nullable
+            private Tenant tenant;
+            @Nullable
+            private Principal principal;
+
+            // expected outcome
+            private boolean isCommandCreationExpectedToSucceed = true;
+            private boolean isCommandExecutionExpectedToSucceed = true;
+            private Tuple2<Tenant, Principal> expectedIsolationKeyStrategy;
```
MatKuhr, line 457: Using VAVR here because the actual cache key depends on the destination name + options which only get added later. Creating a `CacheKey` here and appending that data later via `append` is not super great because `CacheKey` is mutable

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java
```diff
@@ -215,509 +222,433 @@ static Collection<TestParameters> createTestParameters()
...
+        private final Collection<DestinationServiceTokenExchangeStrategy> tokenExchangeStrategies = new HashSet<>();
 
         @Nonnull
-        public
-            UserSpecificTestParametersBuilder
-            expectIsolationCacheKey( @Nullable final Tenant tenant, @Nullable final Principal principal )
+        TestCaseBatchBuilder forAuthenticationTypes( @Nonnull final Collection<AuthenticationType> authenticationTypes )
         {
-            expectedIsolationCacheKeySupplier = () -> CacheKey.of(tenant, principal);
+            if( !testCases.isEmpty() ) {
```
MatKuhr, line 586: prevents invoking the builder methods in the wrong order. I felt this is better (considering this is a test and not productive API) than adding another e.g. `FinalBuilder` class 😄

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommand.java
```diff
@@ -116,16 +116,17 @@ static Try<GetOrComputeSingleDestinationCommand> prepareCommand(
...
+     * | 2   | Lookup then exchange
+     *         - (JWT is available) | User Propagation   | Tenant             | Tenant + Principal |                            |
+     *         - (No JWT available) | User Propagation   | Tenant             | -                  | execution fails            |
      * | 3   | Exchange only        | User Propagation   | Tenant + Principal | Tenant + Principal |                            |
      * | 4   | Forward user token
-     *        (JWT is available)    | User Propagation   | Tenant             | Tenant + Principal |                            |
-     * | 5   | Forward user token
-     *         (JWT is unavailable) | User Propagation   | Tenant             |   -                |Retrieving destination fails|
+     *         - (JWT is available) | User Propagation   | Tenant             | Tenant + Principal |                            |
+     *         - (No JWT available) | User Propagation   | Tenant             | -                  | warning logged             |
```
Johannes-Schneider, line 129: What is the exact reason for just logging a warning here instead of throwing an exception?
Because semantically, it seems we are in an illegal state if the destination uses `User Propagation` but there is no user to propagate.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommand.java
```diff
@@ -158,6 +159,24 @@ Try<Destination> execute()
...
+                    .get(DestinationProperty.AUTH_TOKENS)
+                    .getOrElse(List::of)
+                    .stream()
+                    .map(t -> ((DestinationServiceV1Response.DestinationAuthToken) t).getError())
+                    .anyMatch(Objects::nonNull);
+
+            if( resultContainsTokenErrors && exchangeStrategy != LOOKUP_ONLY ) {
+                log
+                    .warn(
+                        "The destination {} contains token errors. "
```
Johannes-Schneider, line 173: Should we maybe add the used `exchangeStrategy` to the log message? Might become handy when troubleshooting.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommand.java
```diff
@@ -158,6 +159,24 @@ Try<Destination> execute()
...
+                            + "Caching for this destination is skipped and usage of the destination may result in exceptions. "
+                            + "To only access properties of a destination, without performing authentication flows, please use DestinationService.tryGetAllDestinations() instead.",
+                        destinationName);
+                return Try.success(result);
```
Johannes-Schneider, line 177: Personally, I feel it's very odd to not throw an exception here.

Because at this point in time, we know that the user is not exclusively interested in reading the destination's properties (otherwise they would've used `LOOKUP_ONLY`).
So (I think) it is fair to assume that the customer wants to actually connect to the target system.

However, we **know** that authentication at the target system will definitely fail (our `AuthTokenHeaderProvider` **will** throw an exception upon calling the `getHeaders()` method).
So we would we pretend as if everything is okay (i.e. `Try.success`)?

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java
```diff
@@ -46,14 +50,16 @@
...
-import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
-import lombok.SneakyThrows;
 
+@SuppressWarnings( "unchecked" )
+@DisplayName( "Test all Combinations of AuthType, TokenExchangeStrategy, Tenant and Principal" )
 class GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest
 {
+    // region ( CONSTANTS )
```
Johannes-Schneider, line 62: Code folding 🙌

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java
```diff
@@ -170,16 +177,16 @@ static Collection<TestParameters> createTestParameters()
...
                 createBatchTestCases()
                     .forAuthenticationTypes(AUTHENTICATION_TYPES_THAT_DO_NOT_REQUIRE_TOKEN_EXCHANGE)
                     .forAuthenticationTypes(AUTHENTICATION_TYPES_THAT_REQUIRE_TOKEN_EXCHANGE)
                     .forTokenExchangeStrategies(EXCHANGE_ONLY)
-                    .withUserParameters(expectCommandCreationToFail(NO_TENANT, NO_PRINCIPAL))
-                    .withUserParameters(expectCommandCreationToFail(NO_TENANT, PRINCIPAL))
-                    .withUserParameters(expectCommandCreationToFail(TENANT, NO_PRINCIPAL))
-                    .withUserParameters(
-                        expectCommandCreationToSucceed(TENANT, PRINCIPAL)
+                    .withExpectation(forTenantAndPrincipal(NO_TENANT, NO_PRINCIPAL).expectCommandCreationToFail())
```
Johannes-Schneider, line 185: Reads nicely!

Johannes-Schneider (commented):
First round of thoughts: Mainly focused on the production code changes.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceFactory.java
```diff
@@ -124,6 +125,18 @@ private static Destination handleHttpDestination(
...
     }
 
+    private static void throwOnTokenError(
+        final String destinationName,
+        final DestinationServiceV1Response.DestinationAuthToken destinationAuthToken )
+    {
+        if( destinationAuthToken.getError() != null ) {
+            final String msg =
+                "Failed to read authentication token of destination '%s'. The destination service responded with an error: '%s'.\n"
+                    + "In case only the properties of a destination should be accessed, without performing authorization flows, please use the 'getDestinationProperties'  method on 'DestinationService' instead.";
```
Johannes-Schneider, line 135: (comment)
We have to remember to ship this change together with our new APIs (#199) as the exception message otherwise doesn't make sense.

---

I'm a bit worried about the trend of putting implementation names into our exception messages.
On the one hand side, I see that it is (hopefully) great for customers to support themselves.

On the other hand, however, it increases the risk of us overlooking those hard coded strings when we rename / deprecate + replace methods in the future.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/AuthTokenHeaderProviderTest.java
```diff
@@ -102,7 +102,7 @@ void testLegacyFallbackLogic()
     }
 
     @Test
-    void testThrowsOnHeaderError()
+    void testHeaderErrorIsUnexpected()
```
Johannes-Schneider, line 105: This test shouldn't be passing, should it?

release_notes.md
```diff
@@ -32,7 +34,7 @@ docs: https://sap.github.io/cloud-sdk/docs/java/release-notes
 
 - Fix an issue where an invalid hostname in a destination would lead to an empty hostname. The hostname is now accepted.
 - Fix an issue where errors from token flows of destinations retrieved from the BTP destination service were not handled consistently.
-  For the default strategy `FORWARD_USER_TOKEN` errors were handled too eagerly, while for some custom strategies erroneous results would get cached unintentionally.
-  This is now resolved, error handling and caching now behaves as documented.
+  For the non-default strategy `LOOKUP_ONLY` errors were handled too late and results would get cached unintentionally.
```
Johannes-Schneider, line 37: This is of course a breaking behavior change. I think we should add a sentence that goes like

_using `LOOKUP_ONLY` is strongly discouraged. Instead, use [new api here]_ in our #200 (didn't check whether that already happened).

Johannes-Schneider (commented):
Production code changes generally LGTM - I have to go over test updates still.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/AuthTokenHeaderProviderTest.java
MatKuhr: It should, because of [this line](https://github.com/SAP/cloud-sdk-java/pull/190/files#diff-7871d25f7c9cff9bc6968d6acffb4d1ab88988540e6c77af0b5d48c752dd5408L98).

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceFactory.java
MatKuhr: Yeah, I guess the alternative would be to use reflection to dynamically compute the method name 😅

Johannes-Schneider: Yeah... reflection doesn't sound like a nice solution here... But how about a little bit of Java compiler tweaking? 😏 </ironie>

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/AuthTokenHeaderProviderTest.java
Johannes-Schneider: Ohh... I see. Hard to get that while looking at the diff only.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAuthenticationTest.java
```diff
@@ -727,43 +724,4 @@ void testSAPAssertionSSOWithUserTokenExchange()
             .getConfigurationAsJson(anyString(), eq(OnBehalfOf.NAMED_USER_CURRENT_TENANT));
         verify(destinationService, never()).getConfigurationAsJsonWithUserToken(anyString(), any(OnBehalfOf.class));
     }
-
-    @Test
-    void testSAPAssertionSSOFailure()
```
Johannes-Schneider, line 732: Why has this test been removed? The behavior should still be the same (just the exception message has slightly changed)

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceWithoutTokenTest.java
```diff
@@ -74,30 +73,6 @@ void tearDown()
         AuthTokenAccessor.setAuthTokenFacade(null);
     }
 
-    @Test
-    void testForwardUserTokenWithoutPrincipalFails()
```
Johannes-Schneider, line 78: Why has this test been removed? All that changed is the root cause message.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/AuthTokenHeaderProviderTest.java
MatKuhr: Done ✅

release_notes.md
```diff
@@ -12,7 +12,9 @@ docs: https://sap.github.io/cloud-sdk/docs/java/release-notes
 
 - `UriBuilder.build(scheme, userInfo, host, port, path, query, fragment)` has been deprecated in favor of
  `UriBuilder.build(scheme, authority, path, query, fragment)`.
-
+- An earlier version of the [V5 Upgrade Guide](https://sap.github.io/cloud-sdk/docs/java/guides/5.0-upgrade-steps) contained an instruction to move handling of `DestinationAccessExceptions` from `DestinationAccessor.getDestination()` to `destination.getHeaders()`.
+  This instruction was incorrect and has been removed.
+  In case you have followed this instruction, please revert the change.
```
Johannes-Schneider, line 17: Maybe we should move this up to be the first note in the section.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceWithoutTokenTest.java
MatKuhr: After the logic changes this test didn't really test anything that wasn't already tested with better confidence  elsewhere. While it's generally good to have more tests, in this case I felt the risk is high that some part of the stubbing becomes outdated and then the assertions would be giving false confidence.

Same goes for the other removed test above.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java
```diff
@@ -84,640 +90,537 @@ class GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest
...
-                        "No test parameters found for tenant '" + tenant + "' and principal '" + principal + "'."));
+            return DestinationServiceOptionsAugmenter.getTokenExchangeStrategy(options).get();
         }
 
-        @Override
-        public String toString()
+        void stubDestinationRetriever( BiFunction<String, DestinationOptions, Destination> destinationRetriever )
         {
-            return authenticationType + " + " + tokenExchangeStrategy;
+            if( DestinationUtility.requiresUserTokenExchange(getDestination())
```
MatKuhr, line 407: We could get rid of stubbing this behavior if we mocked `DestinationServiceV1Response` instead of `BiFunction<String, DestinationOptions, Destination>`. Because this test is only meaningful if this accurately represents the behavior of `DestinationService::loadAndParseDestination`

 But this requires a bit more refactoring, so I'd rather do this in a follow-up PR.

release_notes.md
MatKuhr: Done ✅

Johannes-Schneider (approved):
LGTM

PR #194, opened 2023-12-18
feat: [Destinations] Warn on Expired Auth Token Usage

MatKuhr:
## Context

This PR adds a warning which is logged when headers for an expired auth token are accessed.

We could also consider throwing an exception, but for now I didn't do that for two reasons:

* The destination service could send incorrect expiry information, or there may be a grace period
* Headers may be accessed also outside of actually executing requests

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


Johannes-Schneider (approved):
LGTM

PR #201, opened 2023-12-21
Fix compatibility v4 / v5 with CAP

newtork:
Currently CAP has to make the following code change in their productive code (non-test) for migrating to SDK v5:

```diff
- httpDestination = new DefaultDestination(destination.getProperties()).asHttp();
+ httpDestination = DefaultHttpDestination.fromMap(destination.getProperties()).build();
```

Implication:
* There's a chance for users of SDK v5 + CAP v2 to experience runtime exception about constructor visibility.

We could solve it by making the constructor public again.

Uses in current CAP here:
* https://github.wdf.sap.corp/cds-java/cds-services/blob/main/cds-feature-remote-odata/src/main/java/com/sap/cds/services/impl/odata/RemoteODataClient.java#L88
* https://github.wdf.sap.corp/cds-java/cds-services/blob/main/cds-feature-remote-hcql/src/main/java/com/sap/cds/services/impl/hcql/RemoteHcqlClient.java#L95


newtork:
We're planning to solve it differently.

newtork:
Closed in favor of this
https://github.wdf.sap.corp/cds-java/cds-services/pull/3723

PR #210, opened 2024-01-04
feat: [Resilience] Improve Downwards Compatibility for Resilience API

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#389.

This PR makes the upgrading the Cloud SDK from 4 to 5 easier. 

Previously, a Cloud SDK 5 user had to exclude any occurrence of `com.sap.cloud.sdk.frameworks:resilience4j` in case it came in transitively. With this change, this is no longer needed, as the Cloud SDK 4 strategy will gracefully be ignored, if there is exactly one alternative.

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/resilience/src/test/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecoratorTest.java
```diff
@@ -68,6 +68,40 @@ void testGetDecorationStrategyThrowsOnMultipleStrategies()
...
+        final ResilienceDecorationStrategy thirdStrategy = mock(ResilienceDecorationStrategy.class);
+        mockDecorationStrategies(firstStrategy, secondStrategy, thirdStrategy);
+        ResilienceDecorator.resetDecorationStrategy();
+
+        assertThatThrownBy(ResilienceDecorator::getDecorationStrategy)
+            .as("More than one non-legacy strategy should still lead to an exception.")
+            .isExactlyInstanceOf(ResilienceRuntimeException.class)
+            .hasCauseExactlyInstanceOf(ObjectLookupFailedException.class);
+
+        ResilienceDecorator.LEGACY_DECORATION_STRATEGY = originalLegacyStrategy;
```
Johannes-Schneider, line 102: Please do not restore static state within the test case itself.
In this very concrete example, it shouldn't be too bad if the static state is never correctly restored, but to not provide future-us with bad examples, I'd still say we shouldn't do it here.

Generally, I'm not a fan of making this "constant" a non-constant for testing purposes only.
However, I don't really see an alternative here... :/

cloudplatform/resilience/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecorator.java
```diff
@@ -26,6 +27,12 @@
 @Slf4j
 public final class ResilienceDecorator
 {
+    static String LEGACY_DECORATION_STRATEGY;
```
Johannes-Schneider, line 30: [minor]

Let's make it more obvious as to why this is explicitly not `final` (as anyone should expect from the name).

```suggestion
    // non final for testing
    static String LEGACY_DECORATION_STRATEGY;
```

Johannes-Schneider (commented):
I generally like the idea.

cloudplatform/resilience/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecorator.java
newtork: I just wouldn't make it a constant then.

cloudplatform/resilience/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecorator.java
```diff
@@ -77,29 +84,52 @@ private static ResilienceDecorationStrategy getDefaultDecorationStrategy()
...
+                    "Too many implementations of %s found. Make sure to only have a single implementation of the interface on your classpath: %s",
+                    ResilienceDecorationStrategy.class.getName(),
+                    classes);
+        final String logMessage =
+            String
+                .format(
+                    "%s. Using any resilience pattern will lead to an exception at runtime UNLESS the %s is explicitly overwritten using 'ResilienceDecorator.setDecorationStrategy(ResilienceDecorationStrategy)'.",
+                    exceptionMessage,
+                    ResilienceDecorationStrategy.class.getName());
+        log.warn(logMessage);
```
newtork, line 131: (Minor)

Using string format instead of log format

cloudplatform/resilience/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecorator.java
```diff
@@ -77,29 +84,52 @@ private static ResilienceDecorationStrategy getDefaultDecorationStrategy()
...
+                            .info(
+                                "Ignoring legacy implementation 'com.sap.cloud.sdk.frameworks.resilience4j.Resilience4jDecorationStrategy' from Cloud SDK version 4.X in favor of: {}",
+                                it))
+                    .findAny()
+                    .get();
+            }
+        }
 
-        return facades.iterator().next();
+        final String classes = facades.stream().map(f -> f.getClass().getName()).collect(Collectors.joining(", "));
```
newtork, line 118: (Comment)

In a personal project I wouldn't concatenate the elements manually, but instead use native `Collection#toString`. It may not be best practice to rely on implicit implementation, but you would get a very similar result.

cloudplatform/resilience/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecorator.java
```diff
@@ -26,6 +27,12 @@
 @Slf4j
 public final class ResilienceDecorator
 {
+    static String LEGACY_DECORATION_STRATEGY;
+
+    static {
```
newtork, line 32: (Minor)

Can't the static block be combined with the variable declaration?

newtork (approved):

cloudplatform/resilience/src/test/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceDecoratorTest.java
MatKuhr: I considered using before/after methods, but I don't see how that would help. Because the only way the current solution messes with other tests would be if tests are executed in parallel (within the same JVM) and other tests still continue to run after this test has failed (and thus failed to restore the original state). Worst case: There are more test failures caused by this test failing.

However, if we use before/after on static stuff with parallel execution (within the same JVM) that would cause (potentially even more) problems. There the race condition would apply more often because the before/after are executed for each test.

The proper solution would probably be to make `ResilienceDecorator` instance based and have all static methods from `ResilienceDecorator` delegate to the default instance. Then you can properly unit test the `ResilienceDecorator ` instances.. but I didn't want to go down that rabbit hole just yet for this change 😄 

A simple test improvement would be try/finally, but it kinda looks ugly 😅

PR #211, opened 2024-01-05
chore: [Destinations] Add Test on Destination Header Evaluation in Cached HttpClients

MatKuhr:
## Context

Extend existing test coverage to also assert on the header behavior


newtork (changes requested):
Important finding, though I'm not sure this should be tested as assertion / "desired state" without comment or backlog item.

I see the root cause is of course not your PR. I would just hope for a more "productive" outcome.

newtork (approved):

PR #212, opened 2024-01-05
fix: [Destinations] Improve Caching for Service Binding Destinations

MatKuhr:
## Context

This PR improves caching for destinations loaded from OAuth service bindings. It removes the random UUID from the destination name and resilience identifier, causing two improvements:

- The destinations now use the same HttpClient
  - As of #218 this is now safe to do
- The resilience for token flows will now be applied correctly across the same OAuth service, even when destination instances are different
  - Potential downside: With this change users can no longer override the default resilience config (in particular timeout) for the token flows. Previously this was possible via an undocumented destination property.
  - [ ] TBD: should we add this possibility back in?

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/resilience/src/test/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceConfigurationTest.java
```diff
@@ -25,23 +25,49 @@ void testCacheConfigurationSerializable()
         final CacheConfiguration configNoParameters = builder.withoutParameters();
         assertThat(configNoParameters.serializable()).isTrue();
         assertThat(configNoParameters.expirationStrategy()).isEqualTo(CacheConfiguration.DEFAULT_EXPIRATION_STRATEGY);
+        assertThat(configNoParameters).isEqualTo(builder.withoutParameters());
```
MatKuhr, line 28: Driveby test improvements

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceTest.java
```diff
@@ -13,20 +16,181 @@
...
 import com.sap.cloud.security.xsuaa.tokenflows.ClientCredentialsTokenFlow;
 import com.sap.cloud.security.xsuaa.tokenflows.XsuaaTokenFlows;
 
 import lombok.SneakyThrows;
 
 class OAuth2ServiceTest
 {
-    private static final ResilienceConfiguration NO_RESILIENCE =
-        ResilienceConfiguration.empty(OAuth2ServiceTest.class.getName() + "_empty");
+    private static final String RESPONSE_TEMPLATE = """
```
MatKuhr, line 44: moved over from `OnPremTest` as the tests are unrelated to any on-prem specifics

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceTest.java
```diff
@@ -13,20 +16,181 @@
...
+
+        SERVER_1.verify(1, postRequestedFor(urlEqualTo("/oauth/token")));
+    }
+
+    @Test
+    void multipleOAuth2ServiceImpl()
+    {
+        OAuth2Service sut;
+        // tokens for the same server and identity should be cached
+        sut = new OAuth2Service(SERVER_1.baseUrl(), IDENTITY_1, TECHNICAL_USER_CURRENT_TENANT);
```
MatKuhr, line 101: enhancement compared to old test

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -67,12 +73,27 @@ class OAuth2Service
...
+        /*
+         * Reasoning for always using ResilienceIsolationMode.TENANT_OPTIONAL, regardless of onBehalfOf:
+         * - for TECHNICAL_USER_CURRENT_TENANT this is trivially correct
+         * - for NAMED_USER_CURRENT_TENANT the resilience should still be applied per-tenant only
+         * - for TECHNICAL_USER_PROVIDER && current tenant != provider the isolation is stronger than it needs to be,
+         *   but the downside is arguably not worth keeping a second configuration for this case only
+         */
+        resilienceConfig =
+            ResilienceConfiguration
+                .of(identity.getId())
```
Johannes-Schneider, line 93: [question]

Why are we using the `identity` here as the "key" for our resilience?
I would have expected the `uri` to be used.

That way, we would apply resilience (i.e. bulkheads) correctly towards the target system.
In contrast to that (i.e. if we were using the `identity`), we would create a new resilience for the same target service if the binding was rotated recently.

Or am I overlooking something here?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
```diff
@@ -90,27 +83,10 @@ public static BuilderWithTargetUrl forTargetUrl( @Nonnull final String targetUrl
...
 
-            // random uuid to make sure equals and hash code works effectively by reference
-            final String destinationName = UUID.randomUUID().toString();
-            destinationBuilder.name(destinationName);
-
-            destinationBuilder
-                .property(
-                    OAuth2HeaderProvider.PROPERTY_OAUTH2_RESILIENCE_CONFIG,
-                    createTokenRetrievalResilienceConfiguration(destinationName));
+            destinationBuilder.name(client.getId());
```
Johannes-Schneider, line 86: [major]

The `client.getId()` could be considered confidential. Therefore, I wouldn't want to use it as the destination name.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoader.java
```diff
@@ -209,6 +204,24 @@ private OAuth2PropertySupplier getOAuth2PropertySupplier( @Nonnull final Service
...
+        @Nonnull final OnBehalfOf behalf,
+        @Nullable final ServiceIdentifier serviceIdentifier )
+    {
+        log.debug("Creating a new OAuth2 destination for service {}.", serviceIdentifier);
+        final DestinationHeaderProvider headerProvider =
+            createHeaderProvider(tokenUri, clientIdentity, behalf, HttpHeaders.AUTHORIZATION);
+        final String idString =
+            Option.of(serviceIdentifier).map(ServiceIdentifier::toString).getOrElse("unknown")
+                + "-"
+                + clientIdentity.getId();
```
Johannes-Schneider, line 221: [major]

Same remark as above: The `clientIdentity.getId()` is confidential and, therefore, shouldn't be used as a destination name IMO.

Johannes-Schneider (changes requested):
I reviewed the production code only.

In there, I'd like to not use the client id as part of the destination name.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
MatKuhr: why would it be confidential? I'd say it's exactly the opposite, this is what you (need to) share publicly. For example, the client ID is also used for reuse services to perform callbacks to an application ([src](https://github.wdf.sap.corp/CPSecurity/Knowledge-Base/blob/master/03_ApplicationSecurity/ProofOfPossession.md#rules-for-token-requests-and-token-forwarding)).

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
MatKuhr: the resilience is applied for the token flows only, not the target system. We could use the tokenUrl as key, but this is probably not isolated enough, since the tokenUrl is typically shared for many (if not all) different OAuth flows. If one service is misconfigured, the circuit breaker could block all other token flows. 

E.g. let's say you are using 5 different services, one of them being the destination service. Now assume you include the destination service in your SaaS registry callback and token requests for it are failing. Now the other 4 services are also blocked (assuming the circuit breaker is triggered by that).

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
Johannes-Schneider: From the linked documentation, I assume you talking about following:

> Hereafter the cases, where token forwarding are **not going to work** and a new token must be requested explicitly:
> 
> [ ... ]
> 
> Reuse Service
> - to Consumer Application
>   - request [consumer-specific token](#consumer-specific-token) with `resource` parameter according to [RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707) <br>``resource`` = ``urn:sap:identity:consumer:clientid:<consumer client id>``

If so, I wouldn't call it _"this is what you (need to) share publicly"_.
Sure, you need to include your `clientid` when fetching that particular token - but that is of course also true for fetching any other token, right?
(I.e. We have to send out `clientid` and even the `clientsecret` to an XSUAA instance in order to retrieve some access token for a specific service)

And even if the returned token is then forwarded to the target system (which will then use it to trigger the application's callback, ultimately), I'd still argue that the token itself is of course also confidential.
You wouldn't share your authorization token publicly, would you?

At least this is how I would think of the entire situation.
Maybe I'm missing some important information or I'm just plain wrong here.

MatKuhr: > Sure, you need to include your clientid when fetching that particular token - but that is of course also true for fetching any other token, right?

Not really, at least not how I understand it. In this example the requester is actually including the client ID of the other application in addition to its own (i.e. you get a call A -> B, B remembers the client ID of A from that call and then requests a token from IAS using both client IDs A and B)

While the signed token is confidential of course, the decoded content i.e. its claims are generally not. Of course, you can potentially have PII in there (e.g. email address) which would be subject to other concerns regarding DPP, but the client id doesn't fall in that category as it is a technical ID

Johannes-Schneider: Okay, I think [this documentation](https://github.wdf.sap.corp/CPSecurity/Knowledge-Base/blob/master/03_ApplicationSecurity/ProofOfPossession.md#service-consumer-specific-token) makes things bit clearer.

Here, the cURL request contains two different `clientid`s, just as you explained, @MatKuhr:

```bash
curl --cert certificate.pem --key key.pem \
     -X POST https://<<customer tenant host>>/oauth2/token \ 
     -d 'grant_type=client_credentials&client_id=<<clientid>>&resource=urn:sap:identity:consumer:clientid:<<consumer identifier>>'
```

(There is the `clientid` and `consumer identifier`, which I assume to be a `clientid` as well, judging from the resource name)

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
Johannes-Schneider: > the resilience is applied for the token flows only, not the target system.

Yes, true. I meant the token service when I said _"target system"_.

---

> We could use the tokenUrl as key, but this is probably not isolated enough

mhmm.. I see.
So it would be nice if we could apply the bulkhead on a "per URL" basis but keep the circuit breakers isolated per "identity"... 🤔

Could be done if we really wanted to (i.e. we could nest two resilient calls) but it's probably not worth the extra code complexity.

MatKuhr: Yes, for now I figured the most straight-forward is to apply the resilience patterns on the same level. The bulkhead generally makes sense to line up with the rate limit on the server side. I'm not 100% sure how to interpret [this document](https://github.wdf.sap.corp/CPSecurity/Knowledge-Base/blob/master/03_ApplicationSecurity/RateLimit.md) on the XSUAA limits, but it sounds like it's per sub account, so we probably could generalize to the token URL and use a much higher value than the default (50) as size. But I don't think it's super needed and error handling is probably much easier if its isolated on client ID.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
MatKuhr: Yes this is what I meant 👍🏻 

But that is ofc not a super strong indicator, so I finally found an explicit statement about this topic in the [relevant RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-2.2) now which states the ID is not a secret:

> The client identifier is not a secret; it is exposed to the resource owner and MUST NOT be used
   alone for client authentication.
   
Of course, having it makes a guessing attack easier in theory, but the client secret is a generated secret for technical auth only (and not a user password) so there is no excuse for it being so short that it can be guessed.

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoaderTest.java
```diff
@@ -361,54 +360,41 @@ void testProxiedDestination()
...
-        assertThat(result.get(OAuth2HeaderProvider.PROPERTY_OAUTH2_RESILIENCE_CONFIG)).isNotEmpty();
-        config = (ResilienceConfiguration) result.get(OAuth2HeaderProvider.PROPERTY_OAUTH2_RESILIENCE_CONFIG).get();
-        assertThat(config.identifier()).startsWith(baseDestination.get(DestinationProperty.NAME).get());
+        assertThat(secondInvocationResult)
+            .as("There should not be a cache in place for proxied destinations.")
+            .isNotSameAs(result);
+        assertThat(secondInvocationResult)
+            .as("The destination objects should be equal so that they use the same HTTP client.")
+            .isEqualTo(result)
+            .isNotSameAs(result);
```
Johannes-Schneider, line 390: [minor]

Duplicated assertion.

```suggestion
            .isEqualTo(result);
```

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceTest.java
```diff
@@ -13,20 +16,180 @@
...
+            .allSatisfy(
+                event -> assertThat(event.getRequest().getHeaders().all())
+                    .noneSatisfy(header -> assertThat(header.key()).isEqualToIgnoringCase("Cookie")));
+    }
+
+    @Test
+    void retrieveAccessTokenWithSameIdentityOnDifferentUrisShouldNotReturnCachedResponse()
+    {
+        final OAuth2Service service1 = new OAuth2Service(SERVER_1.baseUrl(), IDENTITY_1, TECHNICAL_USER_CURRENT_TENANT);
+        final OAuth2Service service2 = new OAuth2Service(SERVER_2.baseUrl(), IDENTITY_1, TECHNICAL_USER_CURRENT_TENANT);
```
Johannes-Schneider, line 182: In this test, we even anticipated the super-edge-case where two services (coincidentally) shared the exact same client identity.

Right now, our updated resilience configuration would NOT separate those two services.

Should we add the token service URI to the resilience config "key" to match this implementation detail?

MatKuhr: Good point, I added the URL in now. Although unlikely, it's probably still better to have it in. Although we currently only support XSUAA as auth server and there the IDs are guaranteed to be unique. But once we add IAS this is probably no longer the case..

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
Johannes-Schneider: Thanks for digging up that RFC.

I'm still wondering whether customers also know about this (I highly doubt it, honestly).
So I'm afraid we will see issue coming our way that claim that we are "exposing client secrets" in our logging.
We would then need to remember that this is indeed not an issue and then also find the link you just posted... Not very lightweight on the support facilitator's shoulders IMO.

So how about we use something like the following for the destination's name: `"Programatic Destination " + identity.getId().hashCode()`

Something like this would, in theory, introduce the "risk" of having equal names for different destinations (due to possible hash collisions), but that shouldn't be an issue for our HttpClient caching as the name is ofc not the only property that is considered.

Any opinions about that?

newtork: (OT)
```diff
 - "Programatic Destination " + identity.getId().hashCode()
 +  "Programatic Destination " + AtomicInteger#getAndIncrement()
 ```

MatKuhr: Discussed offline, used the hash code solution for now 👍🏻

Johannes-Schneider (approved):
LGTM

PR #215, opened 2024-01-10
chore: [Destinations] Improve mTLS Test

MatKuhr:
## Context

Move the test somewhere where it can actually be found 😉 
Also, create an HttpClient5 variant of this test.


cloudplatform/connectivity-apache-httpclient4/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ClientCertificateAuthenticationLocalTest.java
```diff
@@ -121,65 +92,70 @@ static class HostCorrectlyConfiguredTest extends ClientCertificateAuthentication
                 .build();
 
         @RegisterExtension
-        static final WireMockExtension ERP_SERVER =
+        static final WireMockExtension server =
```
newtork, line 95: (Minor)

I think all-caps for constant was better before.

newtork (approved):

PR #216, opened 2024-01-11
Add test case for special characters in domains for URI merging

newtork:
Add test case for special characters in domains for URI merging

Follow up from https://github.com/SAP/cloud-sdk-java/pull/189


MattKuhr (approved):

MatKuhr:
@newtork should work now if you update the branch (see #249 )

PR #218, opened 2024-01-11
fix: [Destination] Improve HTTP Client Cache Header Behavior

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#396.

This PR implements an approach where the HttpClient wrapping is applied for each destination, and only the actual http client (which the wrapper delegates to) is re-used. Wrapping is cheap, and with the current approach this is done without breaking API or behavior changes.

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpClientCache.java
```diff
@@ -61,7 +61,19 @@ public DefaultHttpClientCache( final long duration, @Nonnull final TimeUnit unit
...
+                .expireAfterWrite(duration, unit)
+                /*  // region 🙈
+                  .evictionListener((key, value, cause) -> {
+                      if( cause.wasEvicted() ) {
+                          value.close();
+                      }
+                  })
+                  // endregion*/
+                .ticker(ticker)
+                .build();
```
newtork, line 76: (Minor)

Revert?

newtork, line 76: (Minor)

Revert?

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
```diff
@@ -101,6 +101,9 @@ private Try<HttpClient> tryGetOrCreateHttpClient(
         catch( final RuntimeException e ) {
             return Try.failure(new HttpClientInstantiationException(e));
         }
+        if( destination != null && httpClient instanceof HttpClientWrapper ) {
+            return Try.success(((HttpClientWrapper) httpClient).withDestination(destination));
+        }
```
newtork, line 106: (Comment)

Not super happy about fixing symptoms instead of cause. But I understand it's the most efficient solution time/effort-wise.

newtork (approved):
Not the best™ solution quality. But a reasonable patch.

PR #219, opened 2024-01-12
fix: [Destinations] Fix Client Certificate Test Flakieness

MatKuhr:
## Context

This PR fixes and simplifies the flaky tests on client certificate authentication. 

Previously the WireMock server would sometimes unexpectedly close the TCP connection for the test case of an incorrectly configured server.

It's unclear why exactly the WireMock server sometimes crashed for this test case. I managed to reproduce the crashes locally without any other tests interfering. But serving on HTTPS without server certificates configured isn't intended anyway.

I felt there was no additional coverage gained on the client side from the behavior, so I simplified the test to use WireMock directly instead of providing a custom Jetty config.


cloudplatform/connectivity-apache-httpclient4/src/test/resources/ClientCertificateAuthenticationLocalTest/README.md
```diff
@@ -7,23 +7,16 @@ The credential files are generated from command line. This process can be automa
 
 * Generate key pair
   ```bash
-  openssl req -newkey rsa:2048 -nodes -keyout cca-client.key.pem -x509 -days 750 -out cca-client.csr.pem -subj "/C=DE/ST=Brandenburg/L=Potsdam/O=SAP/OU=SAP/CN=cca-client"
+  openssl req -x509 -newkey rsa:2048 -utf8 -days 3650 -nodes -config client-cert.conf -keyout client-cert.key -out client-cert.crt
```
MatKuhr, line 10: I took the opportunity to re-generate the certificates with a longer lifetime, they would have expired soon

cloudplatform/connectivity-apache-httpclient5/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ClientCertificateAuthenticationLocalTest.java
```diff
@@ -7,308 +7,128 @@
...
-            sslEngine.setSSLParameters(sslParams);
-            if( super.getWantClientAuth() ) {
-                sslEngine.setWantClientAuth(super.getWantClientAuth());
-            }
-
-            if( super.getNeedClientAuth() ) {
-                sslEngine.setNeedClientAuth(super.getNeedClientAuth());
-            }
-
-            /* <FIX> */
```
MatKuhr, line 259: I was also concerned by this comment being here before 😅

cloudplatform/connectivity-apache-httpclient4/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ClientCertificateAuthenticationLocalTest.java
```diff
@@ -7,311 +7,130 @@
...
-        }
+        // no request successful
+        server.verify(0, getRequestedFor(anyUrl()));
     }
 
-    static class HostNotConfiguredTest extends ClientCertificateAuthenticationLocalTest
+    private static WireMockConfiguration buildWireMockConfiguration()
     {
-        private static final CcaTestConfig TEST_CONFIG =
-            CcaTestConfig.builder().clientKeyStoreFile("cca-client.p12").clientKeyStorePassword("cca-password").build();
```
Johannes-Schneider, line 165: [question]

Just for my understanding (and maybe for future references):

This test initially covered the case where the client's `keyStore` was "misconfigured" in the sense of that it didn't contain a certificate for the actual server host, right?
As a consequence, the client wouldn't send any certificate to authenticate itself, ultimately resulting in the server rejecting the incoming request.

Is that understanding correct?

---

That concrete scenario is now no longer tested (I assume the reason is that we were testing HTTP client implementation specifics here, which shouldn't be our concern - so dropping that test is fair), right?

Johannes-Schneider (commented):
Personally, I very much appreciate the test simplification and I'm fine with dropping the "client is _misconfigured_" scenario.
Nevertheless, I'd like to get at least a second opinion on that matter as the test used to be rather complex.

cloudplatform/connectivity-apache-httpclient4/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ClientCertificateAuthenticationLocalTest.java
MatKuhr: My understanding is that before we were testing a case where the server had no server certificate configured for HTTPS, but I'm not super sure on that (see also my comment [in Slack](https://sap-cpe.slack.com/archives/C01AMF777GA/p1705068559320509?thread_ts=1705063744.420829&cid=C01AMF777GA)). It's hard to tell from the code and even harder to tell from debugging, since the server crashes / kills the connection.

MatKuhr: The reason from the coding is that [this line](https://github.com/SAP/cloud-sdk-java/pull/219/files#diff-5f0644a81d909e5cbadbc6bd5edac44b37ed50087326334e92dd67231e8fb9d2L215) configures the server certificate which the server uses to identify itself (according to [these docs](https://wiremock.org/docs/https/) from WireMock) 

On the other hand, if not configured the server should use self-signed certs which should be accepted by the client due to the "trustAll" setting... so yeah, not really sure honestly

MatKuhr: Summary from our offline discussion on Friday: We don't lose relevant coverage. Also, apparently we can't rely on the `SSLHandshakeException`, so we generalized the assertion and instead verify in addition the test server hasn't crashed.

cloudplatform/connectivity-apache-httpclient5/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ClientCertificateAuthenticationLocalTest.java
```diff
@@ -101,13 +102,12 @@ void testClientNotConfigured()
...
+            .isInstanceOfAny(SSLHandshakeException.class, SocketException.class);
 
-        // keystore methods have been used
-        Mockito.verify(destination).getKeyStore();
-
-        // no request successful
-        server.verify(0, getRequestedFor(anyUrl()));
+        // sanity check: we can't be certain we will get an SSLHandshakeException
+        // sometimes we get a SocketException instead, unknown why
+        // thus we run the success tas case again to verify the server hasn't crashed
```
Johannes-Schneider, line 109: [minor]

Typo

```suggestion
        // thus we run the success test case again to verify the server hasn't crashed
```

Johannes-Schneider (approved):
LGTM

PR #220, opened 2024-01-12
[Java15+] Enable text blocks

newtork:
This. Is. It.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -664,6 +664,7 @@
 						<staticGroups>java,*</staticGroups>
 						<staticAfter>false</staticAfter>
 						<removeUnused>true</removeUnused>
+						<compliance>17</compliance>
 						<excludes>
 							<exclude>**/test/resources/**</exclude>
 						</excludes>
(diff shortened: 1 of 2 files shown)


Johannes-Schneider (approved):
image

**EDIT** Ups... das image turned out bigger than intended 😅🙈

---

As long as the pipeline is happy, I'm happy!

Should we look for more places to use text blocks, or are we gonna change those as "drive-bys" in the future?

newtork:
> or are we gonna change those as "drive-bys" in the future?

Yeah... since actual benefit is low. I would expect changes as optional _drive-by_.

pom.xml
```diff
@@ -664,6 +664,7 @@
 						<staticGroups>java,*</staticGroups>
 						<staticAfter>false</staticAfter>
 						<removeUnused>true</removeUnused>
+						<compliance>17</compliance>
```
newtork, line 667:
``` diff
- <compliance>17</compliance>
+ <compliance>${java.version}</compliance>
```

🙈

PR #223, opened 2024-01-15
Enable `PEM` file format for `ClientCertificateAuthentication`

newtork:
Enable PEM "certificates" for _BTP Destination Service_.

Quick reminder [from cloudflare](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/):

<details><summary>How does TLS work</summary>

image

</details>
<details><summary>How does mTLS work</summary>

image
</details>

The current implementation only applies assumed API usage. No error occurs but actual mTLS usage is not yet checked.

## Definition of Done

- [ ] Tests cover the scope above **(currently: only unit test - no integration tested)**
- [x] ~Documentation updated~
- [ ] Release notes updated


newtork:
Closed in favor of https://github.com/SAP/cloud-sdk-java/pull/225

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
```diff
@@ -178,6 +181,27 @@ void testGetKeyStoreWithPkcs12File()
...
+                .name(MOCKED_DESTINATION_NAME)
+                .property(DestinationProperty.KEY_STORE_LOCATION, fileLocation)
+                .property(DestinationProperty.KEY_STORE_PASSWORD, PEM_KEY_STORE_PASSWORD)
+                .property(
+                    DestinationProperty.CERTIFICATES,
+                    createCertificateJson(fileLocation, fileContent, "CERTIFICATE"))
+                .build();
+
+        final Option<KeyStore> keyStore = new DestinationKeyStoreExtractor(testDestination).getKeyStore();
+        assertThat(keyStore).isNotEmpty();
```
MatKuhr, line 202: maybe we can harden this assertion a bit, checking for some of the certificate details e.g. the common name? I think there is some API on the keystore to access the certificate via its alias..

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractor.java
```diff
@@ -259,6 +263,24 @@ private static KeyStore retrieveExistingKeyStore(
...
+    private static KeyStore createNewKeyStoreFromPem( @Nonnull final String data, @Nullable final String password )
+    {
+        try {
+            final String s = new String(Base64.getDecoder().decode(data), StandardCharsets.UTF_8);
+            final int del = s.indexOf("-----BEGIN CERTIFICATE-----");
+            if( del <= 0 ) {
+                throw new IllegalArgumentException("PEM format cannot be parsed.");
+            }
+            final char[] pw = Strings.isNullOrEmpty(password) ? null : password.toCharArray();
+            final KeyStoreReader reader = KeyStoreReader.builder().alias("1").password(pw).build();
```
MatKuhr, line 276: why alias 1? can we use just any identifier or does it actually matter?

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractor.java
```diff
@@ -259,6 +263,24 @@ private static KeyStore retrieveExistingKeyStore(
         }
     }
 
+    @Nonnull
+    private static KeyStore createNewKeyStoreFromPem( @Nonnull final String data, @Nullable final String password )
+    {
+        try {
+            final String s = new String(Base64.getDecoder().decode(data), StandardCharsets.UTF_8);
+            final int del = s.indexOf("-----BEGIN CERTIFICATE-----");
```
MatKuhr, line 271:
```suggestion
            final int delimiter = s.indexOf("-----BEGIN CERTIFICATE-----");
```

(minor) some variable names could probably be improved a bit.

Also I am wondering: This currently assumes the key always comes first, right? is this guaranteed to be the case, or could this potentially be an implementation detail of the destination service? If in doubt maybe we just check explicitly?

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractor.java
```diff
@@ -43,10 +46,11 @@ class DestinationKeyStoreExtractor
     // See the supported key store file extensions:
     // https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/df1bb55a526942b9bee78fea2ebb3162.html
     //Mapping file extension to key store types
-    private static final Map<String, String> SUPPORTED_KEY_STORE_TYPES_AS_KEY_STORE = ImmutableMap.of(
-            "pfx", "PKCS12",
-            "p12", "PKCS12",
-            "jks", "JKS"
+    private static final Map<String, BiFunction<String, String, KeyStore>> SUPPORTED_KEY_STORES = ImmutableMap.of(
+        "pfx", (cert,pw) -> retrieveExistingKeyStore(cert, pw, "PKCS12"),
```
MatKuhr, line 50:
```suggestion
        "pfx", (ks,pw) -> retrieveExistingKeyStore(ks, pw, "PKCS12"),
```

(minor) just an idea to clarify what these strings represent for the 3 keystore cases..

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractX509SslContextProvider.java
```diff
@@ -35,6 +21,8 @@
  */
 abstract class AbstractX509SslContextProvider implements PlatformSslContextProvider
```
MatKuhr, line 22: I am thinking if we still need this class in the future. We could simply always attach the certificates to the destination and get rid of this additional mechanism. 

But then we would have to solve the problem of certificate rotation. This doesn't affect the certificates from destination service, but programatically created destinations or ones created from service bindings. There we would have to re-create the destination (latest) once the certificate rotates. So not sure if that's a good idea..

Anyway, not really related to this PR, just brainstorming..

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/KeyStoreReader.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Builder.Default
+    @Nonnull
+    final Function<Throwable, Certificate[]> fallbackCertificates = ( e ) -> {
+        throw new CloudPlatformException("Provided certificate data did not contain any valid X.509 certificates.", e);
+    };
+
+    @Builder.Default
+    @Nonnull
+    final Function<Throwable, PrivateKey> fallbackPrivateKey = ( e ) -> {
```
MatKuhr, line 63: (minor) why do we have a default for this if it's non null?

MatKuhr (commented):
overall approach looks very promising 👍🏻

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractor.java
newtork: _> This currently assumes the key always comes first, right?_

Yes.

_> is this guaranteed to be the case, or could this potentially be an implementation detail of the destination service?_

Absolutely implementation detail. AFAIK, PEM is completely unordered.

_> If in doubt maybe we just check explicitly?_

Yeah.. would be better, but I was wondering how much effort to expect for enabling PEM support.

This is why I would prefer more integration tests, i.e. manually create PEM files with different orders, with fewer entries, with(out) private-key, etc.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractor.java
newtork: _> why alias 1? can we use just any identifier or does it actually matter?_

"1" is used in the other places as well, no idea why. I think (hope) the TLS/SSL component has some kind of heuristic that tries certificates independent from any alias.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/KeyStoreReader.java
newtork: I was considering to replace the default "fallbackPrivateKey" with an exception thrower of type `DestinationAccessException` instead of `CloudPlatformException`. But I didn't take advantage of that. I'll remove it.

newtork: TODO: change hardcoded. use `DestinationAccessException`.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/KeyStoreReader.java
```diff
@@ -0,0 +1,129 @@
...
+import lombok.NoArgsConstructor;
+
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+class KeyStoreReader
+{
+    @Builder.Default
+    @Nonnull
+    final String alias = "instance-identity";
```
newtork, line 49: TODO: change stateful class to utility class with static method.

PR #230, opened 2024-01-18
chore: [DependencyUpdate] Update XSUAA lib to 3.3.3 (Reloaded)

newtork:
- [x] Update XSUAA lib to 3.3.3
- [x] Release notes

Challenges:
* `java-security-test` brings "Jetty 12" and relies on the API on runtime.
* `wiremock` depends on "Jetty 11" and `wiremock-standalone` shades(!) "Jetty 11".
* `wiremock` + `java-security-test` ("Jetty 12") is not supported
* `wiremock-standalone` + `java-security-test` ("Jetty 12") works, but compatibility is unclear. Accepted risk.


cloudplatform/connectivity-destination-service/pom.xml
```diff
@@ -228,6 +217,7 @@
 					<ignoredUsedUndeclaredDependencies>
 						<!-- include resilience via the aggregator to ensure the default implementation is present -->
 						<ignoredUsedUndeclaredDependency>com.sap.cloud.sdk.cloudplatform:resilience-api</ignoredUsedUndeclaredDependency>
+						<ignoredUsedUndeclaredDependency>org.wiremock:wiremock-standalone</ignoredUsedUndeclaredDependency>
```
newtork, line 220: Relying on transitive `wiremock-standalone` from `java-security-test` is not good.

However declaring it explicitly with a specific version, will risk future incompatibilities with `java-security-test` 🤷‍♂️

CharlesDuboisSAP: Could you add it as a comment above the <ignore>?

cloudplatform/connectivity-destination-service/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
newtork, line 1: (Major)

Please move file to non-productive code.

KavithaSiva:
I would also additionally want the `jetty-bom` in our archetypes to be updated to `12.0.5` .

KavithaSiva (changes requested):
I would also additionally want the `jetty-bom` in our archetype to be updated to `12.0.5`  with a compatibility note that people adding `wiremock` would have to use `wiremock-standalone` for it to work.

cloudplatform/connectivity-destination-service/src/main/webapp/WEB-INF/web.xml
newtork: image

File does not need to be moved. It's not part of the packaged JAR!

newtork:
> I would also additionally want the `jetty-bom` in our archetype to be updated to `12.0.5` with a compatibility note that people adding `wiremock` would have to use `wiremock-standalone` for it to work.

It's working :+1:
https://sdk-v2.cpe.c.eu-de-1.cloud.sap/job/end-to-end-tests/view/Cloud%20SDK%20v5/job/v5-scp-cf-spring-xsuaa/

PR #232, opened 2024-01-18
feat: Extend AuthToken Forwarding

cschubertcs:
## Context

SAP/cloud-sdk-java-backlog#401

This PR allows destinations retrieved from the Destination Service with Authentication Type `NO_AUTHENTICATION` to evaluate the `forwardAuthToken` parameter.

This means any incoming JWT will be forwarded to the target of the destination if the `forwardAuthToken` is set to true and the Authentication Type is either unset or set to `NO_AUTHENTICATION`.

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ComplexDestinationPropertyFactory.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ComplexDestinationPropertyFactory.java
@@ -184,34 +184,29 @@ class ComplexDestinationPropertyFactory
         @Nonnull final DestinationProperties baseProperties,
         @Nonnull final Option<BasicCredentials> basicCredentials )
     {
-        final Option<AuthenticationType> authType =
+        final AuthenticationType authType =
             baseProperties
                 .get(DestinationProperty.AUTH_TYPE)
-                .orElse(() -> baseProperties.get(DestinationProperty.AUTH_TYPE_FALLBACK));
+                .orElse(() -> baseProperties.get(DestinationProperty.AUTH_TYPE_FALLBACK))
+                .onEmpty(
+                    () -> log
+                        .debug(
+                            "No valid JSON primitive '{}' or '{}' defined. Falling back to {}.",
+                            DestinationProperty.AUTH_TYPE,
+                            DestinationProperty.AUTH_TYPE_FALLBACK,
+                            AuthenticationType.NO_AUTHENTICATION))
+                .getOrElse(AuthenticationType.NO_AUTHENTICATION);
 
         final boolean forwardAuthToken = baseProperties.get(DestinationProperty.FORWARD_AUTH_TOKEN).getOrElse(false);
 
-        if( (authType.isEmpty() || authType.get() == AuthenticationType.NO_AUTHENTICATION)
-            && basicCredentials.isDefined() ) {
+        if( authType == AuthenticationType.NO_AUTHENTICATION && basicCredentials.isDefined() ) {
             return AuthenticationType.BASIC_AUTHENTICATION;
         }
 
-        if( (authType.isEmpty() || authType.get() == AuthenticationType.NO_AUTHENTICATION) && forwardAuthToken ) {
+        if( authType == AuthenticationType.NO_AUTHENTICATION && forwardAuthToken ) {
             return AuthenticationType.TOKEN_FORWARDING;
         }
-
-        if( authType.isEmpty() ) {
-            final String msg = "No valid JSON primitive '{}' or '{}' defined. Falling back to {}.";
-            log
-                .debug(
-                    msg,
-                    DestinationProperty.AUTH_TYPE,
-                    DestinationProperty.AUTH_TYPE_FALLBACK,
-                    AuthenticationType.NO_AUTHENTICATION);
-
-            return AuthenticationType.NO_AUTHENTICATION;
-        }
-        return authType.get();
+        return authType;
     }
 
     @Nonnull
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ComplexDestinationPropertyFactory.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ComplexDestinationPropertyFactory.java
@@ -189,21 +189,27 @@ class ComplexDestinationPropertyFactory
                 .get(DestinationProperty.AUTH_TYPE)
                 .orElse(() -> baseProperties.get(DestinationProperty.AUTH_TYPE_FALLBACK));
 
+        final boolean forwardAuthToken = baseProperties.get(DestinationProperty.FORWARD_AUTH_TOKEN).getOrElse(false);
+
+        if( (authType.isEmpty() || authType.get() == AuthenticationType.NO_AUTHENTICATION)
+            && basicCredentials.isDefined() ) {
+            return AuthenticationType.BASIC_AUTHENTICATION;
+        }
+
+        if( (authType.isEmpty() || authType.get() == AuthenticationType.NO_AUTHENTICATION) && forwardAuthToken ) {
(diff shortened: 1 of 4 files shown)


cschubertcs:
Some more tests for the destination service case and test-refactoring required.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
```diff
@@ -0,0 +1,174 @@
...
+import com.sap.cloud.sdk.cloudplatform.resilience.ResilienceConfiguration;
+
+public class ForwardAuthTokenTest
+{
+    static Collection<TestCase> createTestCases()
+    {
+        return List
+            .of(
+                // forwardAuthToken = true && authType = NoAuthentication
+                new TestCase(
```
MatKuhr, line 37: (minor) I feel this is a bit hard to read. Maybe consider switching to a builder as done [here](https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommandWithoutAllDestinationsTest.java#L100)? 

e.g.
```suggestion
                forProperties(DestinationProperty.FORWARD_AUTH_TOKEN, "true")
                        .and(DestinationProperty.AUTH_TYPE, NO_AUTHENTICATION)
                        .expectTokenForwarding()
```

where `forProperties` returns a builder for `TestCase` and `expectTokenForwarding()` as well as `expectTokenIsNotForwarded()` return `TestCase`.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
```diff
@@ -0,0 +1,174 @@
...
+                    Map
+                        .of(
+                            DestinationProperty.FORWARD_AUTH_TOKEN.getKeyName(),
+                            "true",
+                            DestinationProperty.AUTH_TYPE.getKeyName(),
+                            AuthenticationType.NO_AUTHENTICATION),
+                    AssertionType.TOKEN_IS_FORWARDED),
```
MatKuhr, line 44: (minor) building on the builder idea (pun intended), you could omit the type `AssertionType` because the explicit method names `expectTokenForwarding()` and `expectTokenIsNotForwarded()` bring the readability (which these constants are intended for I presume)

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
```diff
@@ -0,0 +1,174 @@
...
+                ResilienceConfiguration.empty("testMultiple"));
+
+        doReturn(serviceResponse).when(adapter).getConfigurationAsJson(eq("/destinations/" + destinationName), any());
+
+        return loader.tryGetDestination(destinationName).get().asHttp();
+    }
+
+    private String buildServiceResponse( final String destinationName, final Map<String, Object> properties )
+    {
+        final JsonElement response = JsonParser.parseString("""
```
MatKuhr, line 130: (minor) We could consider stubbing `DestinationServiceV1Response` with `.fromDestinationServiceV1Response` instead. Might be a bit simpler

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @ParameterizedTest
+    @MethodSource( "createTestCases" )
+    void destinationServiceDestinationShouldFulfillTestCase( @Nonnull final TestCase testCase )
+    {
+        final HttpDestination destination = buildDestinationServiceDestination(testCase.properties);
+        testCase.forwardingAssertion.assertion.accept(destination);
+    }
+
+    private HttpDestination buildDestinationServiceDestination( final Map<String, Object> properties )
```
MatKuhr, line 105: (minor) what additional coverage do we gain from this compared to `localDestinationShouldFulfillTestCase`?

For this test setup the input data for the assertion ends up being the same I think?

MatKuhr (commented):
with release note LGTM

some minor (mostly test related) comments 

we could consider adding one further explicit test asserting that the actual property defined in the map always remains constant. With our immutability this should always be the case, but just to indicate this is intended and expected.

Background: if we were to ever change that the change detection for the destination service cache would potentially break

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
cschubertcs: I think I would keep them, as otherwise the overview of tests ran in the IDE just shows the method reference/lambda, from which you don't know what the expectation is.

Here how it looks now, imagine the `forwardingAssertion=<gibberish>` instead^^
grafik

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
cschubertcs: I had the problem that there was some caching in the destinationLoader which caused unexpected behavior.
So, in general there are different call stacks (at least to some extent), so would like to keep it to stay safe.

MattKuhr: I mean yes, the call stack is different, thus the additional caching for example. But there is no logic related to setting the auth type in the destination service implementation, right? I would think even there must not be, otherwise we probably mess with the change detection.

In the spirit of unit testing we could consider strengthening `DestinationServiceFactoryTest` a bti, asserting the porperty is set correctly, and move this test into the baseline connectivity module

cschubertcs: Actually true, initially the stack looked completely different, but besides some certifacate/authtoken handling it's the same. I created a test for the property in the `DestinationServiceFactoryTest` and removed it from my test.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ComplexDestinationPropertyFactory.java
```diff
@@ -184,28 +184,29 @@ AuthenticationType getAuthenticationType(
...
-
-            final String msg = "No valid JSON primitive '{}' or '{}' defined. Falling back to {}.";
-            log.debug(msg, DestinationProperty.AUTH_TYPE, DestinationProperty.AUTH_TYPE_FALLBACK, fallback);
+        if( authType == AuthenticationType.NO_AUTHENTICATION && basicCredentials.isDefined() ) {
+            return AuthenticationType.BASIC_AUTHENTICATION;
+        }
 
-            return fallback;
+        if( authType == AuthenticationType.NO_AUTHENTICATION && forwardAuthToken ) {
+            return AuthenticationType.TOKEN_FORWARDING;
```
newtork, line 207: (Minor/Nitpick)

Currently you would log:

No valid JSON primitive 'AUTH_TYPE' or 'AUTH_TYPE_FALLBACK' defined. Falling back to NO_AUTHENTICATION.",

But the resulting authentication type could result in "Basic" or "Token Forwarding".
This is/could be a bit misleading. I think would suggest to add a debug statement here and above too.

MatKuhr (approved):

PR #233, opened 2024-01-19
fix: [Destinations] Improve HttpClient 5 Cache Builder API

MatKuhr:


cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ApacheHttpClient5CacheBuilder.java
```diff
@@ -62,7 +62,7 @@ public ApacheHttpClient5CacheBuilder duration( @Nonnull final Duration duration
      * @return The new {@link ApacheHttpClient5Cache} instance.
      */
     @Nonnull
-    public DefaultApacheHttpClient5Cache build()
+    public ApacheHttpClient5Cache build()
```
MatKuhr, line 65: `DefaultApacheHttpClient5Cache` is NOT public, but `ApacheHttpClient5Cache` is

PR #234, opened 2024-01-19
Add test case for empty string PKCS12 keystore

newtork:
Related:
https://github.com/SAP/cloud-sdk-java-backlog/issues/295
https://github.com/SAP/cloud-sdk-java-backlog/issues/331

**This is the KeyStore API for reading:**
```java
KeyStore keyStore;
keyStore.load(InputStream is, char[] password);
keyStore.getKey(String alias, char[] password);
keyStore.getCertificate(String alias);
```

Notice:
* KeyStore API has no password on certificates!
* KeyStore API is format independent.
* Some tools like `openssl` have stronger password requirements (e.g. min length) - programmatically the Java API is more lenient.

**JKS:**
* Loading the KeyStore can be done without password (=`null`).
* JKS `KeyStore#load` "password" is only used for **optional integrity check**
* JKS `KeyStore#getKey` "password" is required for key decryption.

```java
KeyStore keyStore = KeyStore.getInstance("JKS");
keyStore.load(InputStream is, null);
```

**PKCS12:**
* Loading the KeyStore must be done with password (not `null`).
* PKCS12 `KeyStore#load` "password" is **required for certificates decryption**.
* PKCS12 `KeyStore#getKey` "password" is required for key decryption.

```java
KeyStore keyStore = KeyStore.getInstance("JKS");
keyStore.load(InputStream is, new char[0]);
```


newtork:
Test cases are no longer needed.
We don't need to unit-test `KeyStore` API.

PR #196, opened 2023-12-18
Re-Added `UriBuilder.build` but deprecated

CharlesDuboisSAP:
## Context
[[Fix] Invalid Destination Domain Throws](https://github.com/SAP/cloud-sdk-java/pull/189#top) removed a public method by accident. This PR re-adds the deleted method but deprecated.
I manually edited UriPathMerger to use the deprecated method and manually ran the UriPathMergerTest👍

### Feature scope:

- [x] Remove the breaking change

## Definition of Done

- [x] Functionality scope stated & covered
- [x] Tests cover the scope above (manually tested)
- [x] Error handling created / updated & covered by the tests above
- [x] ~Documentation updated~
- [x] Release notes updated

PR #235, opened 2024-01-22
feat: Introduce new GitHub Actions workflow for release preparation

CharlesDuboisSAP:
This PR introduces a new GitHub Actions workflow for preparing a new release of the SAP Cloud SDK for Java. The new workflow is designed to automate the process of bumping the version number, running continuous integration, creating a GitHub release, and creating PRs for JavaDocs and release notes updates. The PR also includes several new GitHub Actions for specific tasks within the workflow, such as waiting for a workflow to complete and triggering a workflow.

The changes in this PR are related to the following backlog items: [#318](https://github.com/SAP/cloud-sdk-java-backlog/issues/318) and [#196](https://github.com/SAP/cloud-sdk-java-backlog/issues/196)

## Changes

- Added a new GitHub Actions workflow for preparing a release.
- Added several new GitHub Actions for specific tasks within the workflow.
- Updated the release notes template.
- Removed the old GitHub Actions workflow for building the project.

## Test run

- [Workflow run](https://github.com/Johannes-Schneider/cloud-sdk-java/actions/runs/7608856197)
- [PR created](https://github.com/Johannes-Schneider/cloud-sdk-java/pull/19)

## Notes for Reviewers

- Please review the new workflow and the associated actions to ensure they are correctly configured.
- Please review the changes to the release notes template to ensure they are appropriate.
- Please check the different git config accounts, could it be streamlined?


release_notes_template.md
```diff
@@ -0,0 +1,30 @@
+## 5.X.0-SNAPSHOT
```
MatKuhr, line 1: (minor) Maybe move this to the scripts directory or something so that it doesn't clutter our root dir

.pipeline/scripts/release_notes_automation.py
```diff
@@ -0,0 +1,151 @@
...
+            file = remove_unchanged_sections(file, unchanged_sections)
+            file = set_header(file, args.version)
+            file = link_github_release(file, args.version)
+            file = clean_up_dependency_updates(file)
+
+            release_notes_list = os.listdir(args.folder)
+            target_file = find_target_file(args.folder, release_notes_list)
+            write_release_notes(args.folder, target_file)
+
+    except KeyboardInterrupt:
```
MatKuhr, line 150: what is this for?

.pipeline/scripts/release_notes_automation.py
```diff
@@ -0,0 +1,151 @@
...
+    with open(file_name, 'r', encoding="utf8") as file:
+        return file.read()
+
+
+def write_file(file_name, data):
+    with open(file_name, 'w', encoding="utf8") as file:
+        file.write(data)
+
+unchanged_sections = [
+"""### 🚧 Known Issues
```
MatKuhr, line 17: I think this only works if we never add any whitespaces anywhere. It may be a bit safer to explicitly split the string by the 4 headlines and then checking if there is anything non-whitespacy in there..

.pipeline/scripts/release_notes_automation.py
```diff
@@ -0,0 +1,151 @@
...
+
+""",
+"""### ✨ New Functionality
+
+- 
+
+""",
+"""### 📈 Improvements
+
+- Dependency Updates:
```
MatKuhr, line 34: Not super sure if we want to keep this format as we have some auto-generated content with the introduction of #236. TBD..

.pipeline/scripts/release_notes_automation.py
```diff
@@ -0,0 +1,151 @@
...
+    file = re.sub(other_dependency_update, "", file)
+    # smaller strings
+    file = re.sub(major_dependency_update, "", file)
+    file = re.sub(minor_dependency_update, "", file)
+    # smallest leftovers
+    file = re.sub(empty_dependency_update, "", file)
+    return file
+
+releases_pattern = re.compile(r"^## ")
+def count_releases(filename):
```
MatKuhr, line 90: seems unnecessary, we could evaluate the version string and extract the patch version, right?

.pipeline/scripts/release_notes_automation.py
```diff
@@ -0,0 +1,151 @@
...
+def find_target_file(folder, filenames):
+    pattern = re.compile(r"^release-notes-\d+-to-\d+.mdx$")
+    highest_release_notes = 0
+
+    for file in filenames:
+        absoulte_path = os.path.join(folder, file)
+        print(file)
+        # if file matches release-notes-X-to-Y.mdx and has less than 15 releases, return its name
+        if pattern.match(file) and os.path.isfile(absoulte_path):
+            if count_releases(absoulte_path) < 15:
```
MatKuhr, line 107: this will only work once, but if we ever get across 5.30 it no longer works. Since it would only require manual file creation once every 15 or so releases it might not be worth the effort / complexity to automate this step

.pipeline/scripts/release_notes_automation.py
CharlesDuboisSAP: It was copy pasted from `set-release-versions.py` Do you want it removed?

.pipeline/scripts/release_notes_automation.py
CharlesDuboisSAP: Not if we do milestone or hotfix releases

.pipeline/scripts/release_notes_automation.py
CharlesDuboisSAP: The script creates the file, I tested it

.pipeline/scripts/release_notes_automation.py
CharlesDuboisSAP: I couldn't figure out a simple way of doing it, plus if it breaks it leaves an empty title for 1 release, which is not a failure since you have to review the end result anyway

.pipeline/scripts/release_notes_automation.py
CharlesDuboisSAP: removed

.pipeline/scripts/release_notes_automation.py
CharlesDuboisSAP: removed

.github/workflows/delete-stale-caches.yaml
```diff
@@ -0,0 +1,24 @@
+name: "Delete Stale Caches"
```
KavithaSiva, line 1: [Major] Workflow is incomplete, doesn't clear caches yet.

.github/workflows/prepare_release.yml
```diff
@@ -1,447 +1,399 @@
...
-      [
-        prepare-release,
-        build,
-        archetypes,
-        codeql,
-        create-release,
-        scan,
-        create_fosstars_report,
-      ]
+    needs: [ bump-version, run-ci, create-release, create-docs-pr, create-release-notes-pr, create-code-pr ]
```
KavithaSiva, line 352: I guess the intention was to handle failure, if they occur in any of these jobs.

But, if you add them in needs I think this block would only run if all the jobs run successfully.

.github/actions/scan-with-blackduck/action.yaml
```diff
@@ -0,0 +1,37 @@
...
+      run: echo "MAJOR_VERSION=$(cat latest.json | jq -r .version | cut -d '.' -f 1)" >> $GITHUB_OUTPUT
+      shell: bash
+
+    - name: Determine Maven Excludes
+      id: get-maven-excludes-for-blackduck
+      run: python .pipeline/scripts/get-maven-excludes.py --filter-key excludeFromBlackDuckScan --filter-value True
+      shell: bash
+
+    - name: BlackDuck Scan
+      uses: SAP/project-piper-action@f17ad180226247ef7f45e40bcb5b20b222b8c04c
```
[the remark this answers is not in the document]
Johannes-Schneider: Yes, it should be possible to use a branch name instead of a commit id here.
I'm just not sure whether we actually want that - if the action changes in an unexpected way, our pipeline might start failing.

.github/actions/scan-with-blackduck/action.yaml
```diff
@@ -0,0 +1,37 @@
...
+          --version=$PROJECT_VERSION \
+      env:
+        PIPER_token: ${{ secrets.BLACKDUCK_TOKEN }}
+        DETECT_MAVEN_EXCLUDED_MODULES: ${{ steps.get-maven-excludes-for-blackduck.outputs.EXCLUDES }}
+        DETECT_MAVEN_BUILD_COMMAND: -pl ${{ steps.get-maven-excludes-for-blackduck.outputs.PREFIXED_EXCLUDES }}
+        DETECT_TIMEOUT: 7200
+        PROJECT_VERSION: ${{ steps.get-major-version.outputs.MAJOR_VERSION }}
+
+    - name: Print Action End
+      if: always()
```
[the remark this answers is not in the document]
Johannes-Schneider: This files contains a reusable action - not a workflow.
The idea is that this piece of code could be "embedded" in other workflows - some might want to let us know that the black duck scan failed, others might not.

For example, the daily (i.e. cron) Black Duck scan job should let us know (as it is only responsible for calling this action).
The `Prepare Release` job, however, also runs a Black Duck scan but shouldn't send a Slack notification about the Black Duck scan - instead, we would most probably rather want to see an "aggregated" failure report in our Slack channel.

.github/workflows/prepare_release.yml
```diff
@@ -1,447 +1,399 @@
-name: Prepare Release
+name: "Prepare Release"
 
 on:
   workflow_dispatch:
     inputs:
-      release_version:
-        description: "Release version, leave empty for default value based on current snapshot"
+      branch:
```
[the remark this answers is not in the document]
Johannes-Schneider: The initial idea here was that there might be other production branches (i.e. `main/v6`) in the future.
But I do agree: the chances for that are rather low, so hard-coding `main` as the default here makes sense.

.github/workflows/prepare_release.yml
Johannes-Schneider: Generally: Yes.
In this case, however, we do have the `if: ${{ failure() }}` (two lines further down), which will let the job run if any of the prerequisites failed.

Johannes-Schneider:
Continued in #259

PR #248, opened 2024-01-23
Destination: Fix certificate access to KeyStore EMPTY-PW PKCS12 file

newtork:
Related:
https://github.com/SAP/cloud-sdk-java-backlog/issues/295
https://github.com/SAP/cloud-sdk-java-backlog/issues/331

When loading a PKCS#12 password-less keystore, it looks like...

<details><summary>(1) we didn't fully setup test assertions.</summary>

image

</details>
<details><summary>(2) we incorrectly used `null` instead of provided empty String.</summary>

image

</details>
<details><summary>(3) we incorrectly used `null` instead of `new char[0]` as fallback.</summary>

image

</details>

---

An overview of KeyStore API behavior:

| |  password  | == NULL | |  | == "" EMPTY |  
-- | -- | -- | -- |  -- |  -- | --
format  |  file | KS load | KS read key | KS read cert | KS load | KS read key
JKS | (jks,pfx) | ✔️ no verification | ✔️ no/"" pw set <br>❌ otherwise | ✔️ | ✔️ correct pw<br>❌ otherwise | (SDK)<br>✔️ key pw = KS pw<br>❌ otherwise
PKCS12 | (p12) | ✔️ no verification  | ✔️ no/"" pw set <br>❌ otherwise | ⚠️  **empty** | ✔️ correct pw<br>❌ otherwise | (SDK) <br>✔️ key pw = KS pw<br>❌ otherwise

---

> _Why was this not solved ~3y ago in the old repo?_

Because while working on [the support issue](https://github.wdf.sap.corp/MA/sdk/issues/5998) I incorrectly assumed using `key-store-password = null` would solve broken PKCS#12 decryption, when in fact it actually disabled it silently. Instead of doing that, we should've reacted with dismissal:

_When the provided file is not proper PKCS#12 format, then it's not up to us to solve it._

Unfortunately we "fixed" it with a bug. And since we skipped proper testing, it didn't come up again. Yikes.
At least no one complained again. 🤷‍♂️


cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
```diff
@@ -367,6 +374,9 @@ void testGetKeyStoreWithPkcs12FileEmptyPassword()
 
         final Option<KeyStore> keyStore = new DestinationKeyStoreExtractor(destination).getKeyStore();
         assertThat(keyStore).isNotEmpty();
+        assertThat(Collections.list(keyStore.get().aliases())).containsExactly("1");
+        assertThat(keyStore.get().getKey("1", null)).isInstanceOf(PrivateKey.class);
```
MatKuhr, line 378: why does null work in this case? I thought this always throws? 😮

MatKuhr (commented):
Release note would be needed.

Also, I think we can update the signature of `KeyStoreReader.java`

```diff
- static PrivateKey loadKey( @Nonnull final Reader keyReader, @Nullable final char[] password )
+ static PrivateKey loadKey( @Nonnull final Reader keyReader, @Nonnull final char[] password )
```

Finally, I think the case PEM file without password (i.e. that line `DestinationKeyStoreExtractor#287` produces the desired result) remains untested, right? Or is this tested elsewhere somewhere?

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
newtork: (This whole API feels _legacy_. And so confusing.)

`KeyStore#getKey(String alias, char[] pass == null )` is allowed when no password (= empty password) is set.

**However** the method we use **for storing** entries is `KeyStore#setKeyEntry(String alias, Key key, char[] pass, Certificate[] certs)`. This does **not** allow for `pass == null`. It'll throw.

For our sanity I added the case `getKey("1", new char[0])`.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
```diff
@@ -302,10 +304,15 @@ void testGetKeyStoreWithPkcs12FileNoPassword()
 
         final Option<KeyStore> keyStore = new DestinationKeyStoreExtractor(destination).getKeyStore();
         assertThat(keyStore).isNotEmpty();
+        assertThat(Collections.list(keyStore.get().aliases())).containsExactly("1");
+        assertThat(keyStore.get().getKey("1", null)).isInstanceOf(RSAPrivateCrtKey.class);
```
MatKuhr, line 308: Isn't this in conflict with this finding from [the BLI comment](https://github.com/orgs/SAP/projects/62/views/1?pane=issue&itemId=52367860):

> * PKCS12 (p12)
>     * password provided null
>        * get certificates -> [] no certificates are being loaded <-- THIS CURRENT BEHAVIOR LEADS TO PROBLEMS

MatKuhr (commented):
I think this justifies a small release note

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
newtork: You're right, I updated the BLI comment / observation. 😅

MatKuhr (approved):
LGMT 😄

PR #256, opened 2024-01-25
feat: Extend forward authtoken further

cschubertcs:
## Context
 

SAP/cloud-sdk-java-backlog#401

This PR extends the token forwarding even further than the inital PR #232.

Now it also evaluates the `HTML5.ForwardAuthToken`, which was initialy evaluated/expected by the approuter.
That way destinations can be reused between the Approuter and SDK without duplicating properties.

Related approuter documentation: https://www.npmjs.com/package/@sap/approuter#optional-additional-properties
## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ForwardAuthTokenTest.java
```diff
@@ -40,6 +40,26 @@ static Collection<TestCase> createTestCases()
                 TestCaseBuilder
                     .forProperty(DestinationProperty.FORWARD_AUTH_TOKEN.getKeyName(), "false")
                     .expectNoTokenForwarding(),
+
+                // HTML5.ForwardAuthToken = true && authType = NoAuthentication
+                TestCaseBuilder
+                    .forProperty(DestinationProperty.APPROUTER_FORWARD_AUTH_TOKEN.getKeyName(), "true")
+                    .and(DestinationProperty.AUTH_TYPE.getKeyName(), AuthenticationType.NO_AUTHENTICATION)
+                    .expectTokenForwarding(),
```
Johannes-Schneider, line 48: What about the case where `HTML5.ForwardAuthToken` is set to `true` but we also have a authentication type other than `NO_AUTHENTICATION` (I.e. we would be ignoring the forward auth token flag)?

This case also seems to be uncovered for the "regular" `forwardAuthToken` property.

MatKuhr: There might be cases where this would actually make sense, e.g. client certificate auth (i.e. mTLS). Or in potentially further cases where the `authorization` header is not occupied by whatever the auth type does.

Maybe we allow it for mTLS but nothing else for now until we get some actual usage examples / requests..

cschubertcs: Okay, if client certificate auth is viable use case for token forwarding we need to rethink the solution.
Right now the logic is to set the `AuthType` from `NO_AUTHENTICATION` to `TOKEN_FORWARDING`. If we do the same for `CLIENT_CERTIFICATE_AUTHENTICATION` we might get into problems later on, where the type is expected to be `CLIENT_CERTIFICATE_AUTHENTICATION` and not `CLIENT_CERTIFICATE_AUTHENTICATION`.

How necessary is that at the moment? Does it make sense to build a different solution to keep the original auth type and allow for token forwarding without setting the auth type?

MatKuhr: I would argue not super necessary, as this is would only be convenience. With the ability to modify destinations via the builder API adding an authorization header isn't that difficult anymore...

If you see an easy/obvious improvement to make the behavior more consistent, sure, otherwise I'd probably leave it as is..

cschubertcs: Yeah, not really without changing the current (an past) logic.

We could introduce a new "property" on a destination which states whether the auth token should be forwarded (which doesn't rely on the AuthType). To be clean we would need to remove/deprecate the `FORWARD_TOKEN` auth type.

I think I would park this approach for the moment until it's actually needed.

Johannes-Schneider (approved):
LGTM

PR #257, opened 2024-01-25
feat: [Destinations] Improve Certificate Loading for SAML Destinations

MatKuhr:
## Context

Loosely related to SAP/cloud-sdk-java-backlog#295.

Improve how certificates are evaluated for the authentication types `SAMLAssertion` and `OAUTH2_SAML_BEARER_ASSERTION`. For these auth types the certificates are used to sign the SAML assertion. Defining separate/additional certificates for mTLS is not possible for these destinations (judging by the options declared in [the docs](https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/oauth-saml-bearer-assertion-authentication?q=Destination%20Service)).

With this PR the certificates are no longer unnecessarily parsed for this case while obtaining the destination.

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
```diff
@@ -73,64 +67,29 @@ class DestinationKeyStoreExtractorTest
...
-                        "tokenServiceUrl":"https://apisalesdemo2.successfactors.eu:443/oauth/token",
-                        "KeyStore":"key-store",
-                        "URL":"https://apisalesdemo2.successfactors.eu:443",
-                        "Name":"sfapi_dest",
-                        "Type":"HTTP",
-                        "companyId":"comepany-id",
-                        "XFSystemName":"SFSF_SalesDemo",
-                        "KeyStoreLocation":"%s",
-                        "clientKey":"client-key",
-                        "Authentication":"OAuth2SAMLBearerAssertion",
```
MatKuhr, line 102: Broken test setup, keystrore should not be loaded for this auth type at all.

I don't see what this test was meant to cover that isn't covered by `testGetKeyStoreWithPkcs12File`, thus deleting it.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
```diff
@@ -144,22 +103,18 @@ void testGetKeyStoreWithJksFile()
             DefaultHttpDestination
                 .builder(VALID_URI)
                 .name(MOCKED_DESTINATION_NAME)
-                .property(DestinationProperty.TRUST_STORE_LOCATION, fileLocation)
-                .property(DestinationProperty.TRUST_STORE_PASSWORD, KEY_STORE_PASSWORD)
+                .property(DestinationProperty.KEY_STORE_LOCATION, fileLocation)
```
MatKuhr, line 106: not sure how this test ever passed

release_notes.md
```diff
@@ -16,6 +16,8 @@
 
 ### 📈 Improvements
 
+- Improve how certificates are evaluated for the authentication types `SAMLAssertion` and `OAUTH2_SAML_BEARER_ASSERTION`.
+  The certificates are no longer unnecessarily parsed when obtaining the destination.
```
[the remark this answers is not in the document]
MatKuhr: Mostly correct, aside from some mix-ups of `TrustStore`, this PR exclusively affects the `.keyStore` property of the destinations.

I discussed this with @newtork earlier this week, because this change is sort of related to fixing #44 which we intend to also do in a follow-up PR.

I would not consider (1) a problem, too unlikely that someone used it that way IMO. For (2) I would consider the previous behavior a bug (although somewhat unlikely that anybody stumbles upon this). In fact, with this change we would make it easier for users to enable mTLS for SAML destinations by adding the relevant certificate via the destination builder API. Previously, one would have to also change the auth type to anything non-SAML, which I would consider a rather hacky workaround.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
MatKuhr: I think on main it is mixing up trust and keystone. It's setting the keystore but asserting on the trust store.

So I don't see how this change is removing any test coverage?

newtork: Args! I thought we'd support JKS (and other) formats for _trust store_, while in fast we are only accepting single x509 certificate files (crt, cer, der). My complaint is invalid.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreExtractorTest.java
```diff
@@ -73,64 +67,29 @@ class DestinationKeyStoreExtractorTest
...
+            DefaultHttpDestination
+                .builder(URI.create("foo.com"))
+                .authenticationType(authType)
+                .property(DestinationProperty.KEY_STORE_LOCATION, "foo")
+                .build();
+        final DestinationKeyStoreExtractor sut = new DestinationKeyStoreExtractor(destination);
+
+        switch( authType ) {
+            case SAML_ASSERTION, OAUTH2_SAML_BEARER_ASSERTION -> assertThat(sut.getKeyStore()).isEmpty();
+            default -> assertThatThrownBy(sut::getKeyStore).isInstanceOf(DestinationAccessException.class);
```
Johannes-Schneider, line 91: I'm not a fan of this assertion... 

How about spying the `DestinationKeyStoreExtractor` instance and verifying whether the `getKeyStore` method is being called.

Johannes-Schneider (approved):
LGTM

PR #259, opened 2024-01-26
refactor: CI/CD Pipelines

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#318

Continuation of #235 

This PR refactors our GitHub actions to, first and foremost, avoid the massive code duplication.
Additionally, it also refactors the release flow, which now provides a lot more automation and should be easier to use in general.

<details>
<summary>How CI Works</summary>

The CI portion of our pipelines has been changed in the following ways:

1. There is now one general purpose CI workflow (`continuous-integration.yaml`).
2. This workflow is triggered whenever a PR (targeting our `main` branch) receives an update.
3. **Additionally**, this workflow is also triggered from within other workflows (i.e. from our new `main-build.yaml`, our `prepare-release.yaml`, and our `deploy-snapshot.yaml`). This is how we are getting rid of the code duplication.

</details>

<details>
<summary>How CD Works</summary>

The CD portion (i.e. performing releases) has also been changed.
Here is the new flow:

1. The release facilitator triggers the `Prepare Release` (`prepare-release.yaml`) workflow. There are sensitive default parameters in place, so no manual input is required.
2. This workflow will do the following things:
    1. Bump the version of our module to the version to be released (e.g. `5.3.0`)
    2. Run the new `continuous-integration.yaml` workflow on the release commit. This includes code checks, such as Black Duck, CodeQL, etc.
    3. Create a new release tag (`rel/<VERSION>`) and also create a new draft release in our GitHub repo
    4. Create a PR for our updated release notes
    5. Create a PR for our updated java docs
    6. Create a PR for our updated code, which also contains further instructions for the release facilitator
    7. If anything along the way fails, everything will be cleaned up automatically by the workflow - no manual interaction needed.
3. If the `Prepare Release` workflow succeeds, the release facilitator will find a new PR with detailed instructions about what to do in our Code Repo (i.e. `SAP/cloud-sdk-java`). These TODOs include the following:
    1.  Review the release commit (there is a link in the PR body)
    2. Review the draft release (there is a link in the PR body)
    3. Review **and approve** the Release Notes PR (there is a link in the PR body)
    4. Review **and approve** the Java Docs PR (there is a link in the PR body)
    5. Review **and approve** the Code Updates (i.e. version bump and resetting the `release-notes.md`)
4. As soon as the code PR has been approved, the `Perform Release` workflow will be triggered automatically.
5. This workflow checks whether all PRs can be merged. If that is not the case, the workflow fails.
    1. In case of any failure, the `Perform Release` workflow can also be triggered manually now.
6. The Workflow will download the release artifacts and push them to Maven Central - this is still a **staging** repo, which needs to be released manually by the release facilitator.
7. The Workflow will also merge all of the PRs

</details>


.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,204 @@
...
+        required: true
+
+env:
+  MVN_CLI_ARGS: "--batch-mode --no-transfer-progress --fail-at-end --show-version"
+
+  DOCS_REPO: SAP/cloud-sdk
+
+jobs:
+  prerequisites:
+    name: "Prerequisites"
```
CharlesDuboisSAP, line 21: This job should only run `if: github.event.review.state == 'APPROVED'`. Otherwise it would run from a simple comment of change requested. Also this `.event.review.state` doesn't exist on a manual trigger, meaning it has to be fetched in the `Determine Branch Names` step.

.github/workflows/prepare-release.yaml
```diff
@@ -348,11 +346,15 @@ jobs:
 
       - name: "Create Code PR"
         run: |
+          COMMIT_URL=${{ github.event.repository.html_url }}/commit/${{ needs.bump-version.outputs.release-commit }}
+          
           PR_URL=$(gh pr create --title "Release ${{ needs.bump-version.outputs.release-version }}" --body "## TODOs
+          - [ ] Review the changes in [the release commit]($COMMIT_URL)
           - [ ] Review **and approve** the [Release Notes PR](${{ needs.create-release-notes-pr.outputs.pr-url }})
           - [ ] Review **and approve** the [JavaDocs PR](${{ needs.create-docs-pr.outputs.pr-url }})
```
CharlesDuboisSAP, line 354: Technically for some reason you don't need to approve those PRs for them to be merged.
I don't care about the checklist, just letting you know.

.github/workflows/prepare-release.yaml
```diff
@@ -348,11 +346,15 @@ jobs:
...
           - [ ] Review the [Draft Release](${{ needs.create-release.outputs.release-url }})
-          - [ ] Review **and approve** this PR")
+          - [ ] Review **and approve** this PR"
+          - [ ] Once the `Perform Release` action is through, head over to the [OSS Repository](https://oss.sonatype.org/#stagingRepositories) and log in with the credentials in the Team Password Safe. There should be a staged release. Check whether everything looks good and, if so, publish the release.)
```
CharlesDuboisSAP, line 357: We could try to use `mvn nexus-staging:release` after `mvn nexus-staging:deploy-staged-repository` in `Perform Release` to avoid this manual step.

CharlesDuboisSAP (dismissed):
- `docs/how-to/release.md` needs to be updated.
- Also the [release facilitator checklist template](https://github.com/SAP/cloud-sdk-java-backlog/blob/main/.github/ISSUE_TEMPLATE/release.md)

.github/workflows/perform-release.yml
Johannes-Schneider: > This job should only run `if: github.event.review.state == 'APPROVED'`.

Unfortunately, this check doesn't work if the workflow is triggered manually.

This is why the `prerequisites` job is checking whether all PRs are actually mergeable.
While this check will run on any kind of PR review submission, the rest shouldn't be executed as the Code PR will not be mergable if nobody accepted the changes.

.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,204 @@
...
+#          deploy
+#
+#          mvn
+#          $MVN_CLI_ARGS
+#          org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13:deploy-staged-repository
+#          -DserverId=ossrh
+#          -DnexusUrl=https://oss.sonatype.org
+#          -DrepositoryDirectory=./temp_local_repo
+#          -DstagingProfileId=$MAVEN_CENTRAL_PROFILE_ID
+#          -DautoReleaseAfterClose=true
```
Johannes-Schneider, line 152: TODO: this needs to be uncommented.

.github/workflows/prepare-release.yaml
Johannes-Schneider: > Technically for some reason you don't need to approve those PRs for them to be merged.

I see... I guess the `mergeable` field just indicates whether there are any conflicts with the base branch.
I'll rework that logic.

Johannes-Schneider:
@CharlesDuboisSAP wrote:
> `docs/how-to/release.md` needs to be updated.

Done.

---

> Also the [release facilitator checklist template](https://github.com/SAP/cloud-sdk-java-backlog/blob/main/.github/ISSUE_TEMPLATE/release.md)

This still looks pretty accurate to me. Especially since it is just linking to the (now updated) `how-to` document.
So no changes needed IMO.

.github/actions/await-workflow/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+      shell: bash
+
+    - name: View Run
+      if: ${{ inputs.commit-status != '' }}
+      id: view-run
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        JSON=$(gh run view ${{ inputs.run-id }} --json url,headSha)
+        echo "URL=$(echo $JSON | jq -r '.url')" >> $GITHUB_OUTPUT
```
newtork, line 39: (Minor/Nitpick)

Weird `echo "... $(echo ...)"` nesting, but I don't have an immediate suggestion.

.github/actions/await-workflow/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+          echo "RUN_SUCCEEDED=false" >> $GITHUB_OUTPUT
+          exit 1
+        fi
+        
+        echo "RUN_SUCCEEDED=true" >> $GITHUB_OUTPUT
+      shell: bash
+
+    - name: Determine Final Commit Status
+      id: determine-final-commit-status
+      if: ${{ always() && inputs.commit-status != '' }}
```
newtork, line 79: (Question)

What does `alway()` do here?

.github/actions/run-workflow/action.yaml
```diff
@@ -0,0 +1,93 @@
...
+        echo "[DEBUG] Starting 'Run Workflow' Action; inputs = ${{ toJson(inputs) }}"
+      shell: bash
+
+    - name: Trigger Workflow
+      id: trigger-workflow
+      run: |
+        PREVIOUS_RUN_ID=$(gh run list --workflow=${{ inputs.workflow}} --commit=${{ inputs.commit-sha }} --json databaseId | jq -r '.[0].databaseId')
+        echo "[DEBUG] Previous run id = '$PREVIOUS_RUN_ID'"
+        
+        gh workflow run "${{ inputs.workflow }}" --ref "${{ inputs.workflow-ref }}" ${{ inputs.parameters }}
```
newtork, line 44: (Question/OT)

Does github actions use prepared statements? Such that we can be sure no-one can inject bash commands using these input arguments? :) Little Bobby Tables

.github/actions/run-workflow/action.yaml
```diff
@@ -0,0 +1,93 @@
...
+            
+            RUN_URL=$(gh run view $LATEST_RUN_ID --json url | jq -r '.url')
+            echo "[DEBUG] ${{ inputs.workflow }} run #$LATEST_RUN_ID successfully triggered: $RUN_URL"
+            echo "[${{ inputs.workflow }} run (#$LATEST_RUN_ID)]($RUN_URL)" >> $GITHUB_STEP_SUMMARY        
+            echo "RUN_ID=$LATEST_RUN_ID" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+        done
+        
+        echo "[DEBUG] Unable to detect new run of workflow '${{ inputs.workflow }}'."
```
newtork, line 63: (Minor)

maybe adding ... "within 60 seconds".

.github/actions/run-workflow/action.yaml
```diff
@@ -0,0 +1,93 @@
...
+      run: |
+        echo "[DEBUG] Waiting for run '${{ steps.trigger-workflow.outputs.RUN_ID }}' to complete..."
+        
+        # we tried using 'gh run watch' instead of doing the polling ourselves, but that failed with an HTTP 500 after a while
+        while :
+        do
+          CONCLUSION=$(gh run view ${{ steps.trigger-workflow.outputs.RUN_ID }} --json conclusion | jq -r '.conclusion')
+        
+          if [[ -z "$CONCLUSION" ]]; then
+            sleep 5
```
newtork, line 78: (Question)

Potential endless loop, right? Or is `.conclusion` guaranteed to be available, even in error case?

.github/actions/scan-with-blackduck/action.yaml
```diff
@@ -0,0 +1,37 @@
...
+      run: echo "MAJOR_VERSION=$(cat latest.json | jq -r .version | cut -d '.' -f 1)" >> $GITHUB_OUTPUT
+      shell: bash
+
+    - name: Determine Maven Excludes
+      id: get-maven-excludes-for-blackduck
+      run: python .pipeline/scripts/get-maven-excludes.py --filter-key excludeFromBlackDuckScan --filter-value True
+      shell: bash
+
+    - name: BlackDuck Scan
+      uses: SAP/project-piper-action@master
```
newtork, line 22: (Question)

There is no version management available? :/

.github/actions/run-workflow/action.yaml
```diff
@@ -0,0 +1,93 @@
+name: "Run Workflow"
```
newtork, line 1: (Comment/Question)

I find "Run" and "Trigger" workflow kinda ambiguous. Maybe I'm unaware of a naming convention though.

Did you consider recycling "trigger workflow" inside "run workflow"?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+      - name: "Setup Java"
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: 17
+
+      - name: "Run Maven Plugins"
+        run: |
+          MVN_GOALS="\
+          net.revelc.code.formatter:formatter-maven-plugin:validate \
```
newtork, line 93: (Comment)

Slightly confused as to why FQN plugin goals are needed(?) here. Not sure, but I can imagine this will pull the latest version from Maven Central opposed to the declared version in `pom.xml`.

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+          mvn $MVN_ARGS
+          
+          if [[ "${{ github.event.inputs.build-release-artifacts }}" == "true" ]]; then
+            VERSION=$(cat latest.json | jq -r .version)
+            mkdir -p $HOME/.sdk-release
+          
+            PYTHON_ARGS="--version $VERSION --path-prefix $HOME/.sdk-release/${{ env.RELEASE_ARTIFACT_NAME }}"
+            if [[ "${{ github.event.inputs.sign-release-artifacts }}" == "true" ]]; then
+              echo "[DEBUG] Signing release artifacts"
+              PYTHON_ARGS="$PYTHON_ARGS --with-signing"
```
newtork, line 143: (Question)

Why have the signing toggle? Is it expensive or long-running? Or are there side effects?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+      - name: "Cache Dependencies"
+        if: ${{ steps.restore-dependencies.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@v3
+        with:
+          key: ${{ env.DEPENDENCIES_CACHE_KEY }}
+          path: ${{ env.DEPENDENCIES_PATH }}
+
+      - name: "Build Module Inventory"
+        run: >
+          python ./scripts/create_module_inventory_file.py
```
newtork, line 160: (Question)

Does the `>` mean the whole run text block will be a single line / opposed to `|`?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+              exit 1
+          fi
+
+      - name: "Upload Release Artifacts"
+        if: ${{ github.event.inputs.build-release-artifacts == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.RELEASE_ARTIFACT_NAME }}
+          path: ~/.sdk-release/${{ env.RELEASE_ARTIFACT_NAME }}
+          retention-days: 1
```
newtork, line 182: (Question)

What's the implication for us having retention days = 1 for the release artifacts?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+        if: github.event.pull_request.head.repo.full_name == 'SAP/cloud-sdk-java'
+        uses: scacap/action-surefire-report@v1
+      - name: Coverage Report
+        run: python .pipeline/scripts/print-coverage.py --jacoco-report-pattern "**/target/site/jacoco/jacoco.csv"
+
+  static-code-analysis:
+    needs: [ context, build ]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
newtork, line 243: (Question)

I understand matrix is being used for deduplication. But how costly is the repeated checkout+setup+restore steps?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+
+      - name: "Test ${{ matrix.task.name }}"
+        working-directory: ./example-${{ matrix.task.archetype }}
+        run: mvn test -B -Dsurefire.logLevel='error'
+
+      - name: "Spotbugs ${{ matrix.task.archetype }}"
+        working-directory: ./example-${{ matrix.task.archetype }}
+        run: >
+          mvn com.github.spotbugs:spotbugs-maven-plugin:check -B
+          -pl !integration-tests
```
newtork, line 392: (Comment)

This module exclusion will soon no longer be necessary.

.github/workflows/deploy-snapshot.yaml
```diff
@@ -0,0 +1,105 @@
+name: Deploy Snapshot to Maven Central
+
+on:
+  workflow_dispatch:
```
newtork, line 4: (Question)

What does this trigger mean..? `on: workflow_dispatch`

.github/workflows/deploy-snapshot.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+      - name: "Get Commit SHA"
+        run: echo "COMMIT_SHA=$(git rev-parse HEAD)" >> $GITHUB_ENV
+      - name: "Get Branch Name"
+        run: echo "BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
+
+      - name: "Trigger CI Workflow"
+        id: trigger-ci
+        uses: ./.github/actions/trigger-workflow
```
newtork, line 37: (Question)

Why not `run-workflow` if it's waiting anyway?

.github/workflows/deploy-snapshot.yaml
```diff
@@ -0,0 +1,105 @@
...
+        run: >
+          mvn
+          $MVN_CLI_ARGS
+          -Durl=https://oss.sonatype.org/content/repositories/snapshots
+          -DrepositoryId=artifactory-snapshots
+          -Dmaven.install.skip=true
+          -Dmaven.test.skip
+          -Dmaven.compiler.showCompilationChanges
+          -Dhttp.keepAlive=false
+          deploy
```
newtork, line 87: (Question)

Does `mvn deploy` use the `target/` "release artifacts" that were previously "downloaded"? I assume they were stored in current-working-directory, this is why the maven command succeeds, right?

.github/actions/await-workflow/action.yaml
Johannes-Schneider: `always()` is a shortcut for `failure() || success()`.
So it will ALWAYS run this step, regardless of whether anything prior to that failed.

.github/actions/run-workflow/action.yaml
Johannes-Schneider: > Does github actions use prepared statements?

I don't think so.
AFAIK, GitHub will just bluntly replace all of these `${{ }}` variables.
So there might be the chance of command injection here.

.github/actions/run-workflow/action.yaml
Johannes-Schneider: Didn't want to add that so we don't need to update it, should we decide to extend/shorten the timeout.
No strong opinion though.

.github/actions/run-workflow/action.yaml
```diff
@@ -0,0 +1,93 @@
...
+        
+          if [[ -z "$LATEST_RUN_ID" || "$LATEST_RUN_ID" == "$PREVIOUS_RUN_ID" ]]; then
+            echo "[DEBUG] No new run detected. Waiting for 5 seconds."
+            sleep 5
+          else
+            echo "[DEBUG] New workflow run detected: '$LATEST_RUN_ID'."
+            
+            RUN_URL=$(gh run view $LATEST_RUN_ID --json url | jq -r '.url')
+            echo "[DEBUG] ${{ inputs.workflow }} run #$LATEST_RUN_ID successfully triggered: $RUN_URL"
+            echo "[${{ inputs.workflow }} run (#$LATEST_RUN_ID)]($RUN_URL)" >> $GITHUB_STEP_SUMMARY
```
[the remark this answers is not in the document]
Johannes-Schneider: Yep, that works

image

.github/actions/run-workflow/action.yaml
Johannes-Schneider: `.conclusion` should be available as soon as the workflow has stopped running (i.e. regardless of whether the run succeeded, failed, or was skipped).

.github/actions/scan-with-blackduck/action.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Get Major Version
+      id: get-major-version
+      run: echo "MAJOR_VERSION=$(cat latest.json | jq -r .version | cut -d '.' -f 1)" >> $GITHUB_OUTPUT
+      shell: bash
+
+    - name: Determine Maven Excludes
+      id: get-maven-excludes-for-blackduck
+      run: python .pipeline/scripts/get-maven-excludes.py --filter-key excludeFromBlackDuckScan --filter-value True
+      shell: bash
```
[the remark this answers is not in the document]
Johannes-Schneider: Yep - too common of a use case I guess.

.github/actions/scan-with-blackduck/action.yaml
Johannes-Schneider: Nope

image

.github/actions/run-workflow/action.yaml
Johannes-Schneider: Good that you point that out. I think this action is actually a left over.

The only two that are still in use are `trigger-workflow` and `await-workflow`.

I'll check again, but I'm pretty sure this action can be removed.

.github/workflows/continuous-integration.yaml
Johannes-Schneider: I'd be surprised if it would pull any version different from the one defined in our `pom` because we are NOT providing a version here, right?

(If we'd use `mvn net.revelc.code.formatter:formatter-maven-plugin:A.B.C:validate` THEN it would use the specified version instead of the one defined in our `pom.xml`.)

.github/workflows/continuous-integration.yaml
Johannes-Schneider: This question is a bit complex... so bear with me.

We do have several requirements for our release (and CI) process:

1. We want to avoid as much duplicated work within our pipelines as possible (i.e. building the SDK multiple times is a waste of time)
2. Releasing any version (i.e. patch version or a new SNAPSHOT) should be an **atomic** operation. In other words: We are either releasing ALL modules or none so that we never end up in an inconsistent state.
3. Our "production" releases (i.e. the ones to Maven Central) require artifact signing. Our snapshot releases (to our internal artifactory), however, do NOT require signing (there is a login required, so I guess it is fair to assume that all the artifacts are legitimate).

With these requirements in mind, our release process looks as follows:

1. The CI workflow builds the SDK.
2. The CI workflow then runs the `generate-release-artifacts.py` script.
3. This script creates an artificial multi-module Maven project that looks likes this:

<details>
<summary>Artificial Maven Project</summary>

```xml
<project>
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.sap.cloud.sdk</groupId>
    <artifactId>sdkdist</artifactId>
    <version>5.3.0</version>
    <properties>
        <maven.deploy.skip>true</maven.deploy.skip>
        <maven.install.skip>true</maven.install.skip>
    </properties>
    <build>
        <plugins>

            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-install-plugin</artifactId>
                <version>3.1.1</version>
                <executions>

                  <execution>
                      <id>install-archetypes-parent</id>
                      <phase>install</phase>
                      <goals>
                          <goal>install-file</goal>
                      </goals>
                      <configuration>
                          <file>artifacts/archetypes/pom.xml</file>
                          <sources></sources>
                          <javadoc></javadoc>
                          <pomFile>artifacts/archetypes/pom.xml</pomFile>
                          <groupId>com.sap.cloud.sdk.archetypes</groupId>
                          <artifactId>archetypes-parent</artifactId>
                          <version>5.3.0</version>
                          <packaging>pom</packaging>
                      </configuration>
                  </execution>

...

                </executions>
            </plugin>

            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-gpg-plugin</artifactId>
                <version>3.1.0</version>
                <executions>

                  <execution>
                      <id>deploy-archetypes-parent</id>
                      <phase>deploy</phase>
                      <goals>
                          <goal>sign-and-deploy-file</goal>
                      </goals>
                      <configuration>
                          <file>artifacts/archetypes/pom.xml</file>
                          <sources></sources>
                          <javadoc></javadoc>
                          <pomFile>artifacts/archetypes/pom.xml</pomFile>
                          <groupId>com.sap.cloud.sdk.archetypes</groupId>
                          <artifactId>archetypes-parent</artifactId>
                          <version>5.3.0</version>
                          <packaging>pom</packaging>
                      </configuration>
                  </execution>
```
</details>

This artificial project is our current approach of making releases atomic:
We are calling the staging plugin on this artificial project. The plugin then creates a staging repository and pushes everything into it.
If (at any point) there is any failure, the plugin takes care of automatically closing the staging repo and nothing bad happens.

<details>
<summary>The Alternative</summary>

As an alternative to this approach, we considered calling the staging plugin for each of our previously created `.jar` files individually.

Personally, I think we could get this approach to work **iff** we somehow are able to create a staging repository manually and push everything into it afterwards.
However, we didn't spend the time to investigate this approach in the past.
And since we were concerned about creating inconsistent releases, we decided for this approach.

</details>

(BTW: This general process has not been touched in this PR and exists already since V3 (or maybe even longer)).

This project contains one `execution` of the `maven-install-plugin` for each module that should be released.
As you can see, we are explicitly installing an existing `.jar` file.
This is to adhere to requirement **1** - instead of re-installing the module, we are using the previously generated files.

Furthermore, if artifact signing is required, there will also be an `execution` of the `maven-gpg-plugin` for each of our released modules.

**This is why it makes a difference**: This plugin is required ONLY for our production releases (which should be signed). For SNAPSHOT deployments, this must not be included.

.github/workflows/continuous-integration.yaml
Johannes-Schneider: Yes.

.github/workflows/continuous-integration.yaml
Johannes-Schneider: The retention days refers to the artifact that is being attached to the GitHub workflow run.

When we are performing a production release, our `Prepare Release` pipeline will call the `Continuous Integration` workflow on the release commit.
This workflow run will have the release artifacts attached to it.

Then (immediately afterwards, still in the `Prepare Release` workflow), those artifacts will be downloaded from the Workflow run and attached to a new GitHub Draft Release - those are permanent.

.github/workflows/continuous-integration.yaml
Johannes-Schneider: It seems to be very cheap:

image


(The above runtime includes all of the "overhead" + the plugin and python script execution)

.github/workflows/deploy-snapshot.yaml
Johannes-Schneider: Yes.

.github/workflows/deploy-snapshot.yaml
Johannes-Schneider: As I outlined above: The "Run Workflow" action is a left-over that will be removed.

.github/workflows/deploy-snapshot.yaml
Johannes-Schneider: Refer to [this comment](https://github.com/SAP/cloud-sdk-java/pull/259#discussion_r1472450294) for the details.

.github/actions/await-workflow/action.yaml
newtork: but why would you put this as part of a conjunction in the if: statement?

.github/workflows/continuous-integration.yaml
newtork: > AFAIK, when calling individual plugins, we have to use the FQN.

I think it wasn't necessary in the past when I ran `mvn formatter:format` without issues. But not anymore.

> we are NOT providing a version explicitly

I checked. If the dependency is declared with version, Maven will us it.
If undeclared, Maven will download the latest version to execute it.

However it doesn't really affect your PR

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,458 @@
...
+
+      - name: "Initialize CodeQL"
+        uses: github/codeql-action/init@v2
+        with:
+          languages: "java"
+          queries: security-extended
+
+      - name: "Build SDK"
+        run: |
+          MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} install -Dmaven.test.skip=true ${{ env.MVN_SKIP_CI_PLUGINS }}"
```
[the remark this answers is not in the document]
newtork: Maybe for a better distinction: would it also work if we used `compile` instead of `install`?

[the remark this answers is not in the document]
newtork: > CodeQL (somehow?) attaches itself to the compiler/build tool

I wonder how. I assume the following two lines do not reference any explicit CodeQL, when resolved, right?

```
MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} install -Dmaven.test.skip=true ${{ env.MVN_SKIP_CI_PLUGINS }}"
mvn $MVN_ARGS
```

.github/workflows/deploy-snapshot.yaml
```diff
@@ -0,0 +1,105 @@
...
+    name: "Continuous Integration"
+    runs-on: ubuntu-latest
+    outputs:
+      ci-run-id: ${{ steps.trigger-ci.outputs.run-id }}
+    permissions:
+      actions: write # needed to trigger the ci-build workflow
+      statuses: write # needed to update the commit status
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
```
KavithaSiva, line 29: As this workflow is always supposed to be run manually, what branch would be checked out?
We don't have branch as one of the inputs provided.

.github/workflows/main-build.yaml
```diff
@@ -0,0 +1,92 @@
+name: "Main Build"
```
KavithaSiva, line 1: This workflow is almost similar to the deploy-snapshot workflow, can't we just provide an option to run this workflow manually and remove the deploy-snapshot workflow or am I missing something?

KavithaSiva (commented):
More comments following ...

.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,267 @@
...
+          
+          echo "RELEASE_VERSION=$RELEASE_VERSION" >> $GITHUB_OUTPUT
+          echo "RELEASE_TAG=rel/$RELEASE_VERSION" >> $GITHUB_OUTPUT
+          echo "CODE_BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_OUTPUT
+          echo "DOCS_BRANCH_NAME=$DOCS_BRANCH" >> $GITHUB_OUTPUT
+          echo "RELEASE_NOTES_BRANCH_NAME=$RELEASE_NOTES_BRANCH" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: "Check Whether Code PR Can Be Merged"
```
KavithaSiva, line 58: All the steps which check if a particular PR can be merged seem to have the same code except for the inputs supplied for e.g branch code name, repo, token etc.

Was there a reason why you did not consider moving it to a reusable action? Or would having a matrix here help?

.github/actions/await-workflow/action.yaml
Johannes-Schneider: Because, otherwise, this step wouldn't be executed if the previous step fails - in other words: If the triggered workflow fails, we wouldn't update the commit status, leaving it as "operation pending", which wouldn't be correct.

.github/actions/run-workflow/action.yaml
Johannes-Schneider: The workflow has been removed.

.github/workflows/continuous-integration.yaml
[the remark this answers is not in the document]
Johannes-Schneider: > Maybe for a better distinction: would it also work if we used compile instead of install?

Good question. Didn't test it honestly.

---

> I wonder how. I assume the following two lines do not reference any explicit CodeQL, when resolved, right?

No, they have nothing to do with CodeQL.
I assume the previous step (i.e. `Setup CodeQL`) does some magic.

.pipeline/scripts/release_notes_automation.py
```diff
@@ -0,0 +1,132 @@
+import argparse
```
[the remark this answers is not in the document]
Johannes-Schneider: > Also, currently, the release notes in our repo only contains the current version's notes, should we change it to also contain all the previous release notes too?

You mean so that anyone who just looks at our code repo gets the full release history?
Or why would you do that?

.github/workflows/deploy-snapshot.yaml
Johannes-Schneider: Good question (I didn't change anything about how the general workflow works, so I'm not super deep into the details).

I assume that the checkout action is smart enough to use the branch the workflow is taken from if nothing else is provided. 
(And you always have to provide the branch of the workflow file)

.github/workflows/main-build.yaml
Johannes-Schneider: The `deploy-snapshot` is deploying to the **Maven Central** SNAPSHOT repo while our `main-build` deploys to the **Internal Artifactory**.

Not sure how much of a difference there is besides the target repository - but we could consider consolidating these two flows.

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,457 @@
...
+          name: ${{ env.RELEASE_ARTIFACT_NAME }}
+          path: ~/.sdk-release/${{ env.RELEASE_ARTIFACT_NAME }}
+          retention-days: 1
+      - name: "Upload SDK M2"
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.SDK_M2_NAME }}
+          path: ${{ env.SDK_M2_PATH }}
+          retention-days: 1
+      - name: "Upload SDK Targets"
```
KavithaSiva, line 189: What are Release Artifacts composed of as opposed to SDK Targets?

.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,267 @@
...
+
+      - name: "Check Whether Docs PR Can Be Merged"
+        run: |
+            PR=$(gh pr view "${{ steps.determine-branch-names.outputs.DOCS_BRANCH_NAME }}" --repo ${{ env.DOCS_REPO }} --json number,reviewDecision,mergeable,statusCheckRollup)
+            NUMBER=$(echo "$PR" | jq -r '.number')
+            DECISION=$(echo "$PR" | jq -r '.reviewDecision')
+            MERGEABLE=$(echo "$PR" | jq -r '.mergeable')
+            
+            if [[ "$MERGEABLE" != "MERGEABLE" ]]; then
+                echo "Docs PR (#$NUMBER) is not mergeable."
```
newtork, line 102: (Minor/Question)

Have you considered dumbing down the PR handling? There is duplicate code and the exact reason why a "branch" is failing is not that important. We'll see anyway when looking up the culprit.

I was wondering whether by dumbing it down to only one error state (`exit 1`), maybe we can save significant LoC.

KavithaSiva (commented):
Really liked the refactoring done, is pretty impressive :)

.github/workflows/perform-release.yml
Johannes-Schneider: Good point. We could just check whether someone approved the PR. 

If that is the case, I think it's fair to assume that everything is also fine.
Will do that 👍

.github/actions/scan-with-blackduck/action.yaml
```diff
@@ -0,0 +1,37 @@
+name: "Scan with BlackDuck"
```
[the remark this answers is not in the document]
Johannes-Schneider: This re-usable action is basically just "inlined" into the workflow that calls it.
So everything that is logged/printed in here will also be visible in the workflow that uses this action.

.github/workflows/perform-release.yml
Johannes-Schneider: [Alex pointed that out as well](https://github.com/SAP/cloud-sdk-java/pull/259#discussion_r1473012174).

I reduced the code by quite a bit. See [cfb86c6](https://github.com/SAP/cloud-sdk-java/pull/259/commits/cfb86c6908fdd6b8f7dcbf4681a5bdf1b890b5c2)

.github/workflows/continuous-integration.yaml
Johannes-Schneider: The SDK Targets contains ALL modules of the SDK (also those that are not meant to be released. Like out test util, for example).

The Release artifacts, however, contain an artificial maven project that contains only modules that are to be released.
Check out [this comment](https://github.com/SAP/cloud-sdk-java/pull/259#discussion_r1472450294) for in-depth info.

.github/workflows/main-build.yaml
KavithaSiva: Can be done in follow up PR., doesn't necessarily have to be in this one.

.github/workflows/deploy-snapshot.yaml
KavithaSiva: Looks like `actions/checkout@v4` would checkout the default branch in the repo.

.github/workflows/continuous-integration.yaml
KavithaSiva: test comment.

.github/workflows/perform-release.yml
```diff
@@ -19,6 +18,7 @@ env:
 jobs:
   prerequisites:
     name: "Prerequisites"
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request_review' && github.event.review.state == 'approved' && startsWith(github.event.pull_request.head.ref, 'RELEASE-')) }}
```
Johannes-Schneider, line 21: **Discussion**

We are still having a `pull_request_review` trigger here. So even though the job would be skipped for any PR that is not coming from a `RELEASE-*` branch, it would still appear within the status checks of every other PR.

So maybe we should get rid of the automatic trigger altogether?

.github/workflows/main-build.yaml
```diff
@@ -0,0 +1,92 @@
...
+          server-password: ${{ secrets.ARTIFACTORY_TOKEN }}
+
+      - name: "Download Release Artifacts"
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ env.RELEASE_ARTIFACT_NAME }}
+          github-token: ${{ github.token }}
+          run-id: ${{ needs.run-ci.outputs.ci-run-id }}
+
+      - name: "Publish Snapshot"
```
MatKuhr, line 68: Isn't this a duplicate of the `deploy-snapshot` workflow? 

I see that `deploy-snapshot` also itself triggers CI again. I figured that is probably for cases where we want to trigger a snapshot deployment manually e.g. from a branch? Maybe it is worth adding a toggle for that somehow so that we can re-use the workflow.

.github/workflows/perform-release.yml
MatKuhr: would agree, one extra button press isn't the world, better not to have our PR checks (even more) cluttered.. no strong opinion though ofc...

.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,183 @@
+name: "Perform Release"
+
+on:
+  pull_request_review:
+    types: [submitted]
+
+  workflow_dispatch:
+    inputs:
```
MatKuhr, line 8: we could consider offering a toggle to skip waiting for mergable (doc) PRs. E.g. if our docs vale action is broken again we may want to not be hard blocked from releasing by that..

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,411 @@
...
+          ref: ${{ github.event.inputs.branch }}
+
+      - name: "Delete Release"
+        if: ${{ needs.create-release.outputs.release-url != '' }}
+        run: gh release delete --repo "${{ github.repository }}" ${{ needs.create-release.outputs.release-name }} --yes
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        continue-on-error: true
+
+      - name: "Delete Release Branch"
```
MatKuhr, line 381: Not sure if deleting the branch before closing the PR is an issue, maybe this gets rejected. Did a quick google but didn't find a definitive answer quickly..

MatKuhr: But with `continue-on-error: true` this shouldn't be a big deal anyway.

MatKuhr (commented):
mostly LGTM, one minor question on the snapshot deployment

didn't look at the release notes python yet

.github/workflows/perform-release.yml
Johannes-Schneider: I removed the automatic trigger and added another checkbox to the release PR body (i.e. _Trigger the Perform Release Workflow_).

.github/workflows/perform-release.yml
Johannes-Schneider: Good idea - I had this in the back of my head but forgot to actually do it.
Now, there is such a toggle.

.github/workflows/prepare-release.yaml
Johannes-Schneider: As the creation of the code PR is the very last step in this workflow, we should never run into a situation where there is a code branch and a code PR (as this would mean that the workflow overall succeeded).
This is why there is no stage to "Delete the Code PR".

.github/workflows/main-build.yaml
Johannes-Schneider: I refactored both workflows to now rely on a re-usable action instead of having the code duplicated.

Currently, we are still lacking parameters for our `Deploy Snapshot` workflow, but maybe that could be improved in an other PR (as the `Deploy Snapshot` workflow is rarely/never used as of now).

.github/actions/deploy-snapshot/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+    description: "The run ID of the 'Continuous Integration' workflow that produced the release artifacts."
+    required: true
+  repository-url:
+    description: "The URL of the repository to which the release artifacts should be deployed."
+    required: true
+  repository-username:
+    description: "The username to use when authenticating with the repository."
+    required: true
+  repository-password:
+    description: "The password to use when authenticating with the repository."
```
MatKuhr, line 15: Not too sure if this is safe to do. Maybe input parameters are logged in plaintext somewhere? Maybe just pass the name of the secret instead?

Johannes-Schneider: Generally, it seems to be a "best practice" to pass around secrets like this
(see the `token` parameter of the [checkout action](https://github.com/actions/checkout?tab=readme-ov-file#usage) for example).

But I'd still agree: As this is our own action, we don't need to pass around the actual secrets. Using the secret names should be sufficient 👍

Johannes-Schneider: Unfortunately, passing the secret names (instead of their actual values), doesn't work as it would require some sort of "nested GitHub macro extension" like so:

```yaml
- run: something
  env:
    GH_TOKEN: ${{ secrets.${{ inputs.secret-name}} }}
```

This doesn't seem to be supported by GH actions.

PR #260, opened 2024-01-26
Remove integration test module from archetype

cschubertcs:
## Context

SAP/cloud-sdk-java-backlog#363.

This PR removes the `integration-tests` module from our archetype and moves all contained files over into the `application` module. This also includes test dependencies needed for the integration test.

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


cschubertcs:
It's maybe debatable whether we want to keep the `application` module or not.

I think I would vote for keeping it, as customers might want to add other modules themselves.
If the archetype is then already prepared to handle multiple modules it makes their life easier.

Besides general overview I don't see any benefit from removing the `application` module.

MatKuhr:
> Besides general overview I don't see any benefit from removing the application module.

No strong opinion, but argument might be that for people who are not already experienced with maven the multi-module setup is somewhat more complicated.

However, if we keep the application module we should apply this task from the BLI:
> Move the [Spring Boot config](https://github.tools.sap/cloudsdk/cap-e2e-app/blob/906da32950414030cdb7276b6f5b9c7e6d705986/pom.xml#L82-L90) into the parent pom

as otherwise you can't run the app from the root dir, which is inconvenient

cschubertcs:
> However, if we keep the application module we should apply this task from the BLI:
> 
> > Move the [Spring Boot config](https://github.tools.sap/cloudsdk/cap-e2e-app/blob/906da32950414030cdb7276b6f5b9c7e6d705986/pom.xml#L82-L90) into the parent pom
> 
> as otherwise you can't run the app from the root dir, which is inconvenient

Yep, looking right now into it.

cschubertcs:
This PR can be merged, I don't see a way to make both, the subdirectory and the parent directory, running the `mvn spring-boot:run` without any parameter.

Using the `-f application` option it will work from the parent directory while keeping the parameter less command working from the subdirectory.

newtork (approved):
Looks good to me.

However several adjustments will be necessary in our documentation portal and the tutorials :)

PR #265, opened 2024-01-30
feat: Automatically Create Release Notes for Dependency Updates

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#196.

This PR introduces a new pipeline script (`dependabot-release-notes.py`) which is capable of transforming a PR body (like #255, for example) into our release notes format.

<details>
<summary>Example Release Notes</summary>

### 📈 Improvements

- Some Improvement.

<details>
<summary>Dependency Updates</summary>

| Dependency | From | To |
| --- | --- | --- |
| [some-artifact](https://search.maven.org/search?q=g:some.group%20AND%20a:some-artifact) (`some.group`) | `1.2.3` | `2.3.4` |

</details>

</details>

Along the script, there is also a matching new github workflow (`dependabot-release-notes.yaml`).
This new workflow is triggered in the following ways:

1. Whenever a new PR (pointing to `main`) is opened[^1].
2. Whenever a PR (pointing to `main`) is updated[^1].
3. Manually.

[^1]: Even though the workflow will be triggered for every PR (independent from the author), it performs some checks up-front to avoid effort in case the PR is not updating dependencies.

### Feature scope:

- [x] Implement PR Body → Release Note Script
  - [x] The script will automatically update our `release_notes.md` and can handle the following scenarios:
    - There are NO existing dependency updates yet (i.e. first PR after our release)
    - There are ALREADY existing dependency updates (i.e. subsequent PR after our release)
    - The release notes are lacking the `### 📈 Improvements` section (which should never really be the case)
- [x] Implement a Workflow to run the new script  

## Definition of Done

- [x] Tests cover the scope above[^2]
- [x] ~Documentation updated~
- [x] ~Release notes updated~

[^2]: There are some unit tests that cover the functionality of our new Python script. However, "end-to-end" testing is not really possible (AFAIK).

### DIFF
--- a/.github/workflows/dependabot-release-notes.yaml
+++ b/.github/workflows/dependabot-release-notes.yaml
@@ -1,0 +1,101 @@
+name: "Dependabot Release Notes"
+
+on:
+  pull_request:
+    types: [ opened, synchronize ]
+    branches: [ main ]
+
+  workflow_dispatch:
+    inputs:
+      pr-number:
+        description: "Pull Request Number"
+        required: true
+
+jobs:
+  context:
+    name: "Collect Context"
+    outputs:
+      pr-number: ${{ steps.determine-pr-info.outputs.PR_NUMBER }}
+      pr-branch: ${{ steps.determine-pr-info.outputs.PR_BRANCH }}
+      pr-body: ${{ steps.determine-pr-info.outputs.PR_BODY }}
+    permissions:
+      pull-requests: read
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Print Context"
+        run: echo "${{ toJson(github) }}"
+
+      - name: "Determine PR Info"
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            PR_NUMBER=${{ github.event.inputs.pr-number }}
+            PR_JSON=$(gh pr view $PR_NUMBER --json headRefName,body)
+            PR_BRANCH=$(echo $PR_JSON | jq -r '.headRefName')
+            PR_BODY=$(echo $PR_JSON | jq -r '.body')
+          
+            echo "[DEBUG] Using PR number (#$PR_NUMBER from $PR_BRANCH) provided by workflow_dispatch event"
+            echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_OUTPUT
+            echo "PR_BRANCH=$PR_BRANCH" >> GITHUB_OUTPUT
+            echo "PR_BODY=$PR_BODY" >> GITHUB_OUTPUT
+            exit 0
+          fi
+          
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            if [[ "${{ github.event.action }}" == "synchronize" ]]; then
+              EVENT_SENDER=$(echo ${{ github.event.sender }} | jq -r 'select(.is_bot == true and .login == "app/github-actions"')
+              if [[ -n "$EVENT_SENDER" ]]; then
+                echo "[DEBUG] The current workflow has been triggered by GitHub actions itself - most probably by this workflow."
+                echo "[DEBUG] Therefore, we can abort this new workflow run."
+                exit 0
+              fi
+            fi
+          
+            PR_NUMBER=${{ github.event.pull_request.number }}
+            PR_BRANCH=${{ github.event.head_ref }}
+            PR_JSON=$(gh pr view $PR_NUMBER --json author,body)
+            PR_AUTHOR=$(echo $PR_JSON | jq -r '.author | select(.is_bot == true and .login == "app/dependabot"')
+            PR_BODY=$(echo $PR_JSON | jq -r '.body')
+          
+            if [[ -n "$PR_AUTHOR" ]]; then
+              echo "[DEBUG] PR #$PR_NUMBER (from $PR_BRANCH) has been created by Dependabot. Continuing the workflow."
+              echo "PR_NUMBER=$PR_NUMBER" >> GITHUB_OUTPUT
+              echo "PR_BRANCH=$PR_BRANCH" >> GITHUB_OUTPUT
+              echo "PR_BODY=$PR_BODY" >> GITHUB_OUTPUT
+            else
+              echo "[DEBUG] PR #$PR_NUMBER (from $PR_BRANCH) has not been created by Dependabot. Aborting workflow."
+              # we are "aborting" the workflow by NOT setting the `PR_NUMBER` output
+            fi
+          fi
+        id: determine-pr-info
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+  update-release-notes:
+    name: "Update Release Notes"
+    needs: [ context ]
+    if: ${{ needs.context.outputs.pr-number != '' }}
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout Repository"
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.context.outputs.pr-branch }}
+
+      - name: "Run Python Script"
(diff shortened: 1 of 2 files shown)


.github/workflows/dependabot-release-notes.yaml
```diff
@@ -0,0 +1,99 @@
...
+          
+            jq -r '.body' <<< "$PR_JSON" > $PR_BODY_FILE
+          
+            if [[ -n "$PR_AUTHOR" ]]; then
+              echo "[DEBUG] PR #$PR_NUMBER (from $PR_BRANCH) has been created by Dependabot. Continuing the workflow."
+              echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+              echo "PR_BRANCH=$PR_BRANCH" >> $GITHUB_ENV
+            else
+              echo "[DEBUG] PR #$PR_NUMBER (from $PR_BRANCH) has not been created by Dependabot. Aborting workflow."
+              # we are "aborting" the workflow by NOT setting the `PR_NUMBER` output
```
newtork, line 54: (Question)

There is no more convenient / proper way to terminate a workflow?

.github/workflows/dependabot-release-notes.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: "Checkout Repository"
+        if: ${{ env.PR_NUMBER != '' }}
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.PR_BRANCH }}
+
+      - name: "Run Python Script"
+        if: ${{ env.PR_NUMBER != '' }}
+        run: |
+          python ./.pipeline/scripts/dependabot-release-notes.py --pr-body-file "${{ env.PR_BODY_FILE }}"
```
newtork, line 70: (Minor/Nitpick)

I would assume the 500+ loc python script can be replaced by a few lines combined awk / sed. 🔥

.pipeline/scripts/dependabot-release-notes.py
```diff
@@ -0,0 +1,536 @@
...
+        if raw.endswith("`"):
+            raw = raw[:-1]
+
+        return raw
+
+
+# Expected format of the Release Notes:
+# ### 📈 Improvements
+# ... (notes)
+# <details><summary>Dependency Updates</summary>
```
newtork, line 116: (Minor)

Not a fan of changing release notes format in the same PR where the automation script becomes introduced.

newtork (commented):
Other than my question about properly "stopping" github actions in error cases, i'm fine with the changes.

However the new format needs to be approved by PO or confirmed by team :)

.github/workflows/dependabot-release-notes.yaml
Johannes-Schneider: Of course there is: You could simply `exit 1`.

However, this would mark the workflow as "Failed" - rendering a big red cross.
If we are simply not setting these variables, the following steps will be skipped (there is nothing to be done).
Therefore, the workflow run will indeed succeed (green checkmark).

.pipeline/scripts/dependabot-release-notes.py
Johannes-Schneider: With this automation in place, our dependency updates would (as of now) become much more verbose.

Therefore, I thought hiding them in a `<details>` block wouldn't be as disturbing as having like 30 lines of dependency updates in every release.
Also: Do consumers really care which patch versions we are using?

Additionally, having a table instead of the free-text updates might look a bit cleaner.
But that is ofc up for discussion.

.github/workflows/dependabot-release-notes.yaml
newtork: According [to the docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) you could add an `if:` statement to the `job` entry. Maybe an unmet if-condition will result in the job status not being shown 🤔 Actually with that we could save the whole condition-checking on each step, maybe.

Johannes-Schneider: Initially, I didn't use that because I anticipated that someone (maybe even just dependabot itself) might update the PR.
For example, there might be yet another new version for the same dependency group and that - maybe - could have been introduced in an existing PR.
Or, another example, someone from the dev team manually reverted one of the updates (e.g. to keep compatibility with the SJB).

In those cases, I thought it might be handy if the automation just picks up the change and re-creates the release notes.

However, this introduced a lot of traffic for **EVERY** PR (because we would use the `pull_request: action: synchronize` trigger).
As we are now just using the `opened` trigger, it should be fine to use the global `if` as in [8229dba](https://github.com/SAP/cloud-sdk-java/pull/265/commits/8229dba656848ab0a237c09ac7e011c3c016735b).

.github/workflows/dependabot-release-notes.yaml
```diff
@@ -0,0 +1,91 @@
...
+  workflow_dispatch:
+    inputs:
+      pr-number:
+        description: "Pull Request Number"
+        required: true
+
+jobs:
+  update-release-notes:
+    name: "Update Release Notes"
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && github.event.pull_request.user.login == 'app/dependabot') }}
```
Johannes-Schneider, line 17: TODO: This workflow is currently also triggered for updates to our github actions. But we don't want to mention those in our release notes.

.pipeline/scripts/dependabot-release-notes.py
```diff
@@ -0,0 +1,536 @@
...
+
+    pr_body: str = pr_body_file.read_text()
+
+    release_notes_file: Path = Path(args.release_notes)
+    if not release_notes_file.is_file():
+        raise Exception(f"'{args.release_notes}' does not exist.")
+
+    old_release_notes: str = release_notes_file.read_text()
+
+    new_updates: list[DependencyUpdate] = PrBodyParser.parse(pr_body)
```
Johannes-Schneider, line 339: TODO: we should apply some sort of filtering to these updates. It doesn't make sense to mention plugins, test-only dependencies, etc. in our release notes.

.github/workflows/dependabot-release-notes.yaml
Johannes-Schneider: Non-Maven dependencies are now ignored: [9cdfa97](https://github.com/SAP/cloud-sdk-java/pull/265/commits/9cdfa979220f6d6c6de58467fa2f45d8e04457d9)

.pipeline/scripts/dependabot-release-notes.py
Johannes-Schneider: Done in [7be2952](https://github.com/SAP/cloud-sdk-java/pull/265/commits/7be2952737a181183a18e01ad8d5be3b6f843462).

newtork (approved):

.github/workflows/dependabot-release-notes.yaml
```diff
@@ -0,0 +1,101 @@
...
+            echo "NOTES_CREATED=false" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          echo "NOTES_CREATED=true" >> $GITHUB_ENV
+          git commit -m "chore: add release notes"
+          git push
+
+      - name: "Add PR Comment"
+        if: ${{ env.PR_NUMBER != '' && env.NOTES_CREATED == 'true' && github.event_name == 'pull_request' && github.event.action == 'opened' }}
```
cschubertcs, line 88: Maybe I'm just missing it, but what happens if the job is not automatically triggered by a pull request, but manually with a PR number? If I read this correctly we don't do anything in that case?

cschubertcs: Ah, nervermind, we just add it to release notes, without a comment. I missed the previous step.

PR #269, opened 2024-01-31
feat: OAuth Property Supplier for IAS

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#391.

This PR adds a new `OAuth2PropertySupplier` implementation that is capable of reading IAS service bindings.
With that, it should - in theory - be possible to fetch tokens for [App-to-Service communication scenarios](https://github.wdf.sap.corp/CPSecurity/Knowledge-Base/blob/master/03_ApplicationSecurity/ProofOfPossession.md#reuse-service-token).

<details>
<summary>Manual E2E Test</summary>

The implementation has been manually E2E tested by extending our [scp-cf-spring-ias-java-17](https://canary.cockpit.btp.int.sap/cockpit/#/globalaccount/s4HanaCloud/subaccount/a89ea924-d9c2-4eab-84fb-3ffcaadf5d24/org/2145ac4f-0ce2-4ecc-aabe-7ef53d1ee1db/space/a140b93a-f555-4538-87eb-ef58ac26edd1/app/5e2f3620-3a25-4da1-b565-12e9b455e3eb/overview) test with the following service:

```java
@RestController
@RequestMapping("/ias")
public class IasSpringController {
    private static final Logger logger = LoggerFactory.getLogger(IasSpringController.class);

    @RequestMapping( method = RequestMethod.GET, path = "/token" )
    public ResponseEntity<?> getToken()
    {
        try {
            logger.info("Requesting IAS token on behalf of '{}'", TenantAccessor.getCurrentTenant());

            final ServiceBindingDestinationOptions options =
                    ServiceBindingDestinationOptions.forService(ServiceIdentifier.of("identity")).build();
            final HttpDestination destination =
                    ServiceBindingDestinationLoader.defaultLoaderChain().getDestination(options);

            final String authorizationHeaderValue =
                    destination
                            .getHeaders()
                            .stream()
                            .filter(h -> HttpHeaders.AUTHORIZATION.equalsIgnoreCase(h.getName()))
                            .map(Header::getValue)
                            .findFirst()
                            .get();

            return ResponseEntity
                    .ok(authorizationHeaderValue.substring(0, Math.min(8, authorizationHeaderValue.length() - 8)) + "****");
        }
        catch( final Exception e ) {
            logger.error("Exception while fetching IAS token.", e);
            return ResponseEntity.internalServerError().build();
        }
    }
}
```

[cf.logs.zip](https://github.com/SAP/cloud-sdk-java/files/14113389/cf.logs.zip)

</details>

### Feature scope:

 
- [x] Implement new `OAuth2PropertySupplier` for IAS
- [x] Refactor logic of how the `OAuth2TokenEndpointsProvider` is created
  - This was needed as the IAS token endpoints are slightly different from the XSUAA ones 

## Definition of Done

- [x] ~Documentation updated~ (not meant for the public audience just yet)
- [x] ~Release notes updated~ (not meant for the public audience just yet)

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
@@ -19,6 +19,7 @@ import com.sap.cloud.sdk.cloudplatform.connectivity.BtpServiceOptions.WorkflowOp
 import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationAccessException;
 import com.sap.cloud.sdk.cloudplatform.tenant.TenantAccessor;
 import com.sap.cloud.sdk.cloudplatform.tenant.TenantWithSubdomain;
+import com.sap.cloud.security.xsuaa.client.OAuth2ServiceEndpointsProvider;
 
 import io.vavr.control.Try;
 
@@ -141,7 +142,7 @@ class BtpServicePropertySuppliers
                     .map(TenantWithSubdomain.class::cast)
                     .map(TenantWithSubdomain::getSubdomain)
                     // TODO: this somewhat feels very fragile. Is there a better way?
-                    .map(subdomain -> "https://" + subdomain + "." + domain + "/oauth2/authorize")
+                    .map(subdomain -> "https://" + subdomain + "." + domain)
                     .map(URI::create);
 
             if( maybeTokenUri.isSuccess() ) {
@@ -152,6 +153,13 @@ class BtpServicePropertySuppliers
                 null,
                 "Unable to determine the IAS token URI.",
                 maybeTokenUri.getCause());
+        }
+
+        @Nonnull
+        @Override
+        public OAuth2ServiceEndpointsProvider getTokenEndpoints()
+        {
+            return DefaultTokenEndpoints.fromIasUri(getTokenUri());
         }
     }
 }
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
@@ -17,6 +17,10 @@ import com.sap.cloud.sdk.cloudplatform.connectivity.BtpServiceOptions.BusinessLo
 import com.sap.cloud.sdk.cloudplatform.connectivity.BtpServiceOptions.BusinessRulesOptions;
 import com.sap.cloud.sdk.cloudplatform.connectivity.BtpServiceOptions.WorkflowOptions;
 import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationAccessException;
+import com.sap.cloud.sdk.cloudplatform.tenant.TenantAccessor;
+import com.sap.cloud.sdk.cloudplatform.tenant.TenantWithSubdomain;
+
+import io.vavr.control.Try;
 
 class BtpServicePropertySuppliers
 {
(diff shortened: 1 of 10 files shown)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -110,4 +121,45 @@ public URI getServiceUri()
...
+        public URI getTokenUri()
+        {
+            final String domain = getCredentialOrThrow(String.class, "domain");
+
+            final Try<URI> maybeTokenUri =
+                TenantAccessor
+                    .tryGetCurrentTenant()
+                    .filter(TenantWithSubdomain.class::isInstance)
+                    .map(TenantWithSubdomain.class::cast)
+                    .map(TenantWithSubdomain::getSubdomain)
```
MatKuhr, line 143: This approach also has the problem that the resulting destination will be single tenant, i.e. you can only execute requests for the tenant it was created with. Currently our other OAuth based destinations are multi-tenant, i.e. the tenant at the time of execution is used.

So we should probably explore what it would take to move this logic to the OAuth service instead. BTW: the subdomain replacement is how it used to work for XSUAA in the past too.

Johannes-Schneider: Good point.

Should be solved with [f41b7ea](https://github.com/SAP/cloud-sdk-java/pull/269/commits/f41b7ea503ed94dbc65e256a376531b94b3698e0).

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/SecurityLibWorkarounds.java
```diff
@@ -0,0 +1,27 @@
+package com.sap.cloud.sdk.cloudplatform.connectivity;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import com.sap.cloud.security.config.CredentialType;
+
+final class SecurityLibWorkarounds
```
MatKuhr, line 8: yikes, not again 🫠

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2PropertySupplier.java
```diff
@@ -20,6 +26,46 @@
...
 public interface OAuth2PropertySupplier
 {
+    /**
+     * A default implementation of the {@link OAuth2ServiceEndpointsProvider} interface that provides convenient access
+     * to the default endpoints of the XSUAA service.
+     *
+     * @since 5.2.1
+     */
+    @Value
+    class DefaultTokenEndpoints implements OAuth2ServiceEndpointsProvider
```
MatKuhr, line 36: I think we only need the token path, so unless we have some use case for the other values this seems a bit overkill..

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2PropertySupplier.java
```diff
@@ -43,6 +89,22 @@ public interface OAuth2PropertySupplier
...
+     * <p>
+     * By default, this method returns endpoints that are specific to the XSUAA service.
+     * <p>
+     * For example, authorization tokens will be retrieved from {@code getTokenServiceUri().resolve("/oauth/token")}.
+     *
+     * @return An instance of {@link OAuth2ServiceEndpointsProvider}.
+     * @since 5.2.1
+     */
+    @Nonnull
+    default OAuth2ServiceEndpointsProvider getTokenEndpoints()
```
MatKuhr, line 103: I guess the idea here is to not break the behaviour of just making `getTokenUri` return the full URL?

Maybe we could change the `OAuth2Service` to append `/oauth/token` by default if the URL has no path to make it a compatible change..

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -110,4 +123,70 @@ public URI getServiceUri()
...
+        }
+
+        @Nonnull
+        @Override
+        public URI getTokenUri()
+        {
+            final URI providerUrl = getCredentialOrThrow(URI.class, "url");
+            final String domain = getCredentialOrThrow(String.class, "domain");
+
+            return TenantAccessor
```
MatKuhr, line 141: why is this code still here? I figured we can simply return the URL here as-is?

Johannes-Schneider:
Continued in #278

PR #270, opened 2024-02-01
Simplify Fosstars Action

rnewbigging:
## Context
 

SAP/cloud-sdk-java-backlog#370

Removed the Fosstars action from prepare-release workflow as it is duplicated in the daily fosstar-report workflow. Changed the reporting branch to `main`. 

### Feature scope:

 
- [x] removed duplicate workflow
- [x] switched reporting branch to `main`

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


rnewbigging:
Tested on [my fork](https://github.com/rnewbigging/cloud-sdk-java/tree/main), reports publish fine on main in new directory.

.github/workflows/fosstars-report.yml
```diff
@@ -12,5 +12,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: SAP/fosstars-rating-core-action@v1.14.0
         with:
-          report-branch: fosstars-report
+          report-branch: main
+          report-file: .github/fosstars/fosstars_report.md
+          badge-file: .github/fosstars/fosstars_badge.svg
```
Johannes-Schneider, line 17: Good idea to move these files into our `.github` folder.

.github/workflows/prepare_release.yml
```diff
@@ -336,22 +336,9 @@ jobs:
...
-    runs-on: ubuntu-latest
-    needs: prepare-release
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          ref: ${{ needs.prepare-release.outputs.branch_name }}
-      - uses: SAP/fosstars-rating-core-action@v1.14.0
-        with:
-          report-branch: fosstars-report
-          token: "${{ secrets.GITHUB_TOKEN }}"
```
Johannes-Schneider, line 350: I think we shouldn't remove this check here.
After all, there might have been changes that were not scanned by the daily job.

So for compliance reasons, running the scan once again on the actual code that is going to be released seems like a good idea.

.github/workflows/fosstars-report.yml
```diff
@@ -12,5 +12,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: SAP/fosstars-rating-core-action@v1.14.0
         with:
-          report-branch: fosstars-report
+          report-branch: main
+          report-file: .github/fosstars/fosstars_report.md
```
Johannes-Schneider, line 16: [question]

Does the action infer the location of the `fosstars_rating.json` file from the `report-file` location?

rnewbigging: I had read that was the case but just double checked my fork and it seems it didn't generate the json 🤔 I'll have another look at this.

rnewbigging: Okay I've found the issue. So the json _is_ being generated, but only to the branch parent directory, i.e. it doesn't seem possible to set a subdirectory like with the markdown report.

So I've been trying to add in a step to move the json after the report is generated, but so far it's been unsuccessful. The job echos that the file exists and it will be moved and the job completes successfully but the file is still in the parent directory. 

I'll push the changes and maybe you'll see directly where I've overseen some syntax detail 😆

.github/workflows/fosstars-report.yml
```diff
@@ -12,5 +12,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: "Move JSON to Fosstars folder"
+        run: |
+          if [ -e fosstars_rating.json ]; then
+            echo "File exists."
+            mv fosstars_rating.json .github/fosstars/
+          else
+            echo "File does not exist."
+          fi
```
Johannes-Schneider, line 33: The issue here is that you are moving the file on the file system of our runner, but you are not commiting and pushing the (local) changes to our repository. 😅

---

As a side note: Using a second _job_ here (instead of just an other _step_ in the previous _job_) seems a bit unnecessary. Or did you have a particular reason in mind for that design?

.github/workflows/fosstars-report.yml
Johannes-Schneider: > I'll push the changes

That the right approach 😉
https://github.com/SAP/cloud-sdk-java/pull/270#discussion_r1477757065

.github/workflows/fosstars-report.yml
rnewbigging: 🤦‍♀️ Thanks for the tip haha. 

And to the job/step question, no, I was thinking about it over the weekend and thought the same as well, I don't know why I went with job in the first place, Friday brain I guess. :)

.github/workflows/fosstars-report.yml
```diff
@@ -7,27 +7,60 @@ on:
...
   create_fosstars_report:
     runs-on: ubuntu-latest
-    name: "Security rating"
+    outputs:
+      json_file: ${{ steps.fosstars_rating.outputs.json_file }}
     steps:
-      - uses: actions/checkout@v3
-      - uses: SAP/fosstars-rating-core-action@v1.14.0
+      - name: Checkout
+        uses: actions/checkout@v3
```
Johannes-Schneider, line 14: [minor]

```suggestion
        uses: actions/checkout@v4
```

.github/workflows/fosstars-report.yml
```diff
@@ -7,27 +7,60 @@ on:
...
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
```
Johannes-Schneider, line 16: [minor]

Specifying the GITHUB_TOKEN is not needed here.

```suggestion
```

.github/workflows/fosstars-report.yml
```diff
@@ -7,27 +7,60 @@ on:
...
           report-branch: main
           report-file: .github/fosstars/fosstars_report.md
           badge-file: .github/fosstars/fosstars_badge.svg
           token: "${{ secrets.GITHUB_TOKEN }}"
 
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: json_report
+          path: fosstars_rating.json
```
Johannes-Schneider, line 31: Why are we not simply moving the artifact within the same `job`? It seems a bit cumbersome to first upload the artifact just to then download and move the file in a different `job`.

Besides that, we can also skip the additional `checkout` step.

.github/workflows/fosstars-report.yml
rnewbigging: An additional note to this. I ended up needing to keep it in two jobs because at least on my repo moving directly from the fosstars action to the commit attempt resulted in an error along the lines of "no permission to commit", but checking out again in the second job resolved this. I couldn't find any reason for this online but if you have ideas why it's the case I'd be interested :)

.github/workflows/fosstars-report.yml
rnewbigging: see [comment](https://github.com/SAP/cloud-sdk-java/pull/270#discussion_r1481145823)

rnewbigging:
Closing this PR as our branch protections prevents Fosstars from reporting directly to `main` and review determined both actions are still needed.

PR #278, opened 2024-02-06
feat: Support IAS in `OAuth2Service`

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#391.
Alternative to #269 

This PR extends our `OAuth2Service` implementation to handle IAS token retrievals.
In contrast to the changes suggested in #269, we are **not** introducing any new public API.

<details>
<summary>Manual E2E Test</summary>

I used following controller in our [v5-scp-cf-spring-ias-sms-subscription test](https://sdk-v2.cpe.c.eu-de-1.cloud.sap/job/end-to-end-tests/view/Cloud%20SDK%20v5/job/v5-scp-cf-spring-ias-sms-subscription/):

```java
@RestController
@RequestMapping( "/ias" )
public class IasController
{
    private static final Logger logger = LoggerFactory.getLogger(IasController.class);

    @GetMapping( path = "/token" )
    protected ResponseEntity<?> testIasTokenRetrieval()
    {
        final String tenantId =
                TenantAccessor.tryGetCurrentTenant().map(Tenant::getTenantId).getOrElse("<no-tenant-id>");
        logger.info("Retrieving IAS token on behalf of tenant '{}'.", tenantId);

        try {
            final ServiceBindingDestinationOptions options =
                    ServiceBindingDestinationOptions.forService(ServiceIdentifier.of("identity")).build();
            final HttpDestination destination =
                    ServiceBindingDestinationLoader.defaultLoaderChain().getDestination(options);

            final String authorizationValue =
                    destination
                            .getHeaders()
                            .stream()
                            .filter(h -> HttpHeaders.AUTHORIZATION.equalsIgnoreCase(h.getName()))
                            .map(Header::getValue)
                            .findFirst()
                            .get();
            final String trimmedAuthorizationValue =
                    authorizationValue.substring(0, Math.min(10, authorizationValue.length()));
            logger
                    .info(
                            "Retrieved IAS authorization token '{}' on behalf of tenant '{}'.",
                            trimmedAuthorizationValue,
                            tenantId);

            return ResponseEntity.ok().body(trimmedAuthorizationValue);
        }
        catch( final Exception e ) {
            logger.error("Failed to retrieve IAS token on behalf of tenant '{}'.", tenantId, e);
            return ResponseEntity
                    .internalServerError()
                    .body("Failed to retrieve IAS token on behalf of tenant '" + tenantId + "'.");
        }
    }
}
```

With that, I ran following scenarios:

1. Login via App Router using the _Provider_ (`id = "a89ea924-d9c2-4eab-84fb-3ffcaadf5d24"; subdomain = "ma"`) tenant: ✅ Working as expected.
2. Login via App Router using the _Subscriber_ (`id = "7a290e94-70de-4583-a3cb-b0217bf9e4ae"; subdomain = "ahrn1puuk"`) tenant: ✅ Working as expected.
3. Call the App directly (without App Router login, therefore implicitly using the _Provider_ Tenant): ✅ Working as expected.

Application logs:
[manual-e2e-test.logs.zip](https://github.com/SAP/cloud-sdk-java/files/14194190/manual-e2e-test.logs.zip)

</details>

### Feature scope:

 
- [x] IAS Binding to Destination transformation
  - [x] Destination is multi-tenant

## Definition of Done

- [x] ~Documentation updated~ (→ not to be documented yet)
- [x] ~Release notes updated~ (→ not to be documented yet)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -195,17 +219,52 @@ private OAuth2TokenResponse executeUserExchangeFlow()
...
+            return XSUAA_ZID_HEADER;
+        }
+    }
+
     @Value
     static class Endpoints implements OAuth2ServiceEndpointsProvider
     {
         private static final String TOKEN_PATH = "/oauth/token";
-        private static final String AUTHORIZE_PATH = "/oauth/authorize";
-        private static final String KEYSET_PATH = "/token_keys";
```
cschubertcs, line 208: Is there somewhere documentation on why those were removed?
I remember, that we never really used them, but assumed that they might have some use down the line?

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2IntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+                    .willReturn(okJson(RESPONSE_TEMPLATE.formatted(token))));
+
+            // no tenant - provider case
+            assertThat(TenantAccessor.tryGetCurrentTenant().isFailure()).isTrue();
+            assertThat(destination.getHeaders()).contains(new Header(HttpHeaders.AUTHORIZATION, "Bearer " + token));
+
+            // call the method a second time, so we can be sure the token is cached correctly
+            assertThat(destination.getHeaders()).contains(new Header(HttpHeaders.AUTHORIZATION, "Bearer " + token));
+
+            verify(1, postRequestedFor(urlEqualTo("/oauth2/token")).withHost(equalTo("provider.ias.domain")));
```
cschubertcs, line 98: I roughly remember that with XSUAA we usually had the subdomain as the part until the first dot of the url mentioned in the service binding.

Does the IAS binding now only provide the `domain` and the subdomain gets added as a pre-fix? In the end this logic is handled by the security library, right?

cschubertcs:
Left some comments for understanding how things are supposed to work.
Besides those, this looks already good, especially that the API didn't receive any breaking changes!

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
Johannes-Schneider: As far as I'm aware, the `AUTHORIZATION_PATH` is only needed for some Auth Code flows and, therefore, not interesting to us.
The `KEYSET_PATH` is also just used to validate incoming tokens (which is a concern of the Security Lib, but doesn't matter in our case here).

But we wanted to validate those assumptions by running a few of our E2E tests.

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2IntegrationTest.java
Johannes-Schneider: No, there is still a `url` within the IAS binding. This `url`, apparently, already contains the provider subdomain.

However, we are using the security lib to take care of replacing the subdomain of the provider with the current tenant's one at runtime.

cschubertcs: Nevermind, yesterday I didn't see the `url` property in the mocked service binding above. Like that the test makes more sense to me.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -192,37 +222,227 @@ private OAuth2TokenResponse executeUserExchangeFlow()
...
+             *   but the downside is arguably not worth keeping a second configuration for this case only
+             */
+            final var resilienceConfig =
+                ResilienceConfiguration
+                    .of(tokenUri.getHost() + "-" + identity.getId())
+                    .isolationMode(ResilienceIsolationMode.TENANT_OPTIONAL)
+                    .timeLimiterConfiguration(timeLimiter);
+
+            // copy the additional parameters to prevent accidental manipulation after the `OAuth2Service` instance has been created.
+            final var additionalParameters = new HashMap<>(this.additionalParameters);
```
cschubertcs, line 432: General question/discussion for the team: Do want to use the `var` keyword, even in such small cases or rather for bigger cases (e.g. multilevel generics)?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ServiceBindingLibWorkarounds.java
```diff
@@ -0,0 +1,13 @@
+package com.sap.cloud.sdk.cloudplatform.connectivity;
+
+import com.sap.cloud.environment.servicebinding.api.ServiceIdentifier;
+
+final class ServiceBindingLibWorkarounds
```
cschubertcs, line 5: (Question) Is this planned to be migrated to the Service Binding lib?

cschubertcs (approved):
looks good to me!

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
Johannes-Schneider: I think it wouldn't be a good idea to now start using `var` as it would bring the code base in a pretty inconsistent state.
So I replaced these `var` variables.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ServiceBindingLibWorkarounds.java
Johannes-Schneider: Actually, as the IAS is officially considered GA, I added it to the Service Binding Lib in https://github.com/SAP/btp-environment-variable-access/pull/189.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -192,37 +222,228 @@ private OAuth2TokenResponse executeUserExchangeFlow()
...
+                        false);
+            default -> throw new DestinationAccessException(
+                "Unhandled TenantPropagation Strategy: %s.".formatted(tenantPropagationStrategy));
+        }
+
+        return Try.of(flow).getOrElseThrow(e -> new TokenRequestFailedException("Failed to resolve access token.", e));
+    }
+
+    @Nonnull
+    static UriBuilder builder()
```
MatKuhr, line 258: (minor) do we really need all of the interfaces here? In addition to bloating the code it also forces us to declare all of the implementing methods public, even though they are actually not public ofc

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -192,37 +222,228 @@ private OAuth2TokenResponse executeUserExchangeFlow()
...
+            withTimeLimiter( @Nonnull final ResilienceConfiguration.TimeLimiterConfiguration timeLimiter );
+
+        @Nonnull
+        OAuth2Service build();
     }
 
-    @Value
-    static class Endpoints implements OAuth2ServiceEndpointsProvider
+    @Getter( AccessLevel.PACKAGE )
+    static class Builder implements UriBuilder, IdentityBuilder, OptionalParametersBuilder
```
MatKuhr, line 307: (side note) Imagine Lombok could auto-generate these but with the ability to override individual methods, that would be the dream. I guess one could almost make that work using `@Delegate` but at least one major problem with that approach is the `delegate` would `return this` which would mess things up in most cases.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -70,81 +73,56 @@ class OAuth2Service
...
+
+        final OAuth2TokenResponse tokenResponse = ResilienceDecorator.executeSupplier(() -> switch (onBehalfOf) {
+            case TECHNICAL_USER_PROVIDER -> executeClientCredentialsFlow(null);
+            case TECHNICAL_USER_CURRENT_TENANT -> {
+                final Tenant tenant = TenantAccessor.tryGetCurrentTenant().getOrNull();
+                yield executeClientCredentialsFlow(tenant);
             }
-        }, resilienceConfig);
+            case NAMED_USER_CURRENT_TENANT -> executeUserExchangeFlow();
+            default -> throw new IllegalStateException("Unknown behalf " + onBehalfOf);
```
MatKuhr, line 115:
```suggestion
```

(minor) The default case is redundant, the Java compiler enforces all cases are handled

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -153,23 +131,75 @@ String retrieveAccessToken()
...
+                        identity,
+                        zidHeaderValue,
+                        tenantSubdomain,
+                        additionalParameters,
+                        false))
             .getOrElseThrow(e -> new TokenRequestFailedException("Failed to resolve access token.", e));
     }
 
+    @Nullable
+    private String getTenantId( @Nullable final Tenant tenant )
```
MatKuhr, line 162: (minor) Not sure if the code gains or loses clarity by extracting these two into their own methods.. IMO it hides a little bit the fact that the stuff is nullable and not expected to be there for all cases, but yeah, matter of taste

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -153,23 +131,75 @@ String retrieveAccessToken()
...
+    {
+        return tenant == null ? null : tenant.getTenantId();
+    }
+
+    @Nullable
+    private String getTenantZoneId( @Nullable final String tenantId )
```
MatKuhr, line 178: (minor) maybe we don't need to mention the zone term so prominently? sure, the header is still called x-zid, but for us we should probably always use tenant id as term. For IAS we are also not calling it app-tid or anything like that..

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -70,81 +73,56 @@ class OAuth2Service
...
+            case TECHNICAL_USER_CURRENT_TENANT -> {
+                final Tenant tenant = TenantAccessor.tryGetCurrentTenant().getOrNull();
+                yield executeClientCredentialsFlow(tenant);
             }
-        }, resilienceConfig);
+            case NAMED_USER_CURRENT_TENANT -> executeUserExchangeFlow();
+            default -> throw new IllegalStateException("Unknown behalf " + onBehalfOf);
+        }, resilienceConfiguration);
 
-        if( tokenResponse == null ) {
```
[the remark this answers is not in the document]
Johannes-Schneider: Urgh... as I was using a Java 17 "switch expression" (no idea whether this is the correct name), the formatter **silently** skipped this method when checking whether everything is fine 🙈

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
Johannes-Schneider: I guess the alternative would be to just put everything into the Builder implementation, right?
That way, we're getting around the interfaces at the cost of not having a "safe-to-use" API (which isn't too bad, considering that the class is ofc for internal use only).
Additionally, we wouldn't notice an "incorrect" usage of our APIs during development but only when actually running the code.

Even though I'm not a fan of introducing API that can be used incorrectly, I see the point that this is costing quite a bit of code.
So considering that this is internal API only, I think it should be okay to get rid of the "safety" features.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
Johannes-Schneider: I renamed the methods to `...OrNull` to improve the clarity.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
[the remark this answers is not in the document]
MatKuhr: haha okay we'll personally I'd be fine with that, considering that the modern switch is a lot more convenient, so I'd rather live with non-consistently-formatted code for such specific cases

MatKuhr (approved):
LGTM 😎

PR #280, opened 2024-02-06
chore: [DevOps] Improve Test Isolation

MatKuhr:
## Context

https://github.com/SAP/cloud-sdk-java-backlog/issues/368

This PR explicitly marks tests which rely on global, static state with `@Isolated`, guaranteeing they won't be executed in parallel to any other tests.


cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationRetrievalStrategyResolverTest.java
```diff
@@ -44,21 +42,19 @@
 
 import io.vavr.Tuple;
 import io.vavr.Tuple3;
-import io.vavr.control.Try;
 
+@Isolated // required due to implicit behavior of TokenRule
```
MatKuhr, line 46:
```suggestion
@Isolated
```

PR #290, opened 2024-02-08
feat: Introduce TestContext and improve test utilities

MatKuhr:
# Pull Request Description 🤖

## Changes
This PR introduces changes to the test utilities and the resilience4j module. The main changes are as follows:

1. Added `TestContext` class which provides a context for tests. This context can be used to set and clear properties related to tenants, principals, and auth tokens. This class also implements `AuthTokenContextApi`, `TenantContextApi`, and `PrincipalContextApi` interfaces which provide methods to set and clear these properties.

2. Refactored the existing tests to use the new `TestContext` class. This involved replacing calls to `MockUtil` methods with the corresponding methods in `TestContext`.

3. Removed the `MockUtil` class and its test class `MockUtilTest` as they are no longer needed.

4. Added `AuthTokenContextApi`, `TenantContextApi`, and `PrincipalContextApi` interfaces. These interfaces provide methods to set and clear auth tokens, tenants, and principals respectively.

5. Added `XsuaaTokenMocker` class which provides a method to create a mock XSUAA token.

6. Updated the `pom.xml` file of the testutil-core module to include `junit-jupiter-api` as a compile scope dependency.

## Impact
These changes improve the test utilities by providing a more intuitive way to set up the context for tests. This will make it easier to write and understand tests. The changes are backward compatible as they only add new functionality and do not remove or change existing functionality.


testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TenantContextApi.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.sap.cloud.sdk.cloudplatform.tenant.DefaultTenant;
+import com.sap.cloud.sdk.cloudplatform.tenant.Tenant;
+import com.sap.cloud.sdk.cloudplatform.tenant.TenantThreadContextListener;
+
+public interface TenantContextApi extends TestContextApi
+{
+    default Tenant setTenant()
+    {
+        return setTenant("tenant");
```
Johannes-Schneider, line 11: [minor]

Let's make it a bit easier to search for this particular tenant id, just in case we are ever wondering where this tenant is coming from.

```suggestion
        return setTenant("default-test-tenant");
```

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/PrincipalContextApi.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.sap.cloud.sdk.cloudplatform.security.principal.DefaultPrincipal;
+import com.sap.cloud.sdk.cloudplatform.security.principal.Principal;
+import com.sap.cloud.sdk.cloudplatform.security.principal.PrincipalThreadContextListener;
+
+public interface PrincipalContextApi extends TestContextApi
+{
+    default Principal setPrincipal()
+    {
+        return setPrincipal("principal");
```
Johannes-Schneider, line 11: [minor]

Same remark as for the `TenantContentApi`.

```suggestion
        return setPrincipal("default-test-principal");
```

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TestContext.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+        });
+    }
+
+    @Override
+    public void afterEach( final ExtensionContext extensionContext )
+    {
+        if( resetCaches ) {
+            CacheManager.invalidateAll();
+        }
```
Johannes-Schneider, line 88: [comment]

This code makes the current race conditions with our caching pretty obvious:

As our caches are _true_ `static` fields (in contrast to our `ThreadContext`, which obviously uses `ThreadLocal`), there is the chance that some tests fail due to them relying on some caching behavior, but a second, totally unrelated, test might have reset our caches in between.

Of course, this race condition is not being introduced in this PR, but the symptoms might get worse as we are resetting all caches _by default_ when using this extension.
In contrast to that, we are using this particular API pretty rarely in our unit current tests (right? Maybe I'm just not aware that we are using it very commonly?).

Also: There are some caches (e.g. the one in our `OAuth2Service`) that are not registered within our `CacheManager`.
So we are not really catching everything in here.

Johannes-Schneider (commented):
I just looked at the new `TestContext` extension.

Personally, I like the idea of using our `ThreadContext` instead of manipulating the `static` Facades in our accessors.

I'm just a bit concerned about resetting our caches after every test (see my comment).

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TestContext.java
MatKuhr: Yes, I am currently in the process of making this reset behavior the non-default, and highlight it should only be used together with `@Isolated` to be safe.

Yep, I also immediately thought of `OAuth2Service` in that regard 😄 
I was considering adding it in a separate PR. Still, some cache reset behavior is not triggered by the `CacheManager`, e.g. enabling/disabling change detection. So yeah, there is still a bit of thinking to be done what exactly we want to put in here and what not.

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TestContextApi.java
```diff
@@ -0,0 +1,20 @@
+package com.sap.cloud.sdk.testutil;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Baseline interface for the {@code TestContext} interfaces.
+ */
+public interface TestContextApi
```
Johannes-Schneider, line 9: [comment - just a thought]

Do we actually need this new interface? What if we were to simply use the actual `ThreadContext` instead?
We could then use `@Delegate` in the implementation, with some convenience methods (i.e. the ones from the `TenantContext`, ...) on top.

Not really sure if that approach yields any benefits, but it could be "cool" 😅

testutil/testutil-core/src/test/java/com/sap/cloud/sdk/testutil/TestContextTest.java
```diff
@@ -0,0 +1,145 @@
...
+        @BeforeEach
+        @AfterEach
+        void beforeAfterEachSanityCheck()
+        {
+            sanityCheck();
+        }
+
+        @Test
+        @DisplayName( "The context should contain the prefilled values, regardless of how exactly they were filled." )
+        void testPrefilledThreadContext()
```
Johannes-Schneider, line 122: [idea]

What if our new Test extension would work similarly to the WireMock JUnit5 extension.

In other words, what if we could define a test method like this:

```java
@Test
void testWithSomeContext(final TestContext tc, final Tenant t, final Principal p, final AuthToken a) {
...
}
```

Hereby, our TestExtension could look at the method parameters to automatically create some instances **and** put them into a new `ThreadContext` 🤯
(We could still keep the "static"/manual setup logic ofc, so simplify the setup for classes that use a common set of properties).

testutil/testutil-core/src/test/java/com/sap/cloud/sdk/testutil/TestContextTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        @Test
+        @DisplayName( "Overriding prefilled values should be possible." )
+        void testPrefilledPropertiesCanBeOverriden()
+        {
+            assertThat(TenantAccessor.tryGetCurrentTenant()).contains(tenant);
+            sut.setTenant("newTenant");
+            assertThat(TenantAccessor.tryGetCurrentTenant())
+                .doesNotContain(tenant)
+                .contains(new DefaultTenant("newTenant"));
```
Johannes-Schneider, line 140: [minor]

```suggestion
            final Tenant newTenant = sut.setTenant("newTenant");
            assertThat(TenantAccessor.tryGetCurrentTenant())
                .doesNotContain(tenant)
                .contains(newTenant);
```

cloudplatform/caching/pom.xml
```diff
@@ -83,6 +83,10 @@
 			<artifactId>commons-lang3</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.sap.cloud.sdk.testutil</groupId>
+			<artifactId>testutil-core</artifactId>
```
Johannes-Schneider, line 88: [comment]

I know that the `<scope>test` is managed in our parent pom, but it looks very odd to not have this declared explicitly here.

cloudplatform/cloudplatform-connectivity/pom.xml
```diff
@@ -130,5 +130,10 @@
 			<artifactId>mockito-core</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.sap.cloud.sdk.testutil</groupId>
+			<artifactId>testutil-core</artifactId>
+			<scope>test</scope>
```
Johannes-Schneider, line 136: [minor]

It seems we are using the `<scope>` somewhat inconsistently. Let's decide for one approach and then try to stick with it.

Johannes-Schneider (commented):
I didn't check all of the test changes yet, but left some idea/thoughts already.

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TestContext.java
```diff
@@ -0,0 +1,166 @@
...
+        @Nonnull final ReflectiveInvocationContext<Method> invocationContext,
+        @Nonnull final ExtensionContext extensionContext )
+        throws Throwable
+    {
+        if( !withThreadContext ) {
+            invocation.proceed();
+            return;
+        }
+        ThreadContextExecutor.using(context).execute(() -> {
+            try {
```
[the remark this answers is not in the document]
MatKuhr: Hmm not sure what you mean, exactly..

> However, we are always replacing any existing ThreadContext with our instance in the interceptTestMethod class.

There is no existing ThreadContext at the time `interceptTestMethod` is called. While you can create a ThreadContext, you can't set a ThreadContext during the test setup / before the execution (unless you modify the ThreadContextFacade, but then probably nothing works as intended anymore). Because the context gets duplicated when calling `ThreadContextExecutor.using(context)` this setup works and is tested e.g. [here](https://github.com/SAP/cloud-sdk-java/pull/290/files#diff-009cea225e95f9223b7ad8988e15c74fce163364edba9a997b7097f819990b46R134).

> a scenario, where I'm manually creating a ThreadContext during my test setup

You can still set your desired context within your test execution of course and override the context coming from the junit extension. And for the part where you execute your own ThreadContext within your test the `setProperty(..)` method has no effect, but I'd argue that is working as it should be?

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TestContextApi.java
MatKuhr: Currently we only have this because all of the default methods in the other interfaces need access to `setProperty` which they gain by extending this interface..

For now this seemed the easiest to avoid having a 1k LoC `TestContext` class with all the boilerplate overloads 😄

testutil/testutil-core/src/test/java/com/sap/cloud/sdk/testutil/TestContextTest.java
MatKuhr: Yeah was thinking about something like that too, could be cool convenience 👍🏻 

Maybe `@WithTestContext( mockDefaults = true ) ` where you can specify if you want tenant/principal/token to be set by default or not. 

Although I'd probably only pass the `TestContext` to the method which the method could use to set more stuff...

testutil/testutil-core/src/main/java/com/sap/cloud/sdk/testutil/TestContext.java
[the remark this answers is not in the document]
Johannes-Schneider: > There is no existing ThreadContext at the time interceptTestMethod is called. While you can create a ThreadContext, you can't set a ThreadContext during the test setup / before the execution (unless you modify the ThreadContextFacade, but then probably nothing works as intended anymore).

You're right. I was thinking about a scenario where a `ThreadContext` already exists at the time of an `@BeforeEach/All` method, but that cannot happen unless one is indeed overriding the facade.

So I figured the code makes sense to support a scenario like the following:

```java
@RegisterExtension
static TestContext sut = TestContext.withThreadContext();
static Tenant tenant;

@BeforeEach
void setTenant() {
    tenant = sut.setTenant();
}

@Test
void test() {
    // sanity
    assertThat(TenantAccessor.tryGetCurrentTenant()).contains(tenant);
    final var initialContext = ThreadContextAccessor.getCurrentContext();

    ThreadContextExecutors.execute(() -> {
        final var newTenant = sut.setTenant("new-tenant");

        assertThat(TenantAccessor.tryGetCurrentTenant()).contains(newTenant);
        assertThat(ThreadContextAccessor.getCurrentContext()).isNotSameAs(initialContext);
    });
}
```

testutil/testutil-core/src/test/java/com/sap/cloud/sdk/testutil/TestContextTest.java
Johannes-Schneider: > Although I'd probably only pass the TestContext to the method which the method could use to set more stuff...

The idea was to make that a bit more dynamic.
So you don't have to specify all of these parameters - if you don't need an `AuthToken`, just don't mention it in the method signature.
The extension could be smart enough to figure that out by itself by just looking at the method signature.

Similarly, if I just want a tenant to be present, but I don't need to further adjust my current `ThreadContext`, just specify the `final Tenant` parameter in your method signature. The extension will take care of creating a `ThreadContext` with a tenant set under the hood for you.
(That way, we don't need yet another annotation just mock a few defaults)

Nevertheless, we probably wouldn't make too much use of that feature in our unit tests to justify the implementation effort (even though it would be a cool exercise IMO 😉).

MatKuhr: Yeah, while that would be cool, I'd say it's probably somewhat unintuitive to have the presence of a method parameter determine the implicit behavior of whether or not a tenant is present, as a reader one only understands the code if you look into the details how TenatContext works 😄

Johannes-Schneider (approved):
I didn't check each and every changed test line, but those that I checked looked good to me.

PR #293, opened 2024-02-09
feat:Allow manipulation of Urls read from service binding

KavithaSiva:
## Context
 

https://github.wdf.sap.corp/MA/sdk/issues/8706

With the above support issue we discovered that `BtpServicePropertySuppliers`  supplies urls provided in the service binding directly without any pre-processing for the `Destination` to be built from the binding.

Although, this works in most cases, most notably in the Business logging service, this causes issues as the service binding seems to have some path appended to the urls, which gets appended later on while making calls from clients generated for this service.

for e.g. in the given binding
```
"endpoints": {
  "configservice": "https://business-logging.config_api.example.com/buslogs/configs",
  "readservice": "https://business-logging.read_api.example.com/odata/v2/com.sap.bs.businesslogging.DisplayMessage",
  "textresourceservice": "https://business-logging.text_api.example.com/buslogs/configs/textresource",
  "writeservice": "https://business-logging.write_api.example.com.cfapps.sap.hana.ondemand.com/buslogs/log"
}
```

For e.g Ideally the Destination created for `writeservice` should only have `https://business-logging.write_api.example.com.cfapps.sap.hana.ondemand.com` and not include the additional path "/buslogs/log".

This PR aims at removing such known redundant paths from the service binding URLs obtained.

### Feature scope:

 
-  [x] Add logic to remove redundant path from URL
- [ ] Add tests

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
@@ -55,7 +55,7 @@ class BtpServicePropertySuppliers
                     .withUrlKey(BusinessLoggingOptions.CONFIG_API, "configservice")
                     .withUrlKey(BusinessLoggingOptions.TEXT_API, "textresourceservice")
                     .withUrlKey(BusinessLoggingOptions.READ_API, "readservice")
-                    .withUrlKey(BusinessLoggingOptions.WRITE_API, "writeservice")
+                    .withUrlKey(BusinessLoggingOptions.WRITE_API, "writeservice","buslogs/log")
                     .factory());
 
     private static final List<OAuth2PropertySupplierResolver> DEFAULT_SERVICE_RESOLVERS = new ArrayList<>();
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
@@ -52,10 +52,10 @@ class BtpServicePropertySuppliers
                 ServiceIdentifier.of("business-logging"),
                 MultiUrlPropertySupplier
                     .of(BusinessLoggingOptions.class)
-                    .withUrlKey(BusinessLoggingOptions.CONFIG_API, "configservice")
-                    .withUrlKey(BusinessLoggingOptions.TEXT_API, "textresourceservice")
+                    .withUrlKey(BusinessLoggingOptions.CONFIG_API, "configservice", "/buslogs/configs")
+                    .withUrlKey(BusinessLoggingOptions.TEXT_API, "textresourceservice", "/buslogs/configs/textresource")
                     .withUrlKey(BusinessLoggingOptions.READ_API, "readservice")
-                    .withUrlKey(BusinessLoggingOptions.WRITE_API, "writeservice","buslogs/log")
+                    .withUrlKey(BusinessLoggingOptions.WRITE_API, "writeservice", "buslogs/log")
                     .factory());
 
     private static final List<OAuth2PropertySupplierResolver> DEFAULT_SERVICE_RESOLVERS = new ArrayList<>();
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
@@ -5,9 +5,9 @@
 package com.sap.cloud.sdk.cloudplatform.connectivity;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.function.Function;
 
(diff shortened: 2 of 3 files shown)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -52,10 +52,10 @@ class BtpServicePropertySuppliers
                 ServiceIdentifier.of("business-logging"),
                 MultiUrlPropertySupplier
                     .of(BusinessLoggingOptions.class)
-                    .withUrlKey(BusinessLoggingOptions.CONFIG_API, "configservice")
-                    .withUrlKey(BusinessLoggingOptions.TEXT_API, "textresourceservice")
+                    .withUrlKey(BusinessLoggingOptions.CONFIG_API, "configservice", "/buslogs/configs")
+                    .withUrlKey(BusinessLoggingOptions.TEXT_API, "textresourceservice", "/buslogs/configs/textresource")
                     .withUrlKey(BusinessLoggingOptions.READ_API, "readservice")
-                    .withUrlKey(BusinessLoggingOptions.WRITE_API, "writeservice")
+                    .withUrlKey(BusinessLoggingOptions.WRITE_API, "writeservice", "buslogs/log")
```
Johannes-Schneider, line 58: [major]

This is a behavior breaking change, which - personally - I wouldn't declare as a "bugfix".
So this is a problem IMO.

---

From the customer issue, it looks like the service binding contains further properties:

```
{
    "configUrl": "https://buslog-prod-business-logging-core-service-appconfig.cfapps.eu20.hana.ondemand.com",
    "configurationPath": "buslogs/configs",
    "endpoints": {
        "configservice": "https://buslog-prod-business-logging-core-service-appconfig.cfapps.eu20.hana.ondemand.com/buslogs/configs",
        "readservice": "https://buslog-prod-business-logging-odata-service.cfapps.eu20.hana.ondemand.com/odata/v2/com.sap.bs.businesslogging.DisplayMessage",
        "readservicev4": "https://buslog-prod-business-logging-odata-service.cfapps.eu20.hana.ondemand.com/odata/v4/com.sap.bs.businesslogging.DisplayMessage",
        "textresourceservice": "https://buslog-prod-business-logging-core-service-appconfig.cfapps.eu20.hana.ondemand.com/buslogs/configs/textresource",
        "writeservice": "https://buslog-prod-business-logging-write-service.cfapps.eu20.hana.ondemand.com/buslogs/log"
    },
    "readLogPath": "/odata/v2/com.sap.bs.businesslogging.DisplayMessage",
    "readLogPathv4": "/odata/v4/com.sap.bs.businesslogging.DisplayMessage",
    "readUrl": "https://buslog-prod-business-logging-odata-service.cfapps.eu20.hana.ondemand.com",
    "writeLogPath": "buslogs/log",
    "writeUrl": "https://buslog-prod-business-logging-write-service.cfapps.eu20.hana.ondemand.com"
}
```

For every entry in `endpoints`, there seems to be a `<service>Url` entry (on root level).
(`endpoints.configservice` → `configUrl`, `endpoints.readservice` → `readUrl`, ...)

So instead of changing the existing behavior, what if we were to introduce new entries in the `BusinessLoggingOptions`, such as `BASE_WRITE_API` (name TBD), which then points to these properties on root level?

KavithaSiva: >This would be a behavior breaking change, right?

Yes, it is but, the URLs supplied now, wouldn't work with the generated clients. 
Wouldn't that be the basic use case we are trying to satisfy?

KavithaSiva: >So instead of changing the existing behavior, what if we were to introduce new entries in the BusinessLoggingOptions, such as BASE_WRITE_API (name TBD), which then points to these properties on root level?

I agree that we can introduce new `BusinessLoggingOptions`, but I am not sure for what purpose the `WRITE_API` would be used for if there is also a `BASE_WRITE_API`.

KavithaSiva: >For every entry in endpoints, there seems to be a <service>Url entry (on root level).
I will ask the user to attach the full binding to see if the <service>Url entry exists for all four services now in the binding.

Johannes-Schneider: > Yes, it is but, the URLs supplied now, wouldn't work with the generated clients.
Wouldn't that be the basic use case we are trying to satisfy?

I guess it depends on how exactly the specs looks like... But I'd generally agree.

Nevertheless, I'd be curious to know whether the customer is actually the first person to try this out, or whether there are other applications, that (maybe?) use a different spec file, which happens to work with our current implementation.

newtork: <details><summary>I would prefer a "BASE_WRITE_API" introduction over another argument.</summary>

Also I think `.withUrlKey(BusinessLoggingOptions.CONFIG_API, String, String)` is hard to read/understand. And I'm not sure whether this change breaks or degrades the intention the of the interface 🤔 

Don't have much more-productive feedback :S
</details>

KavithaSiva:
We discussed with Business logging colleagues, where we pitched the idea of them contributing to our repo directly.

PR #296, opened 2024-02-12
feat/chore: Refactor test utilities module and update OAuth2Service

MatKuhr:
## Overview 🤖

This pull request includes changes related to the refactoring of the test utilities module and updates to the OAuth2Service.

## Details

- The test utilities module has been refactored. The `testutil-core` artifact has been removed and its contents have been moved to the `testutil` artifact. This change is reflected in the updates to the `pom.xml` files of various modules where the dependency on `testutil-core` has been replaced with `testutil`.

- The `OAuth2Service` has been updated to register its cache with the `CacheManager`. This change allows for better control over the cache, especially in testing scenarios. The `clearCache` method has been removed from `OAuth2Service` and the cache clearing is now handled through the `CacheManager`.

- The `OAuth2ServiceTest` has been updated to reflect the changes in `OAuth2Service`. The test now checks if the cache is registered with the `CacheManager`.

- The `surefire.logLevel` in the root `pom.xml` has been changed from `warn` to `error`.

- The `module-inventory.json` file has been updated to reflect the changes in the module structure.

## Instructions for the Reviewer

Please review the changes in the context of the refactoring of the test utilities module and the updates to the `OAuth2Service`. Pay special attention to the changes in the `OAuth2Service` and its test, and verify if the new approach to cache management is correctly implemented.


pom.xml
```diff
@@ -69,7 +69,7 @@
 		<surefire.groups />
 		<surefire.excludedGroups />
 		<surefire.maxMemorySize>1024m</surefire.maxMemorySize>
-		<surefire.logLevel>warn</surefire.logLevel>
+		<surefire.logLevel>error</surefire.logLevel>
```
MatKuhr, line 72: Maybe somewhat controversial change.. I just felt that we have a lot of spammy warning logs where the warnings are not relevant for testing (e.g. because we intentionally test broken/deprecated/internal stuff)

So to keep the log a bit cleaner and making it easier to spot actual errors

Johannes-Schneider: Personally, I'm fine with that change.

Especially since we don't need to alter the `pom.xml` to change the log level (instead, just use `-Dsurefire.logLevel=debug` when running from the CLI).

Johannes-Schneider (approved):
LGTM

PR #308, opened 2024-02-16
chore: [DevOps] Enable Auto-Merge for Dependabot PRs

MatKuhr:


.github/workflows/dependabot-automerge.yaml
```diff
@@ -0,0 +1,37 @@
+name: dependabot merger
```
Johannes-Schneider, line 1: For anyone who might be wondering: This workflow is a copy&paste from what we already [have in the Service Binding Lib](https://github.com/SAP/btp-environment-variable-access/blob/main/.github/workflows/dependabot-automerge.yml), where this workflow has shown to be working as expected.

There is still the risk that we might use different branch protection rules etc. in this repo here, but we should be able to figure this out once this workflow is triggered.

Johannes-Schneider (approved):
LGTM

PR #322, opened 2024-02-22
Make Service Binding Library usage backwards-compatible

newtork:
Related BLI
* https://github.com/SAP/cloud-sdk-java-backlog/issues/412

We've noticed an incompatibility with SJB (**2.2.0**):
* They haven't updated to service-binding-lib [`0.10.3`](https://github.com/SAP/btp-environment-variable-access/releases/tag/0.10.3) yet. At runtime class loading issues will occur.

More context:
* To enable SJB you need to set service-binding-lib `scope=provided`
* When using `scope=compile` duplicate entries in class-path will break class loading.


MatKuhr (approved):
For context: Done temporarily to improve compatibility with other products.

I'll create a Slack reminder to revert this a few weeks down the line 😉

PR #331, opened 2024-02-27
feat: IAS App-to-Service Communication

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#373.

This PR adds a new `ServiceBindingDestinationLoader` implementation that is capable of dealing with (currently) imaginary service bindings from re-use services that use IAS for authentication.

The suggested Service Binding format looks like this:

```jsonc
{
  "credentials": {
    "authentication-service": {
      "service-label": "identity", // mandatory
      "app-name": "my-app-name" // optional
    },
    "endpoints": {
      "some-endpoint": {
        "protocol": "http", // optional, "http" by default
        "uri": "http://some-endpoint.com", // mandatory for "http" protocol
        "always-requires-token": true, // optional, defaults to true
        "requires-mtls": true // optional, defaults to true
      }
    }
  }
}
```

### Feature scope:

 
- [x] Assume and parse a service binding format

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,142 @@
...
+    private final ServiceBindingDestinationLoader delegateLoader;
+
+    public IdentityAuthenticationServiceBindingDestinationLoader()
+    {
+        this(null);
+    }
+
+    // for testing purposes
+    IdentityAuthenticationServiceBindingDestinationLoader(
+        @Nullable final ServiceBindingDestinationLoader delegateLoader )
```
newtork, line 38: (minor)

I would replace `@Nullable` with `@Nonnull`. With the current code base, the additional null check can be avoided, right?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,142 @@
...
+                .withOption(BtpServiceOptions.IasOptions.withTargetUri(endpoint.uri));
+
+        if( endpoint.mTLSOnly ) {
+            optionsBuilder.withOption(BtpServiceOptions.IasOptions.withMTLSAuthenticationOnly());
+        }
+
+        return delegateLoader.tryGetDestination(optionsBuilder.build());
+    }
+
+    private static boolean isIdentityAuthenticationServiceBinding( @Nonnull final ServiceBinding serviceBinding )
```
newtork, line 74: (Comment)

I think this is fine for now. I wonder whether we'll need to make this an accessible / modifiable `Predicate<ServiceBinding>` instead in the future.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,142 @@
...
+
+        return allEndpoints.get(0);
+    }
+
+    @Value
+    private static class EndpointEntry
+    {
+        String name;
+        URI uri;
+        boolean mTLSOnly;
```
newtork, line 103: (Major-ly)

I also noticed this in other (recently added) code. Camel-case also applies to abbreviations in our code-base:
```suggestion
        boolean mTlsOnly;
```

It's a nitpick, but I like the consistency :)

newtork (commented):
Progress looks good. I'm only concerned about
1. customizability _(for future JSON format changes)_
1. testability _(for current event-broker JSON format)_

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
Johannes-Schneider: Agreed.

I used your remark to also refactor our existing (but not yet communicated) APIs:

```diff
- IasOptions.withMTLSOnlyAuthentication()
+ IasOptions.withMutualTlsOnly()
```

Personally, I felt like spelling our the **m** somewhat looks better.
Any opinions on those changes?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
Johannes-Schneider: As we are working together with the Event Broker colleagues (who are taking a front runner position in the entire IAS topic) to define a standard format for IAS-backed services, I'd hope that we can get around further configuration options in this implementation.

In other words: Services either adhere to our standard or won't be supported by the SDK. As we are the de-facto only library (AFAIK) in the entire SAP ecosystem to support IAS connectivity, I feel we are in a very strong position to "force" our standard onto service providers.

Nevertheless, I don't see any blockers for future extensibility of our implementation in case there are (once again) varying service binding formats.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
newtork: > I felt like spelling out the m somewhat looks better.

I agree.. though not a strong opinion 😁

cloudplatform/connectivity-oauth/src/main/resources/META-INF/services/com.sap.cloud.sdk.cloudplatform.connectivity.ServiceBindingDestinationLoader
```diff
@@ -1 +1,2 @@
 com.sap.cloud.sdk.cloudplatform.connectivity.OAuth2ServiceBindingDestinationLoader
+com.sap.cloud.sdk.cloudplatform.connectivity.IdentityAuthenticationServiceBindingDestinationLoader
```
newtork, line 2: (Minor)

Until we have a JSON syntax aligned, let's disable the auto loading.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,141 @@
...
+        this.delegateLoader = delegateLoader;
+    }
+
+    @Nonnull
+    @Override
+    public Try<HttpDestination> tryGetDestination( @Nonnull final ServiceBindingDestinationOptions options )
+    {
+        final ServiceBinding serviceBinding = options.getServiceBinding();
+        if( !isIdentityAuthenticationServiceBinding(serviceBinding) ) {
+            return Try.failure(new DestinationNotFoundException("The bound service is not backed by the IAS service."));
```
newtork, line 49: (Minor)

Please use a static exception, to avoid expensive stacktrace generation. (and Dynatrace users complain)

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,318 @@
...
+
+        final ServiceBindingDestinationOptions.Builder optionsBuilder =
+            ServiceBindingDestinationOptions
+                .forService(ServiceIdentifier.of("identity"))
+                .onBehalfOf(options.getOnBehalfOf())
+                .withOption(BtpServiceOptions.IasOptions.withTargetUri(endpoint.uri));
+
+        if( !endpoint.requiresTokenForTechnicalAccess
+            && options.getOnBehalfOf() != OnBehalfOf.NAMED_USER_CURRENT_TENANT ) {
+            optionsBuilder.withOption(BtpServiceOptions.IasOptions.withMutualTlsOnly());
```
Johannes-Schneider, line 87: I'm wondering whether this is correct for the "current Tenant is subscriber" case... Probably not, right?

So it should be 

```suggestion
        if( !endpoint.requiresTokenForTechnicalAccess
            && options.getOnBehalfOf() == OnBehalfOf.TECHNICAL_USER_PROVIDER ) {
            optionsBuilder.withOption(BtpServiceOptions.IasOptions.withMutualTlsOnly());
```

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        if( bindingView.endpoints.size() > 1 ) {
+            log
+                .warn(
+                    "The IAS-based service binding for service '{}' contains multiple HTTP endpoints. Only the first one will be used.",
+                    bindingView.serviceIdentifier);
+        }
+
+        return bindingView.endpoints.get(0);
```
Johannes-Schneider, line 109: **Discussion**

Should we already introduce some API for customers to select the endpoint the want? Or do we wait until we actually encounter a feature request?

newtork: (Comment)

CAP reminded us; first element may not necessarily be consistent in JSON format. 
Even in Java-case having `LinkedHashMap` (maybe) it may not be consistent in service binding side.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
Johannes-Schneider: As discussed, this implementation has been shifted towards the `OAuth2PropertySupplier` as this class (the `IdentityAuthenticationServiceBindingDestinationLoader`) cannot determine whether the current tenant is the provider.

Also see the changes for within the `IasOptions` class for reference.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -200,19 +201,47 @@ private void attachIasCommunicationOptions( @Nonnull final OAuth2Options.Builder
...
                 return false;
             }
 
-            return true;
+            final Boolean noTokenRequired =
+                options
+                    .getOption(BtpServiceOptions.IasOptions.NoTokenForTechnicalProviderAccess.class)
+                    .getOrElse(false);
+            if( !noTokenRequired ) {
+                // no token flag is NOT set --> so we need a token
```
newtork, line 217: args :D

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,311 @@
...
+                return false;
+            }
+
+            try {
+                return "identity".equalsIgnoreCase(authenticationService.getString("service-label"));
+            }
+            catch( final ValueCastException e ) {
+                // the `service-label` entry is not a string
+                return false;
+            }
```
newtork, line 160: (Minor)

Please add a trace/debug log statement.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -200,19 +201,47 @@ private void attachIasCommunicationOptions( @Nonnull final OAuth2Options.Builder
             }
         }
 
-        private boolean mTLSOnlyIsSupported()
+        private boolean skipTokenRetrieval()
         {
             final OnBehalfOf behalf = options.getOnBehalfOf();
             if( behalf == OnBehalfOf.NAMED_USER_CURRENT_TENANT ) {
```
MatKuhr, line 207: it's a bit confusing to have this split from the other behalf checks. How about transforming this to the new fancy switch statement? Then we can also omit implementing the `Unhandled OnBehalfOf ..` case...

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,311 @@
...
+        return bindingView.endpoints.get(0);
+    }
+
+    @Value
+    private static class IasServiceBindingView
+    {
+        @Nonnull
+        ServiceIdentifier serviceIdentifier;
+        @Nullable
+        String applicationName;
```
MatKuhr, line 111: These seem to be unused, what do you think we need them for in the future? Or are those leftovers?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,311 @@
...
+            final TypedMapView authenticationService = getAuthenticationServiceOrNull(credentials);
+
+            if( authenticationService == null || !isBackedByIas(authenticationService) ) {
+                return null;
+            }
+
+            final String applicationName = getApplicationNameOrNull(authenticationService);
+            final List<HttpEndpointEntry> endpoints = getHttpEndpointsOrThrow(credentials);
+            return new IasServiceBindingView(
+                serviceBinding.getServiceIdentifier().orElse(NULL_IDENTIFIER),
```
MatKuhr, line 128: Any particular reason why we should allow bindings without an identifier?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        private static TypedMapView getAuthenticationServiceOrNull( @Nonnull final TypedMapView credentials )
+        {
+            if( !credentials.containsKey("authentication-service") ) {
+                return null;
+            }
+
+            try {
+                return credentials.getMapView("authentication-service");
```
MatKuhr, line 140: Feels a bit bad to have around 200 LoC just for loading a few JSON fields.. Personally I'd probably wrap some of this into a `getOrDefault(str, default)` method, but not sure how much better that looks..

MatKuhr: Or we directly replace this with gson or jackson deserialization maybe?

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoaderTest.java
```diff
@@ -0,0 +1,253 @@
...
+import com.sap.cloud.environment.servicebinding.api.DefaultServiceBinding;
+import com.sap.cloud.environment.servicebinding.api.DefaultServiceBindingAccessor;
+import com.sap.cloud.environment.servicebinding.api.ServiceBinding;
+import com.sap.cloud.environment.servicebinding.api.ServiceIdentifier;
+import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationAccessException;
+import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationNotFoundException;
+
+import io.vavr.control.Try;
+
+@Isolated( "because the tests manipulate the global default ServiceBindingAccessor" )
```
MatKuhr, line 33: 👌🏼

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoaderTest.java
```diff
@@ -0,0 +1,253 @@
...
+                entry("endpoints.foo.always-requires-token", false));
+
+        final ServiceBindingDestinationLoader delegate = mockDelegateLoader(delegateOptions -> {
+            assertThat(delegateOptions.getServiceBinding()).isSameAs(IDENTITY_BINDING);
+            assertThat(delegateOptions.getOnBehalfOf()).isEqualTo(OnBehalfOf.NAMED_USER_CURRENT_TENANT);
+            assertThat(delegateOptions.getOption(BtpServiceOptions.IasOptions.IasTargetUri.class))
+                .containsExactly(URI.create("https://foo.uri"));
+            assertThat(delegateOptions.getOption(BtpServiceOptions.IasOptions.NoTokenForTechnicalProviderAccess.class))
+                .contains(true);
+            assertThat(delegateOptions.getOption(BtpServiceOptions.IasOptions.IasCommunicationOptions.class)).isEmpty();
```
MatKuhr, line 125: these assertions are the same as for the case above, right? If so, what does this test really assert, other than the behalf not impacting the assertions? Maybe we can reduce the code duplication between these test cases still a bit to make it clearer what the differences are between them?

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
```diff
@@ -142,29 +142,21 @@ public static OptionsEnhancer<?> withTargetUri( @Nonnull final URI targetUri )
...
-         * <p>
-         * <b>Caution:</b> This option cannot be combined with {@link OnBehalfOf#NAMED_USER_CURRENT_TENANT}.
+         * Creates an instance of {@link NoTokenForTechnicalProviderAccess}.
          *
-         * @return An instance of {@link OptionsEnhancer} that will lead to mTLS authentication only.
+         * @return A new {@link NoTokenForTechnicalProviderAccess} instance.
          */
         @Nonnull
-        public static OptionsEnhancer<?> withMTLSAuthenticationOnly()
+        public static OptionsEnhancer<?> withMutualTlsForTechnicalProviderAuthenticationOnly()
```
MatKuhr, line 150: I feel that even with such a lengthy name it isn't super clear. One could also read this as "only a technical user of the provider account needs to do mTLS" or "only the authentication part for the provider case (not the OAuth flow) requires mTLS"... 

How about:
```suggestion
        public static OptionsEnhancer<?> withoutTokenForBehalfTechnicalUserProvider()
```

or something similar?

MatKuhr: I feel the code would become a bit easer to read if we align the naming in the different places a bit more. Of course, the semantic is ever so slightly different in the different places, but still 😄

cloudplatform/connectivity-oauth/src/main/resources/META-INF/services/com.sap.cloud.sdk.cloudplatform.connectivity.ServiceBindingDestinationLoader
Johannes-Schneider: Initially, Alex and I discussed whether it would be a good idea to auto load this new implementation into all Cloud SDK applications without having a (somewhat) aligned standard for the Service Binding format.

However, as we further refined how we think those service bindings should look like, I'd agree that it doesn't really hurt to already load this new implementation.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
Johannes-Schneider: I'd argue that it doesn't provide any value if we were to log the exception here since it is a valid outcome that the `authentication-service.service-label` is not a String (even though the chances are pretty minimal).

Unfortunately, we don't have any API within the SBL to check whether a field is of a certain type so that we have to rely on throwing and catching exception for the moment.
Personally, I'd be happy to introduce such new API, but I don't know if the benefit justifies the effort.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
Johannes-Schneider: > Feels a bit bad to have around 200 LoC just for loading a few JSON fields

Agreed. Let me see what I can do about it.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/IdentityAuthenticationServiceBindingDestinationLoader.java
Johannes-Schneider: Actually just left-overs → Removed.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
Johannes-Schneider: mhmm... What do you think about `withoutTokenForTechnicalProviderUser`?

MatKuhr: sounds good 👍🏻

MatKuhr (approved):
lgtm 👍🏻

PR #332, opened 2024-02-27
feat: [Destinations] Support Zero Trust Identity Service

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#376.

This PR introduces support for the Zero Trust Identity Service and the related credential type `X509_ATTESTED` in IAS service bindings.

### API Design

- [ ] Dedicated Module
  - (+) Can be referenced by `connectivity-oauth` or other potentially modules as `<optional>` dependency
  - (+) Using `<optional>` means no new dedicated public API / class-loading is needed
  - (+) Doesn't come in by default, so doesn't "pollute" the dependency tree of apps that don't need it
  - (+) We can include the linux-specific dependency as users who need this module will almost definitely need it
  - (-) It may be tempting for future developers to simply add this dependency in other places of the SDK, without the optional tag
    - I tried adding an enforcer rule but the enforcer plugin does not seem capable enough to cover this. I tried using `banTransitiveDependencies` but that wasn't working (at all, really, it bans any transitive dependency and configuring it not to is at a pain at best, I can't get it to work reliably)
- [ ] `KeyStore ZeroTrustIdentityService.getInstance().getOrCreateKeyStore()` 
  - (+) Easy to use
  - (+) Doesn't require Spiffe library in public API
  - (-) It's hard for the user to know when the KeyStore will expire
- [ ] Fallback to file system for local testing by customizing the service binding
  - (+) Based on fixed properties -> SDK doesn't have to do any guessing / fallback
  - (-) Properties are completely "made up", users have to mock the binding locally
  - (-) Mocking bindings locally is super clunky

E2E Test PR: [link](https://github.tools.sap/cloudsdk/cloud-sdk-java-tests/pull/46)
Successful E2E test run: [link](https://sdk-v2.cpe.c.eu-de-1.cloud.sap/blue/organizations/jenkins/end-to-end-tests%2Fv5-scp-cf-spring-ias/detail/v5-scp-cf-spring-ias/299/pipeline/627/). 
Documentation: [PR](https://github.com/SAP/cloud-sdk/pull/1756)

### Feature scope:
 
- [x] Load certificates via SPIFFE workload API
- [x] Fallback to file system access for local testing
- [x] Integration into OAuth flows with credential type `X509_ATTESTED`

## Definition of Done

- [x]  Error handling created / updated & covered by the tests above
- [x] Documentation updated
- [x] Release notes updated


cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityDestinationEnhancer.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        this.service = service;
+    }
+
+    @Nonnull
+    @Override
+    public Try<HttpDestination> tryGetDestination( @Nonnull ServiceBindingDestinationOptions options )
+    {
+        if( !ZTIS_IDENTIFIER.equals(options.getServiceBinding().getServiceIdentifier().orElse(null)) ) {
+            return Try.failure(new DestinationNotFoundException());
```
newtork, line 36: (Minor)

Please use static exceptions to reduce unnecessary-exception count.

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -0,0 +1,103 @@
...
+    private static final Duration DEFAULT_SOCKET_TIMEOUT = Duration.ofSeconds(10);
+    private final Lazy<X509Source> source;
+    private KeyStoreCache keyStoreCache = null;
+
+    ZeroTrustIdentityService()
+    {
+        source = Lazy.of(this::initX509Source);
+    }
+
+    ZeroTrustIdentityService( Lazy<X509Source> source )
```
newtork, line 44: (Minor)

Can this be `Supplier<X509Source>` instead? `Lazy` implements this interface.

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -0,0 +1,103 @@
...
+
+    synchronized KeyStore getOrCreateKeyStore()
+    {
+        final X509Svid svid = getX509Svid();
+        final KeyStore.Entry privateKeyEntry = new PrivateKeyEntry(svid.getPrivateKey(), svid.getChainArray());
+
+        if( keyStoreCache != null && !svid.getLeaf().getNotBefore().after(keyStoreCache.lastUpdated()) ) {
+            return keyStoreCache.keyStore();
+        }
+        final KeyStore keyStore = loadKeyStore(svid);
```
newtork, line 84: (Minor)

I think I would rather have this method throw generic `Exception` than having wrapped everything into `RuntimeException`

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -0,0 +1,267 @@
...
+ * Implementation to access X.509 certificates provided by the Zero Trust Identity Service (ZTIS).
+ *
+ * @since 5.7.0
+ */
+@Beta
+@Slf4j
+@RequiredArgsConstructor( access = AccessLevel.PRIVATE )
+public class ZeroTrustIdentityService
+{
+    static final ServiceIdentifier ZTIS_IDENTIFIER = ServiceIdentifier.of("zero-trust-identity");
```
newtork, line 51: (Question)

Just wondering, this identifier is set, right? It's unlikely there will be a renaming done to this service(?)

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
```diff
@@ -132,11 +137,43 @@ protected List<String> getOAuthPropertyPath()
     ClientIdentity getCertificateIdentity()
     {
         final String clientid = getOAuthCredentialOrThrow(String.class, "clientid");
+
+        final String exactCredentialType = getOAuthCredentialOrThrow(String.class, "credential-type");
```
newtork, line 141: (Question)

Is this a potential behavior change? AFAIK previous service bindings didn't "require" a `credential-type` field. Now it potentially throws an exception. Though I'm not sure on the impact. Presumably the field is/was always there. However, I don't know how likely mocked / at-runtime created service bindings are.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
```diff
@@ -132,11 +137,43 @@ protected List<String> getOAuthPropertyPath()
...
+            getClass()
+                .getClassLoader()
+                .loadClass("com.sap.cloud.sdk.cloudplatform.connectivity.ZeroTrustIdentityService");
+        }
+        catch( final ClassNotFoundException e ) {
+            throw new CloudPlatformException(
+                "Failed to load implementation for credential type X509_ATTESTED. Please ensure the 'connectivity-ztis' module is present.",
+                e);
+        }
+        final ZeroTrustIdentityService ztis = ZeroTrustIdentityService.getInstance();
```
newtork, line 165: (Minor)

Not sure whether I like the same classpath interaction with a class from an `optional` dependency. The chance is high someone assumes default class availability from reading the code. However, there the class-loader check at the top, so maybe it's fine :)

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -97,7 +99,33 @@ class OAuth2Service
...
+                .name("oauth-destination-ztis-" + identity.getId().hashCode())
+                .keyStore(((ZtisClientIdentity) identity).keyStore())
+                .build();
+        final CloseableHttpClient httpClient;
+        try {
+            httpClient = (CloseableHttpClient) HttpClientAccessor.getHttpClient(destination);
+        }
+        catch( final ClassCastException e ) {
+            throw new DestinationAccessException(
+                "For the X509_ATTESTED credential type the 'HttpClientAccessor' must return instances of 'CloseableHttpClient'.",
```
newtork, line 124: (Question)

Is this cast-/catch-handling really necessary? I.e. i'm wondering in what circumstances is returned `HttpClient` not `CloseableHttpClient`. 

Alternatively, could you reduce LOCs on this method? ^^

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/SecurityLibWorkarounds.java
```diff
@@ -17,11 +27,50 @@ private SecurityLibWorkarounds()
...
+        if( rawType.equals(X509_GENERATED) || rawType.equals(X509_ATTESTED) ) {
+            // these particular credential types are only supported by the Security Client Lib > 3.3.5
             return CredentialType.X509;
         }
 
         return CredentialType.from(rawType);
     }
-}
+
+    record ZtisClientIdentity( @Nonnull String id, @Nonnull KeyStore keyStore) implements ClientIdentity {
```
newtork, line 38: (Comment)

Args.. the formatting is all over the place with records :D

pom.xml
```diff
@@ -730,6 +730,7 @@
 						<failOnWarning>true</failOnWarning>
 						<failBuild>true</failBuild>
 						<ignoreNonCompile>false</ignoreNonCompile>
+						<ignoreUnusedRuntime>true</ignoreUnusedRuntime>
```
newtork, line 733: (Question)

Is this because of the optional dependency? :|

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -247,6 +248,9 @@ private void attachClientKeyStore( @Nonnull final OAuth2Options.Builder optionsB
         private KeyStore getClientKeyStore()
         {
             final ClientIdentity clientIdentity = getClientIdentity();
+            if( clientIdentity instanceof ZtisClientIdentity ) {
+                return ((ZtisClientIdentity) clientIdentity).keyStore();
```
newtork, line 252: (Question)

> It's hard for the user to know when the KeyStore will expire

This means `OAuth2ServiceBindingDestinationLoader#tryGetDestination(...)` will result in a _service binding_ with a potentially (soon to be outdated) `KeyStore`? 

Do we suggest users to never cache/store the object, e.g. not beyond request scope?

MatKuhr: Correct. I felt that this is consistent with our other destinations. For both Destinations both from the Destination Service and Service Bindings or if you build a destination manually. We recommend to load/create them per-use.

One alternative would be to implement a custom KeyStore that internally refreshes on-demand, and that didn't sound like a good alternative to go for..

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
MatKuhr: Good catch, yes I think you are right. Will fix!

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
MatKuhr: Yep at least in this place I hope with the explicit comment as well it is clear. But in any case, as long as one only adds usages of `ZeroTrustIdentityService` when it's actually needed and not prematurely it should be fine.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
MatKuhr: I figured we have no guarantee that we will get `CloseableHttpClient` when the user has overridden our default factory. Low probability probably, but I figured no harm in throwing an explicit error message.

I'll inline the return, saving 2 LoC 😄

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/SecurityLibWorkarounds.java
MatKuhr: true, I didn't even realize the indentation was off. We don't gain much from using a getter here, I'll consider switching to Lombok value.

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
MatKuhr: Yep, it won't change. I figured not to add a constant to `ServiceIdentifier` for now because we don't need users to use it and also we currently don't have an implementation of our `ServiceBindingDestinationLoader` that would handle the binding.

I had this in a [previous version of the PR](https://github.com/SAP/cloud-sdk-java/pull/332/commits/6278ca3b15e9b969cd42424e2e6efd5ddf61cbca#diff-9826d7f7fc1c4959ed264a286480e28d4fe9288801ae5c3009b5f13e6394564dL17), just to explore the API idea, but removed it since I felt it was a bit too clunky and not necessarily needed.

pom.xml
MatKuhr: Not quite, instead it's about runtime dependencies in general. I feel like the rule doesn't make much sense for those, the whole point is that they are unused but needed 😄 

For this PR, for example [this runtime dependency](https://github.com/SAP/cloud-sdk-java/pull/332/files#diff-bd55831c37dbc9713654078124d9b116f4003cf72dea4da34ad4f1660601f0f6R88).

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
MatKuhr: Fixed ✅

Right now this would not have been an issue, because if `credential-type` is not defined the SDK defaults to `binding-secret, so `getCertificateIdentity` would have never run in that case. Still, changed it to option.

pom.xml
newtork: Strange that this didn't come up earlier. But I agree

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -117,21 +115,24 @@ X509Source initX509Source()
         try {
             return DefaultX509Source.newSource(x509SourceOptions);
         }
-        catch( final SocketEndpointAddressException | X509SourceException e ) {
-            final String msg =
-                """
-                    Failed to initialize the Zero Trust Identity Service X509 source. \
-                    Double-check that the SPIRE agent is running and the io.spiffe dependencies required for your operating system are present. \
-                    For Cloud Foundry, the agent is typically provided by the zero_trust_sidecar_buildpack and the required dependencies are io.spiffe:grpc-netty-linux. \
-                    """;
```
newtork: (Question)

Reason for change, was you felt like this is a too-detailed exception-message. Prone to be outdated?
Probably linking the documentation page in the exception message below is good replacement(?)

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -117,21 +115,24 @@ X509Source initX509Source()
...
+            throw new CloudPlatformException("Failed to load the certificate from the default unix socket.", e);
         }
     }
 
     @Nonnull
     X509Svid getX509Svid()
     {
-        return source.get().getX509Svid();
+        try {
+            return source.get().getX509Svid();
```
newtork, line 127: (Minor)

I'm not sure whether a null check would be necessary here.
If the implementation of `source` is our `FileSystemX509Source` then it's fine: non-null.

But if it is `DefaultX509Source.newSource(x509SourceOptions)` then I can't tell for sure. 🤔

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
MatKuhr: Yeah, it was not up to date anymore with the fact that the file system is also an option. So too much details for an exception msg, thus linking the docs instead

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
newtork: For source, sure, but I'm concerned about `#getX509Svid()`

MatKuhr: The get just comes from it being a `Lazy`, so either an exception is thrown or it's non-null 😉

newtork (approved):
Looks LGTM to me.

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
MatKuhr: Finally got what you meant, added a null check as discussed 👍🏻

PR #337, opened 2024-02-29
fix: DwC Auth Token not available

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#424.

I think we just forgot to implement this earlier. This change makes our `AuthTokenAccessor` actually return a token in the DwC case. Without this change, one always gets an exception (unless using CAP).

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-dwc/pom.xml
+++ b/cloudplatform/connectivity-dwc/pom.xml
@@ -70,6 +70,10 @@
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.auth0</groupId>
+			<artifactId>java-jwt</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>jcl-over-slf4j</artifactId>
 			<scope>runtime</scope>
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -35,6 +35,9 @@ public class DwcHeaderUtils
      * The name of the header that contains the Deploy with Confidence scopes information.
      */
     public static final String DWC_SUBDOMAIN_HEADER = "dwc-subdomain";
+    /**
+     * The name of the header that contains the Deploy with Confidence JWT token.
+     */
     public static final String DWC_JWT_HEADER = "dwc-jwt";
 
     /**
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -96,9 +96,10 @@ public class DwcHeaderUtils
      * @return The value of the {@link #DWC_JWT_HEADER} header.
      * @throws DwcHeaderNotFoundException
      *             if the header was not found.
+     * @since 5.6.0
      */
     @Nonnull
-    public static String getDwcTokenOrThrow()
+    public static String getDwcJwtOrThrow()
     {
         return getNonEmptyDwcHeaderValue(DWC_JWT_HEADER);
     }
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -35,6 +35,7 @@ public class DwcHeaderUtils
      * The name of the header that contains the Deploy with Confidence scopes information.
      */
     public static final String DWC_SUBDOMAIN_HEADER = "dwc-subdomain";
+    public static final String DWC_JWT_HEADER = "dwc-jwt";
 
     /**
      * This method fetches the value of the {@link #DWC_TENANT_HEADER} header or throws an
@@ -83,6 +84,20 @@ public class DwcHeaderUtils
             throw new DwcHeaderNotFoundException("Header value of " + DWC_USER_HEADER + " has no logon name.");
         }
         return logonName;
+    }
+
+    /**
+     * This method fetches the value of the {@link #DWC_JWT_HEADER} header or throws an
+     * {@link DwcHeaderNotFoundException} if the header was not found.
+     *
+     * @return The value of the {@link #DWC_JWT_HEADER} header.
+     * @throws DwcHeaderNotFoundException
+     *             if the header was not found.
+     */
+    @Nonnull
+    public static String getDwcTokenOrThrow()
+    {
+        return getNonEmptyDwcHeaderValue(DWC_JWT_HEADER);
     }
 
     @Nonnull
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacade.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacade.java
@@ -1,0 +1,41 @@
+package com.sap.cloud.sdk.cloudplatform.security;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import com.auth0.jwt.JWT;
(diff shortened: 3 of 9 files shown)


cloudplatform/security/src/main/resources/META-INF/services/com.sap.cloud.sdk.cloudplatform.security.secret.SecretStoreFacade
```diff
@@ -1 +0,0 @@
-com.sap.cloud.sdk.cloudplatform.security.secret.ScpCfSecretStoreFacade
```
MatKuhr, line 1: driveby cleanup of dead code

cloudplatform/security/src/main/resources/META-INF/services/com.sap.cloud.sdk.cloudplatform.security.AuthTokenFacade
```diff
@@ -1 +0,0 @@
-com.sap.cloud.sdk.cloudplatform.security.DefaultAuthTokenFacade
```
MatKuhr, line 1: removing this should be fine because the Accessor uses this facade regardless by default (without classloading)

CharlesDuboisSAP: It doesn't if the `DwcAuthTokenFacade` is class loaded

CharlesDuboisSAP: It doesn't if the `DwcAuthTokenFacade` is class loaded

CharlesDuboisSAP (commented):
Should this be E2E tested?

cloudplatform/connectivity-dwc/pom.xml
```diff
@@ -69,6 +69,10 @@
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.auth0</groupId>
+			<artifactId>java-jwt</artifactId>
+		</dependency>
```
newtork, line 75: (Question)

This is not "test" scoped, because it comes in "compile" transitively, right?

MatKuhr: not transitive, otherwise it wouldn't be (allowed to be) listed here. Usage is [here](https://github.com/SAP/cloud-sdk-java/pull/337/files#diff-5295696d29a6734eb094fcad0b32195dadfc42918bcd7c3e153e2a05f219b653R6).

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacade.java
```diff
@@ -0,0 +1,44 @@
...
+    @Nonnull
+    @Override
+    public Try<AuthToken> tryGetCurrentToken()
+    {
+        @Nullable
+        final ThreadContext currentContext = ThreadContextAccessor.getCurrentContextOrNull();
+        if( currentContext != null && currentContext.containsProperty(AUTH_TOKEN) ) {
+            return currentContext.getPropertyValue(AUTH_TOKEN);
+        }
+        return Try.of(DwcAuthTokenFacade::extractAuthTokenFromDwcHeaders);
```
newtork, line 30: (Minor)

Please note: not storing the potential exception in thread context will lead to multiple/many exception instantiations (2x per `tryGetCurrentToken`).

MatKuhr (commented):
currently drafting an E2E test [here](https://github.tools.sap/cloudsdk/cloud-sdk-java-tests/pull/44) based on [this test code](https://github.tools.sap/cloudsdk/dwc-cap-orbits/commit/ac9724b6c96a261bcee6fa8c11eb014254990a1b).

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacade.java
newtork: I'm not sure we need to address this right here/now.

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
```diff
@@ -85,6 +89,20 @@ public static String getDwcPrincipalIdOrThrow()
...
     }
 
+    /**
+     * This method fetches the value of the {@link #DWC_JWT_HEADER} header or throws an
+     * {@link DwcHeaderNotFoundException} if the header was not found.
+     *
+     * @return The value of the {@link #DWC_JWT_HEADER} header.
+     * @throws DwcHeaderNotFoundException
+     *             if the header was not found.
+     */
```
Johannes-Schneider, line 99: [minor]

```suggestion
     * @since 5.6.0
     */
```

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
```diff
@@ -85,6 +89,20 @@ public static String getDwcPrincipalIdOrThrow()
...
+    @Nonnull
+    public static String getDwcTokenOrThrow()
```
Johannes-Schneider, line 101: [minor]

For consistency...?

```suggestion
    public static String getDwcJwtOrThrow()
```

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacade.java
```diff
@@ -0,0 +1,44 @@
...
+import com.sap.cloud.sdk.cloudplatform.DwcHeaderUtils;
+import com.sap.cloud.sdk.cloudplatform.security.exception.AuthTokenAccessException;
+import com.sap.cloud.sdk.cloudplatform.thread.ThreadContext;
+import com.sap.cloud.sdk.cloudplatform.thread.ThreadContextAccessor;
+
+import io.vavr.control.Try;
+
+/**
+ * Auth token facade for Deploy with Confidence (DwC) environment.
+ */
```
Johannes-Schneider, line 16: [minor]

```suggestion
  * @since 5.6.0
  */
```

cloudplatform/connectivity-dwc/src/test/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacadeTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.vavr.control.Try;
+
+class DwcAuthTokenFacadeTest
+{
+    @Test
+    void testFacadeIsPickedUpAutomatically()
+    {
+        assertThat(AuthTokenAccessor.getAuthTokenFacade()).isInstanceOf(DwcAuthTokenFacade.class);
+    }
```
Johannes-Schneider, line 27: 👍

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacade.java
```diff
@@ -0,0 +1,44 @@
...
+            return currentContext.getPropertyValue(AUTH_TOKEN);
+        }
+        return Try.of(DwcAuthTokenFacade::extractAuthTokenFromDwcHeaders);
+    }
+
+    @Nonnull
+    private static AuthToken extractAuthTokenFromDwcHeaders()
+    {
+        try {
+            final String token = DwcHeaderUtils.getDwcTokenOrThrow();
```
Johannes-Schneider, line 37: [question]

Can you remind me what we decided regarding "DwC header injection" attacks (i.e. someone somehow calls the app and injects their own `dwc-*` headers)?
I know we discussed this topic a while back but cannot quite recall what we decided.

MatKuhr: we figured not an issue for various reasons, including for example:
- the dwc dependencies don't come in automatically and it's decently unlikely that they can come transitively outside of DwC
- even if they are present, the multi tenancy would not work at all without the dwc headers, so any multi-tenant app should detect this during development

PR #342, opened 2024-03-05
[House keeping] Fix PMD warnings

newtork:
- [x] Fix warnings
- [x] Reduce thresholds


cloudplatform/cloudplatform-core/src/main/java/com/sap/cloud/sdk/cloudplatform/thread/ThreadContextExecutor.java
```diff
@@ -274,6 +274,7 @@ public void execute( @Nonnull final Executable executable )
     }
 
     @Nullable
+    @SuppressWarnings( "PMD.SignatureDeclareThrowsException" )
     private <T> T call( @Nonnull final Callable<T> callable )
         throws Exception
```
CharlesDuboisSAP, line 279: Not sure about this one. Otherwise the PR LGTM

newtork: The method legitimately throws `Exception` because this is what's coming from provided `Callable#call`.

I checked whether we could catch it eagerly, but this would result in a bigger refactoring. Therefore I only added a comment to the suppression annotation.

CharlesDuboisSAP (approved):

PR #346, opened 2024-03-06
Add implementation reference to interface parameter declaration in JavaDoc

newtork:
Related
https://github.com/orgs/SAP/projects/62/views/5?pane=issue&itemId=55522157

Not sure whether this brings any benefit. But the user was unable to locate the implementation classes by themselves.

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
@@ -61,6 +61,8 @@ public class OAuth2DestinationBuilder
          * @param behalf
          *            The OAuth2 token representative.
          * @return The same builder instance.
+         * @see com.sap.cloud.security.config.ClientCertificate
+         * @see com.sap.cloud.security.config.ClientCredentials
          * @since 4.10.0
          */
         @Nonnull


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
```diff
@@ -61,6 +61,8 @@ public interface BuilderWithTokenEndpoint
          * @param behalf
          *            The OAuth2 token representative.
          * @return The same builder instance.
+         * @see com.sap.cloud.security.config.ClientCertificate
+         * @see com.sap.cloud.security.config.ClientCredentials
          * @since 4.10.0
```
CharlesDuboisSAP, line 66: I would rather add:
```java
         * @param clientIdentity
         *            The OAuth2 Client Identity. Can be a {@link ClientCredentials}
```

CharlesDuboisSAP (dismissed):

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2DestinationBuilder.java
MatKuhr: if we do that we should mention both implementations I'd think.. But personally I'm also a fan of the `@see` to point to related stuff, we can also do both 😉

PR #347, opened 2024-03-06
feat: Allow access to failed batch request

CharlesDuboisSAP:
## Summary

This PR addresses the issue raised in the backlog item [#426](https://github.com/SAP/cloud-sdk-java-backlog/issues/426). It introduces changes to allow users to access responses of failed "sub-requests" in a batch request. This is achieved by mapping the batch request content to the response via the `Content-ID`.

## Changes

The changes in this PR are as follows:

1. The `requireHealthyResponse` method in the `ODataHealthyResponseValidator` class is modified. It now tries to extract the `Content-ID` from the HTTP response and uses it to create an `ODataResponseException`.

2. The field `request` of `ODataException` now stores the specific failed OData batch changeset.

3. `testBatchWithErrorInChangeset()` checks that `exception.getRequest()` now always return the failed changeset.

## Instructions for Reviewers

Please review the changes in the context of the batch request handling. Pay special attention to the error handling and the new test case. Please also verify if the `Content-ID` is correctly extracted from the HTTP response and if the `request` field is correctly set in the `ODataException`.


Johannes-Schneider (approved):
LGTM

Issue #344, opened 2024-03-05
Get Content-ID Response Header from A Batch Response In A Changeset


SpaceCondor:
**Is your feature request related to a problem? Please describe.**
It doesn't seem to be possible to get Content-ID (or other arbitrary headers) of a batch response. For example, if I create a changeset like this:

```
BatchRequestBuilder batchRequestBuilder = new BatchRequestBuilder("/v1/");
        batchRequestBuilder.addChangeset(createDocument1, createDocument2, createDocument3));

BatchResponse batchResponse = batchRequestBuilder.execute(destination);
```
I can see the data in the request/response log like the following:

```
[Thu Feb 29 22:06:11 2024] [10.0.0.2] [pid=4152] [Request] "POST /v1/$batch HTTP/1.1"
OData-Version:4.0
Content-Type:multipart/mixed;boundary=batch_17d7a8c3-d3c7-42d4-b6a6-06ef1d975008
Content-Length:2173
Host:xxxxxxxxx.com:50000
Connection:Keep-Alive
User-Agent:Apache-HttpClient/4.5.14 (Java/21.0.2)
Cookie:**************************************************************************
Accept-Encoding:gzip,deflate
X-Forwarded-Proto:https

--batch_17d7a8c3-d3c7-42d4-b6a6-06ef1d975008
Content-Type: multipart/mixed;boundary=changeset_d3c4fa68-582a-46fd-8691-d24272625cda

--changeset_d3c4fa68-582a-46fd-8691-d24272625cda
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 1

POST Orders HTTP/1.1
Accept: application/json
Content-Type: application/json

{"@odata.type":"#SAPB1.Document"..........}

--changeset_d3c4fa68-582a-46fd-8691-d24272625cda
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 2

POST Orders HTTP/1.1
Accept: application/json
Content-Type: application/json

{"@odata.type":"#SAPB1.Document"..........}

--changeset_d3c4fa68-582a-46fd-8691-d24272625cda
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 3

POST Orders HTTP/1.1
Accept: application/json
Content-Type: application/json

{"@odata.type":"#SAPB1.Document"..........}

--changeset_d3c4fa68-582a-46fd-8691-d24272625cda--

[...]

PR #348, opened 2024-03-06
Return failed batch response

SpaceCondor:
## Context
 
This is related to the [issue I opened.](https://github.com/SAP/cloud-sdk-java/issues/344)

This builds off of the PR that was just made to get the Content-ID of a failed batch response (https://github.com/SAP/cloud-sdk-java/pull/347).  I've reworked it to contained the failed batch request instead of just the Content-ID.

## Changes
(taken & modified from previous PR)

The changes in this PR are as follows:

The requireHealthyResponse method in the ODataHealthyResponseValidator class is modified. It now tries to extract the Content-ID from the HTTP response and uses it in tandem with a new Map in the ODataRequestBatch to grab the original request.

A new field failedBatchRequest is added to the ODataResponseException class. This field stores the failed request.

A new test case testFailedBatchRequestNumber is added to the ODataV4BatchRequestUnitTest class. This test case checks if the failedBatchRequest field in the ODataResponseException is correctly set.


CharlesDuboisSAP:
This is our very first open-source contribution and your efforts are appreciated.
Thank you for making this contribution!

PR #349, opened 2024-03-08
fix: [Destinations] Fix Caching Isolation of HttpClients for Destinations with PrincipalPropagation

MatKuhr:
## Context

This PR fixes an issue where `PrincipalPropagation` did not lead to user-isolated http clients.

The problem was introduced with version 5.0. Previous versions evaluated the on-premise proxy headers upon the `.asHttp()` call, adding them as static headers to the destination, and thus including them in the equals check.

### Feature scope:

- [x] fix Httpclient 4 
- [x] fix Httpclient 5
- [x] Add warnings for unexpected scenarios

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
+++ b/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
@@ -80,7 +80,7 @@ public abstract class AbstractHttpClientCache implements HttpClientCache
                 "Failed to get HttpClientCache. Falling back to creating a new http client instance."
                     + " This is unexpected and will be changed to fail instead in a future version of Cloud SDK."
                     + " Please analyze the attached stack trace and resolve the issue.";
-            log.warn(msg, maybeCache.getCause());
+            log.error(msg, maybeCache.getCause());
             return Try.ofSupplier(createHttpClient);
         }
 
@@ -91,7 +91,7 @@ public abstract class AbstractHttpClientCache implements HttpClientCache
                 "Failed to create cache key for HttpClient. Falling back to creating a new http client instance."
                     + " This is unexpected and will be changed to fail instead in a future version of Cloud SDK."
                     + " Please a the attached stack trace and resolve the issue.";
-            log.warn(msg, maybeKey.getCause());
+            log.error(msg, maybeKey.getCause());
             return Try.ofSupplier(createHttpClient);
         }
 
--- a/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
+++ b/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
@@ -79,7 +79,7 @@ public abstract class AbstractHttpClientCache implements HttpClientCache
             final String msg =
                 "Failed to get HttpClientCache. Falling back to creating a new http client instance."
                     + " This is unexpected and will be changed to fail instead in a future version of Cloud SDK."
-                    + " Analyze the attached stack trace and resolve the issue.";
+                    + " Please analyze the attached stack trace and resolve the issue.";
             log.warn(msg, maybeCache.getCause());
             return Try.ofSupplier(createHttpClient);
         }
@@ -90,7 +90,7 @@ public abstract class AbstractHttpClientCache implements HttpClientCache
             final String msg =
                 "Failed to create cache key for HttpClient. Falling back to creating a new http client instance."
                     + " This is unexpected and will be changed to fail instead in a future version of Cloud SDK."
-                    + " Analyze the attached stack trace and resolve the issue.";
+                    + " Please a the attached stack trace and resolve the issue.";
             log.warn(msg, maybeKey.getCause());
             return Try.ofSupplier(createHttpClient);
         }
--- a/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
+++ b/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
@@ -74,13 +74,24 @@ public abstract class AbstractHttpClientCache implements HttpClientCache
                 : httpClientFactory.createHttpClient();
         };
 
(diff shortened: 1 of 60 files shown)


cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpClientCache.java
```diff
@@ -76,12 +82,27 @@ protected Try<Cache<CacheKey, HttpClient>> getCache()
     @Override
     protected Try<CacheKey> getCacheKey( @Nonnull final HttpDestinationProperties destination )
     {
-        return Try.of(() -> {
-            if( DestinationUtility.requiresUserTokenExchange(destination) ) {
-                return CacheKey.ofTenantAndPrincipalOptionalIsolation().append(destination);
-            }
-            return CacheKey.ofTenantOptionalIsolation().append(destination);
-        });
+        if( !DestinationUtility.requiresUserTokenExchange(destination)
```
MatKuhr, line 85: An alternative approach would be to make `requiresUserTokenExchange` return true for `PRINCIPAL_PROPAGATION`.

However, `requiresUserTokenExchange` is also used to:
- determine if a user token is required for communicating with the destination service
  - `PRINCIPAL_PROPAGATION` destinations do not need this
- determine the cache key for the destination object
  - `PRINCIPAL_PROPAGATION` do not need to be user-isolated, only their HttpClients

For these reasons I went with the current approach of only computing the check within the http client cache

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
```diff
@@ -74,13 +74,24 @@ private Try<HttpClient> tryGetOrCreateHttpClient(
                 : httpClientFactory.createHttpClient();
         };
 
-        final Try<Cache<CacheKey, HttpClient>> maybeCache =
-            getCache().onFailure(cause -> logCachePropagationFailure("Could not get HttpClient cache.", cause));
+        final Try<Cache<CacheKey, HttpClient>> maybeCache = getCache();
+        if( maybeCache.isFailure() ) {
+            final String msg =
+                "Failed to get HttpClientCache. Falling back to creating a new http client instance."
+                    + " This is unexpected and will be changed to fail instead in a future version of Cloud SDK."
```
MatKuhr, line 81: Maybe discussion worthy...

Reasoning for the message: This case feels like legacy that shouldn't happen, but in order to not break any existing code I would leave it as warning message..

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpClientCache.java
Johannes-Schneider: I think we should convert this GH comment into a regression test for our `DestinationUtility` class.
That way, we can avoid someone stumbling over the code and thinking they found a bug (because `PRINCIPAL_PROPAGATION` of course requires a user token when talking with an actual S/4 system).

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
Johannes-Schneider: Personally, I agree with making this case VERY apparent to the user as this should definitely not happen.
I'd even go as far as saying we should log this on `error` level.

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AbstractHttpClientCache.java
```diff
@@ -74,13 +74,24 @@ private Try<HttpClient> tryGetOrCreateHttpClient(
...
         final Try<CacheKey> maybeKey = destination != null ? getCacheKey(destination) : getCacheKey();
-        maybeKey.onFailure(cause -> logCachePropagationFailure("Could not get HttpClient cache key.", cause));
 
-        if( maybeCache.isFailure() || maybeKey.isFailure() ) {
+        if( maybeKey.isFailure() ) {
+            final String msg =
+                "Failed to create cache key for HttpClient. Falling back to creating a new http client instance."
+                    + " This is unexpected and will be changed to fail instead in a future version of Cloud SDK."
+                    + " Please a the attached stack trace and resolve the issue.";
+            log.warn(msg, maybeKey.getCause());
```
Johannes-Schneider, line 94: Same here: This case is so unexpected and seems to indicate such a severe issue so that I would log it on `error` level.

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpClientCache.java
```diff
@@ -76,12 +82,27 @@ protected Try<Cache<CacheKey, HttpClient>> getCache()
...
+            && destination.getAuthenticationType() != AuthenticationType.PRINCIPAL_PROPAGATION ) {
```
Johannes-Schneider, line 86: Personally, I think we should return `true` for `DestinationIUtility.requiresUserTokenExchange(PRINCIPAL_PROPAGATION)` and rather code the exceptional cases (i.e. destination service communication and determining the cache key) explicitly.
That somehow feels cleaner to me.

MattKuhr: I see the point in this. The reason I didn't go for this so far is that I think the utility class was primarily created for the destination service related logic. Second, it is technically not correct to say "requiresUserTokenExchange" for any destination, because token forwarding works and is the default.

I'll take the opportunity to try and come up with a better alternative that is a bit clearer overall

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpClientCache.java
MatKuhr: Done ✅

Johannes-Schneider (approved):

PR #350, opened 2024-03-08
[Draft] [OData] Expose Content-Id of failing changeset response

newtork:
This approach...

* **does not consume** the lazily evaluated `HttpResponse` from the batch response. All lazy behavior properties of `ODataRequestResultMultipartGeneric` are kept maintained.
* **no longer throws away** the batch meta information but instead combines it with the original payload to new class `Entry(String meta, String payload)` whereas the meta information like `Content-ID` can be extracted easily.
* All changes are not public API breaking. The affected classes are package-private.

### DIFF
--- a/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultMultipartGeneric.java
+++ b/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultMultipartGeneric.java
@@ -122,7 +122,7 @@ public class ODataRequestResultMultipartGeneric
         }
 
         final ODataRequestResultGeneric result = new ODataRequestResultGeneric(request, response);
-        result.disableBufferingHttpResponse(); // the HttpResponse is already evaluated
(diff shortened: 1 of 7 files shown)


datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultMultipartGeneric.java
```diff
@@ -122,6 +122,7 @@ public ODataRequestResultGeneric getResult( @Nonnull final ODataRequestGeneric r
         }
 
         final ODataRequestResultGeneric result = new ODataRequestResultGeneric(request, response);
+        result.disableBufferingHttpResponse(); // the HttpResponse is already evaluated
```
newtork, line 125:
```suggestion
        result.disableBufferingHttpResponse(); // the artificial HttpResponse is static, no buffer required
```

Johannes-Schneider (commented):
Changes generally LGTM - just some minor improvement points.

CharlesDuboisSAP:
Merged in https://github.com/SAP/cloud-sdk-java/pull/347

PR #356, opened 2024-03-12
Fix: Business Logging Service Binding removed from Endpoint URLs

rnewbigging:
## Context
 

SAP/cloud-sdk-java-backlog#416

Some Business Logging service endpoints contain redundant paths that get appended again when generated clients are called, creating bad paths. The fix removes paths from the returned URLs for Business Logging service, so they may be later appended without conflict. 

### Feature scope:

 
- [x] remove paths from endpoint urls for Business Logging

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
@@ -74,21 +74,7 @@ final class MultiUrlPropertySupplier<T extends OptionsEnhancer<T>> extends Defau
         URI endpointUrl = getCredential(URI.class, "endpoints", bindingKey).get();
         final boolean removePath = urlKeys.get(option).get(bindingKey).get();
         if( removePath ) {
-            final String path = endpointUrl.getPath();
-            try {
-                endpointUrl =
-                    new URI(
-                        endpointUrl.getScheme(),
-                        endpointUrl.getAuthority(),
-                        null,
-                        endpointUrl.getQuery(),
-                        endpointUrl.getFragment());
-            }
-            catch( URISyntaxException e ) {
-                throw new DestinationAccessException(
-                    "Unable to create a URI from '" + endpointUrl + "' by removing path '" + path + "'",
-                    e);
-            }
+            endpointUrl = endpointUrl.resolve("/");
         }
         return endpointUrl;
     }
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
@@ -61,11 +61,11 @@ final class MultiUrlPropertySupplier<T extends OptionsEnhancer<T>> extends Defau
         final Optional<String> maybeBindingKey = urlKeys.get(option).keySet().stream().findFirst();
         if( maybeBindingKey.isEmpty() ) {
             throw new IllegalStateException(
-                    "Found option value "
-                            + option
-                            + " for "
-                            + enhancerClass.getName()
-                            + ", but no URL key was registered for this value. Please ensure that for each possible choice a URL key is registered.");
+                "Found option value "
+                    + option
+                    + " for "
+                    + enhancerClass.getName()
+                    + ", but no URL key was registered for this value. Please ensure that for each possible choice a URL key is registered.");
         }
 
         final String bindingKey = maybeBindingKey.get();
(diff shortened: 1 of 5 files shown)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
```diff
@@ -55,17 +58,39 @@ public URI getServiceUri()
...
+                        endpointUrl.getAuthority(),
+                        null,
+                        endpointUrl.getQuery(),
+                        endpointUrl.getFragment());
+            }
+            catch( URISyntaxException e ) {
+                throw new DestinationAccessException(
+                    "Unable to create a URI from '" + endpointUrl + "' by removing path '" + path + "'",
+                    e);
+            }
```
MatKuhr, line 91: Maybe this can be simplified a bit using `endpointUrl.resolve("/")`

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
```diff
@@ -55,17 +58,39 @@ public URI getServiceUri()
...
                     + ", but no URL key was registered for this value. Please ensure that for each possible choice a URL key is registered.");
         }
+
+        final String bindingKey = maybeBindingKey.get();
         log.debug("Option {} selected, using binding key {}.", option, bindingKey);
-        return getCredential(URI.class, "endpoints", bindingKey).get();
+
+        URI endpointUrl = getCredential(URI.class, "endpoints", bindingKey).get();
+        final boolean isPathless = urlKeys.get(option).get(bindingKey).get();
+        if( isPathless ) {
```
MatKuhr, line 76: Maybe a bit more clear:

```suggestion
        final boolean removePath = urlKeys.get(option).get(bindingKey).get();
        if( removePath ) {
```

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
```diff
@@ -30,12 +33,12 @@
 final class MultiUrlPropertySupplier<T extends OptionsEnhancer<T>> extends DefaultOAuth2PropertySupplier
 {
     private final Class<T> enhancerClass;
-    private final Map<OptionsEnhancer<T>, String> urlKeys;
+    private final Map<OptionsEnhancer<T>, Map<String, Supplier<Boolean>>> urlKeys;
```
MatKuhr, line 36: This makes the code a bit more complicated. Maybe worthwhile if we have further cases like this, but I'm not sure how many further services we will add to our list of services we supported by default.

An alternative might be to create another small class extending `MultiUrlPropertySupplier` and overriding `getServiceUri()`, did you consider that?

MatKuhr (commented):
Currently the new logic isn't used yet, right? With this solution that would happen in `BtpServicePropertySuppliers#72`, right??

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
```diff
@@ -114,7 +139,18 @@ static final class Builder<T extends OptionsEnhancer<T>>
...
         {
-            urlKeys.put(enhancer, urlKey);
+            return withUrlKey(enhancer, urlKey, () -> enhancerClass == BtpServiceOptions.BusinessLoggingOptions.class);
+        }
+
+        @Nonnull
+        private Builder<T> withUrlKey(
+            @Nonnull final OptionsEnhancer<T> enhancer,
+            @Nonnull final String urlKey,
+            @Nonnull final Supplier<Boolean> isPathless )
```
MatKuhr, line 149: discussed offline a bit, leaving this here before I forget it again: another generic solution alternative might be:

```suggestion
            @Nonnull final Function<URI, URI> uriTransformation )
```

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
rnewbigging: I agree I'm not super happy with the nested mapping, It makes retrieving the values a bit messy. I haven't thought about an additional class, I can test that out in another PR, though to me that seems a bit counterintuitive, since we just spent a great deal of time consolidating classes for V5 😅. But can't hurt to try.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/MultiUrlPropertySupplier.java
rnewbigging: Would this not mean then that we'd be computing the URIs on builder init rather than when retrieving via `getServiceURI()`. I'm not sure of the typical usage patterns here, if we tend to `get` the URIs soon after construction then it shouldn't be an issue, but I would assume since the original logic is in the get method that we prefer the lazy approach?

MatKuhr: The function would be invoked at the same point in time where currently `Supplier<Boolean>` is invoked, so there would not be any difference in that I think..

For most cases the function would be identity, but for the case where we want to modify the URI it could be e.g. `uri -> uri.resolve("/")`. The advantage over `Supplier<Boolean>` would be that you can do anything to the URL, e.g. appending a path or some query even maybe..

rnewbigging: Ah okay, I misunderstood how a `Function<>` was used then. Thanks for the clarification 👍

MatKuhr (approved):

PR #359, opened 2024-03-14
Improve error handling for Missing Connectivity Proxy information

MatKuhr:
## Context

This PR improves the error handling in case adding on-premise headers failed. The error message is now more expressive and prepares a future update to change the behavior to be more consistent with the rest of the SDk.

### Feature scope:
 
- [x] Increase relevant error messages to `error` log level
- [x] Add some comments to clarify some rather implicit interactions

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
@@ -224,7 +224,7 @@ public final class DefaultHttpDestination implements HttpDestination
                         .stream()
                         .filter(Objects::nonNull)
                         .map(value -> new Header(HttpHeaders.AUTHORIZATION, value))
-                        .collect(Collectors.toList());
+                        .toList();
 
                 if( headersToAdd.isEmpty() ) {
                     final String msg =
@@ -972,10 +972,10 @@ public final class DefaultHttpDestination implements HttpDestination
                 throw new IllegalArgumentException("Cannot build a HttpDestination without a URL.");
             }
 
-            // NOT using the typed property here since this would break change detection in our ScpCfDestinationLoader
-            property(DestinationProperty.TYPE.getKeyName(), DestinationType.HTTP.toString());
+            if( get(DestinationProperty.TYPE).isEmpty() ) {
+                property(DestinationProperty.TYPE, DestinationType.HTTP);
+            }
 
-            // handle proxy type == OnPremise
             if( builder.get(DestinationProperty.PROXY_TYPE).contains(ProxyType.ON_PREMISE) ) {
                 try {
                     return proxyHandler.handle(this);
@@ -986,7 +986,6 @@ public final class DefaultHttpDestination implements HttpDestination
                     log.error(msg, e);
                 }
             }
-
             return buildInternal();
         }
 
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
@@ -982,7 +982,13 @@ public final class DefaultHttpDestination implements HttpDestination
                 }
                 catch( final Exception e ) {
                     final String msg =
-                        "Unable to resolve proxy configuration for destination. This destination cannot be used for anything other than reading its properties.";
+                        """
+                            Unable to resolve proxy configuration for destination. \\
+                            This destination cannot be used for anything other than reading its properties. \\
+                            This is unexpected and will be changed to fail instead in a future version of Cloud SDK. \\
+                            Please analyze the attached stack trace and resolve the issue. \\
+                            In case only the properties of a destination should be accessed, without performing authorization flows, please use the 'getDestinationProperties'  method on 'DestinationService' instead.\\
+                            """;
                     log.error(msg, e);
                 }
             }
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
@@ -110,7 +110,7 @@ class DefaultHttpDestinationBuilderProxyHandler
      *         found.
      */
     @Nonnull
-    ServiceBinding getServiceBindingConnectivity()
+    private ServiceBinding getServiceBindingConnectivity()
     {
         final ServiceBindingAccessor serviceBindingAccessor = getServiceBindingAccessor();
 
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
@@ -7,10 +7,8 @@ package com.sap.cloud.sdk.cloudplatform.connectivity;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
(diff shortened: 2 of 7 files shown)


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -972,21 +972,26 @@ public DefaultHttpDestination build()
                 throw new IllegalArgumentException("Cannot build a HttpDestination without a URL.");
             }
 
-            // NOT using the typed property here since this would break change detection in our ScpCfDestinationLoader
-            property(DestinationProperty.TYPE.getKeyName(), DestinationType.HTTP.toString());
+            if( get(DestinationProperty.TYPE).isEmpty() ) {
+                property(DestinationProperty.TYPE, DestinationType.HTTP);
+            }
```
Johannes-Schneider, line 977: This seems to be a somewhat unexpected change. Why are we no longer making sure the type of a `HttpDestination` is actually `Http`?

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
```diff
@@ -49,14 +39,17 @@ DefaultHttpDestination handle( @Nonnull final DefaultHttpDestination.Builder bui
         throws DestinationAccessException,
             DestinationNotFoundException
     {
+        if( !builder.get(DestinationProperty.PROXY_TYPE).contains(ProxyType.ON_PREMISE) ) {
+            throw new IllegalStateException(
+                "Proxy handler was invoked unexpectedly for destination "
+                    + builder.get(DestinationProperty.NAME).getOrElse("unnamed-destination")
+                    + " which does not have proxy type 'ON_PREMISE'.");
```
Johannes-Schneider, line 46: How about we transform this exception into an warning (or maybe even error?) log statement and just return the destination without doing anything?

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
```diff
@@ -117,27 +109,32 @@ private static OnBehalfOf deriveOnBehalfOf( @Nonnull final DefaultHttpDestinatio
...
+
+        if( serviceBindings.isEmpty() ) {
+            throw new DestinationAccessException(
+                "Unable to resolve connectivity service binding. No service bindings found matching Connectivity.");
+        } else if( serviceBindings.size() > 1 ) {
+            throw new DestinationAccessException(
+                "Unable to resolve connectivity service binding. More than one service bindings found that match Connectivity.");
+        } else {
+            return serviceBindings.get(0);
         }
```
Johannes-Schneider, line 131: [minor - nitpicky]

```suggestion
        return serviceBindings.get(0);
```

cloudplatform/cloudplatform-connectivity/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandlerTest.java
```diff
@@ -61,84 +64,59 @@ public Try<HttpDestination> tryGetDestination( @Nonnull ServiceBindingDestinatio
...
-        final DefaultHttpDestination result = new DefaultHttpDestinationBuilderProxyHandler().handle(builder);
-
-        assertThat(result).isNotNull();
-        verify(destinationLoader).tryGetDestination(argThat(options -> {
-            return OnBehalfOf.TECHNICAL_USER_CURRENT_TENANT == options.getOnBehalfOf()
-                && options.getServiceBinding() == connectivityService
-                && uri.equals(options.getOption(ProxyOptions.class).get().getUri())
-                && options.getOption(ProxyOptions.class).get().getHeaders().isEmpty();
-        }));
-    }
```
Johannes-Schneider, line 127: Why has this test been removed?

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
MatKuhr: I considered this, but since this is internal API only, the only misuse would come from internal usage and we would probably miss it if its just a log entry

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
MatKuhr: The only case where this would not be HTTP is if the user explicitly is building a `DefaultHttpDestination` for a destination with some other type than HTTP, e.g. MAIL. We could also consider to fail in that case, but since the type isn't really used for anything and I figured it's not super needed...

cloudplatform/cloudplatform-connectivity/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandlerTest.java
MatKuhr: IMO It didn't cover anything that isn't covered sufficiently elsewhere. Compare `testNoAuthWithTenantId` and `testNotPrincipalPropagation`.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
Johannes-Schneider: Yeah, I also thought about this. But if this is internal API anyways (and we are using it in a single place only, as of now) why do we introduce this double-check? (As far as I could see, we are already checking whether the proxy type is on-prem when building our Default Http destination)

MatKuhr: yeah, this change was motivated by some unit tests that directly called this handler and passed in INTERNET destinations. So this check is more of a sanity-check. We could also rely (exclusively) on testing in the places where we use the handler (i.e. testing the builder).

For now I figured no harm in checking it here though, it makes the production code a bit more robust and it's less likely for tests to be malformed. But no strong opinion

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
Johannes-Schneider: No super strong opinion, so fine with me.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestinationBuilderProxyHandler.java
Johannes-Schneider: No super strong opinion, so fine with me.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
```diff
@@ -323,6 +323,10 @@ private static DestinationServiceV1Response deserializeDestinationResponse( @Non
                         "Found a destination without name. A name is required for destinations defined in the destination service.");
                 }
             })
+            // it's important that we build the generic default destination here
```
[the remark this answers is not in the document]
Johannes-Schneider: As discussed online: Should be fine for now. if we ever receive customer requests that want to use the convenience methods that the `HttpDestinationProperties` interface offers just for reading those properties, we can still think about a solution.

Johannes-Schneider (approved):

PR #368, opened 2024-03-22
fix: Redundant NPE in `DefaultHttpDestination#equals` and `#hashCode`

newtork:
Follow-up from #341

Not critical, since NPE is being caught by default.
However it's easily possible to avoid throwing the exception.

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreComparator.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreComparator.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.Enumeration;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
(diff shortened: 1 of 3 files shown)


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationKeyStoreComparator.java
```diff
@@ -45,8 +46,12 @@ static int resolveKeyStoreHashCode( @Nonnull final KeyStore ks )
...
     @Nonnull
-    static Certificate[] resolveCertificatesOnly( @Nonnull final KeyStore ks )
+    static Certificate[] resolveCertificatesOnly( @Nullable final KeyStore ks )
     {
+        if( ks == null ) {
+            return new Certificate[0];
+        }
+
         final ArrayList<Certificate> out = new ArrayList<>();
         try {
```
MatKuhr, line 56: (minor) for which cases do we actually need the try here? Is there any legitimate failure case we need to handle?

newtork: "Everything" on `KeyStore` is prone to exceptions being thrown. That's why we need a `try`.

MatKuhr (approved):

PR #369, opened 2024-03-22
chore: PMD rule for `@Nonnull` / `@Nullable` on fields of public types

newtork:
Follow-up https://github.com/SAP/cloud-sdk-java/pull/368

<details><summary>Context: mapping `int priority` to `severity`</summary>

image

</details>

### DIFF
--- a/.pipeline/config.yml
+++ b/.pipeline/config.yml
@@ -30,4 +30,4 @@ stages:
     pmd:
       high: '0'
       normal: '1'
-      low: '0'
+      low: '669'
--- a/codestyle/sdk_specific_pmd_rules.xml
+++ b/codestyle/sdk_specific_pmd_rules.xml
@@ -447,6 +447,34 @@
         </example>
     </rule>
 
+    <rule name="MissingNonnullNullableAnnotationOnPublicTypes" language="java"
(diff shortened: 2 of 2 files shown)


MatKuhr (commented):
I'm a bit unsure about this one. Those are a lot of warnings, and adding `@Nonnull` everywhere is kinda annoying / code bloat. 

I guess the unwritten rule should be: If it's nullable always annotate it as such. Otherwise it will be assumed to be nonnull.
Unfortunately we can't make a PMD rule for that I guess. Would have to be some static code style analysis that goes through the AST and tries to figure out if `null` can potentially be assigned to a field not annotated as such..

CharlesDuboisSAP (commented):
image
There is an IntelliJ warning `Suggest @Nullable annotation` that we could enable and that catches the NPE in https://github.com/SAP/cloud-sdk-java/pull/368

CharlesDuboisSAP:
[lombok says](https://github.com/projectlombok/lombok/issues/2310#issuecomment-574638681) adding `@Nonnull` everywhere is is massive performance penalty and waste of resources.
They also recommend using an IDE level reminder for this use case.

Issue #371, opened 2024-03-22
Memory leak for on-premise calls via BTP destination with release 5.5.0


stephangutknecht:
## Issue Description

In [btc-cross-gateway-abap-srv](https://github.tools.sap/Business-Data-Transformation-Solution/btc-cross-gateway-abap-srv), we are using cloud-sdk to call on-premise systems via BTP destination. 
With release 5.5.0, we realized calling on-premise systems seems to steadily increase the used memory of our app instance. The pattern we see is that first, CSRF token requests are send (as token is invalid) and afterwards, the actual request is send to the destination. The memory on our app instance is not decreasing after processing is done.

We have seen exceptions like this:

`Time out occurred because of a probable connection leak. Please execute your request with try-with-resources to ensure resources are properly closed.If you are using the OData client instead to execute your request, explicitly consume the entity of the associated HttpResponse using EntityUtils.consume(httpEntity)`

With release 5.4.0, we do not see these issues. Still, memory seems to be decreasing rather slowly, we will further investigate as well on our side.

### Impact / Priority
 
Affected development phase: Release
 
Impact: Blocked

Timeline: Go-Live is in 3 weeks

## Error Message

We do see client failures when services is out-of-memory. It is however not crashing.

## Maven Dependency

com.sap.btc.cross.gateway.abap:btc-cross-gateway-abap-srv:jar:1.0.0
 +- com.sap.cds:cds-starter-spring-boot:jar:2.6.1:compile
 |  +- com.sap.cds:cds-services-api:jar:2.6.1:compile
 |  |  +- com.sap.cds:cds4j-api:jar:2.6.1:compile
 |  |  \- com.sap.cloud.environment.servicebinding.api:java-core-api:jar:0.10.3:compile
 |  +- com.sap.cds:cds-framework-spring-boot:jar:2.6.1:runtime
 |  +- org.springframework.boot:spring-boot-starter-web:jar:3.2.2:compile
 |  |  +- org.springframework.boot:spring-boot-starter-json:jar:3.2.2:compile
 |  |  |  +- com.fasterxml.jackson.datatype:jackson-datatype-jdk8:jar:2.15.3:compile
 |  |  |  \- com.fasterxml.jackson.module:jackson-module-parameter-names:jar:2.15.3:compile
 |  |  +- org.springframework.boot:spring-boot-starter-tomcat:jar:3.2.2:compile
[...]

PR #372, opened 2024-03-22
fix: Consume CSRF Token Retrieval response `HttpEntity` manually

newtork:
Potentially related https://github.com/SAP/cloud-sdk-java/issues/371

I know that `HEAD` requests shouldn't result in an HTTP response body.
But from my experience, we can't rely on _proper reasoning_.


cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultCsrfTokenRetriever.java
```diff
@@ -88,9 +89,13 @@ private Header retrieveCsrfTokenResponseHeader(
...
         headers.forEach(( k, values ) -> values.forEach(v -> csrfTokenRequest.addHeader(k, v)));
 
-        //The service path gets appended to the destination URI inside the execute request
+        // The service path gets appended to the destination URI inside the execute request
         final HttpResponse csrfResponse = httpClient.execute(csrfTokenRequest);
         final Header header = csrfResponse.getFirstHeader(X_CSRF_TOKEN_HEADER_KEY);
+
+        // Consume the HttpEntity and close the connection
+        EntityUtils.consume(csrfResponse.getEntity());
+
```
CharlesDuboisSAP, line 98:
```suggestion

        ((Closeable) httpClient).close();
```
Our `AbstractHttpClientFactory` always creates a `CloseableHttpClient` but we weakened the return to `HttpClient`.
Not sure if we reuse `HttpClient`s but couldn't find where

MatKuhr: We actually **must not** close the http client, otherwise we will break future connections which require this specific HTTP client

release_notes.md
```diff
@@ -16,7 +16,8 @@
 
 ### 📈 Improvements
 
-- 
+- Consume CSRF Token Retrieval response `HttpEntity` manually, thus closing underlying `InputStream` eagerly.
```
MatKuhr, line 19: Not sure this is worth a release note. Nothing should change for services that respond properly to HEAD requests

cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultCsrfTokenRetriever.java
```diff
@@ -88,9 +89,13 @@ private Header retrieveCsrfTokenResponseHeader(
         // Additional headers
         headers.forEach(( k, values ) -> values.forEach(v -> csrfTokenRequest.addHeader(k, v)));
 
-        //The service path gets appended to the destination URI inside the execute request
+        // The service path gets appended to the destination URI inside the execute request
         final HttpResponse csrfResponse = httpClient.execute(csrfTokenRequest);
         final Header header = csrfResponse.getFirstHeader(X_CSRF_TOKEN_HEADER_KEY);
+
+        // Consume the HttpEntity and close the connection
```
MatKuhr, line 96: Maybe:

```suggestion
        // The entity is expected to be null
        // But to be safe, discard any potential payload to ensure the connection is closed
```

To indicate this is not the expected case..

newtork: I phrased it a bit differently while adding log statement

MatKuhr (approved):

PR #388, opened 2024-04-04
feat: [Destinations] Increase Http Client Cache Duration

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#430.

### Feature scope:
 
- [x] Change to `afterLastAccess`
- [x] Increase duration to `1 Hour`

## Definition of Done

- [x] [Documentation updated](https://github.com/SAP/cloud-sdk/pull/1771)
- [x] Release notes updated


Johannes-Schneider (commented):
This change drastically increases the risk of customers running out of connections **IFF** they are not consuming batch responses correctly.
Personally, I'd be willing to take that risk, especially given that every somewhat sophisticated IDE should warn about using an `Autoclosable` outside of an `try` with resource block.
Nevertheless, I wouldn't be surprised to see some support issues coming our way, but 🤷‍♂️

Still, I'd like a second pair of eyes (and especially a second brain) to think about any implicit implications of that change.

newtork (approved):
I agree with the concerns from @Johannes-Schneider.

They made me look into `finalize()` and `AutoCloseable` API contracts and garbage collector behavior. But this didn't lead anywhere reasonable.

I think we'll take the chance.

PR #392, opened 2024-04-08
Try fix dependabot merger one more time

MatKuhr:
please


cschubertcs (approved):
I guess that looks fine? If their token works that's nice to reuse =)

PR #404, opened 2024-04-18
feat: [Destinations] Support OAuth2 Refresh Token Authentication Type

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#439.

This PR adds support for [OAuth Refresh Token Authentication](https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/oauth-refresh-token-authentication?version=Cloud).

### API Design

**Idea:** Add public API `DestinationServiceOptionsAugmenter refreshToken(@Nonnull final String refreshToken )`

Pro:

* Similar to [JS SDK API](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destinations#destination-fetch-options)
* Similar to the other two already existing options
* Refresh token is automatically part of cache key
  * In case the token is user-spceific we automatically get the correct cache isolation 

Contra:

* We have to be careful with log statements on the cache key

### Feature scope:
 
- [x] Update Enum
- [x] New API for passing the refresh token
- [x] Implementation
- [x] Tests

## Definition of Done

- [x] [Documentation updated](https://github.com/SAP/cloud-sdk/pull/1799)
- [x] Release notes updated


cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationRetrievalStrategyResolver.java
```diff
@@ -77,42 +70,36 @@ static DestinationRetrievalStrategyResolver forAllDestinations(
...
-            case ONLY_SUBSCRIBER:
-                behalfTechnicalUser = TECHNICAL_USER_CURRENT_TENANT;
-                break;
-            // sanity check
-            default:
-                throw new IllegalStateException(
-                    "Unexpected retrieval strategy "
-                        + retrievalStrategy
-                        + " when building a request towards the destination service.");
+        if ( refreshToken != null) {
```
MatKuhr, line 83: One could argue about what we should do if the user combines various values for `tokenExchangeStrategy` together with a `refreshToken`. None of the available values really make much sense together with a `refreshToken`.

Currently this code just ignores w/e `tokenExchangeStrategy` is set and just applies the refresh token flow.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationRetrievalStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.Accessors;
+
+@Getter
+@Accessors( fluent = true )
+@RequiredArgsConstructor( access = AccessLevel.PRIVATE )
+@EqualsAndHashCode
+@ToString
+final class DestinationRetrievalStrategy
```
MatKuhr, line 24: Replaces the static inner class that used to be [here](https://github.com/SAP/cloud-sdk-java/pull/404/files#diff-46b739c3e44de31f9838334d2a8347c32df25eb764b71869af01ce3992a439b2L53)

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAuthenticationTest.java
```diff
@@ -50,487 +59,281 @@ class DestinationServiceAuthenticationTest
...
-        final DestinationServiceAdapter destinationService = mock(DestinationServiceAdapter.class);
-
-        final Object payloadSuccess =
+        final ImmutableMap<String, String> properties =
             ImmutableMap
-                .<String, Object> builder()
-                .put("owner", ImmutableMap.of("SubaccountId", "a89ea924-d9c2-4eab", "InstanceId", "foobar"))
-                .put(
-                    "destinationConfiguration",
+                .<String, String> builder()
```
Johannes-Schneider, line 161: [minor]

I think this syntax is so much more readable compared to the `Map.of(...)` thingy.

So, as we are touching basically all test cases in this class anyways, why not use a consistent style?

Johannes-Schneider (commented):
Implementation and refactorings overall look pretty good!
Added just some minor comments.

However, I'd like to get a second review, as we are touching some pretty important code here.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationRetrievalStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+    private final String token;
+
+    static DestinationRetrievalStrategy withoutToken( @Nonnull final OnBehalfOf behalf )
+    {
+        return new DestinationRetrievalStrategy(behalf, NONE, null);
+    }
+
+    static DestinationRetrievalStrategy withUserToken( @Nonnull final OnBehalfOf behalf, @Nonnull final String token )
+    {
+        Objects.requireNonNull(token, "User token must not be null.");
```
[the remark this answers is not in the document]
MatKuhr: True, we usually don't. I added this here because it wouldn't fail later on, but just lead to an empty header value, which is kinda difficult to track down. 

Found out while improving unit tests, we previously had no coverage at all to check if the headers were actually being added.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAuthenticationTest.java
MatKuhr: 👍🏻✅

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationRetrievalStrategy.java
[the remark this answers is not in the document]
Johannes-Schneider: But in that case, we might want to use something like `If (token == null || token.isBlank())`, no?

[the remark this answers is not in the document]
MatKuhr: Good point, didn't think of that. Switched to `token.isBlank()` now, which would throw an NPE in case it's null, which is fine as the parameter is annotated non-null.

Johannes-Schneider (dismissed):
LGTM

Johannes-Schneider (approved):

PR #409, opened 2024-04-22
feat: Generic (Expert) API for String-Based BtpServiceOptions

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#434.

This PR adds a string-based API for creating `OptionsEnhancer` instances in the `BtpServiceOptions` class, which is meant to aid experts (i.e. CAP developers) to circumvent referencing specific classes.

## Definition of Done

- [x] ~Documentation updated~ (we decided to not document this expert API)
- [x] ~Release notes updated~ (we decided to not document this expert API)

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
@@ -46,6 +46,9 @@ public final class BtpServiceOptions
 
     /**
      * Generically creates a new instance of {@link OptionsEnhancer} for the given enhancer name and parameters.
+     * <p>
+     * <b>Note:</b> This API is meant for expert users that want to use the SDK in a more dynamic way. For the vast
+     * majority of consumers, we strongly recommend using the type-safe methods defined in this class.
      *
      * @param enhancerName
      *            The name of the enhancer to be created. This should be the simple class name of the enhancer.
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
@@ -53,6 +53,7 @@ public final class BtpServiceOptions
      *            The parameters to pass to the chosen enhancer. These should be the same (including their order) as the
      *            parameters of the type-safe methods defined in this class.
      * @return An instance of {@link OptionsEnhancer} that can be used to configure a destination.
+     * @since 5.9.0
      */
     @Beta
     public static
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
@@ -59,6 +59,7 @@ public final class BtpServiceOptions
      * @since 5.9.0
      */
     @Beta
+    @Nonnull
     public static
         OptionsEnhancer<?>
         withGenericOption( @Nonnull final String enhancerName, @Nonnull final Object... parameters )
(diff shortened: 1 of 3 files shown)


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
```diff
@@ -25,6 +30,47 @@
...
+     * @param parameters
+     *            The parameters to pass to the chosen enhancer. These should be the same (including their order) as the
+     *            parameters of the type-safe methods defined in this class.
+     * @return An instance of {@link OptionsEnhancer} that can be used to configure a destination.
+     * @since 5.9.0
+     */
+    @Beta
+    public static
+        OptionsEnhancer<?>
+        withGenericOption( @Nonnull final String enhancerName, @Nonnull final Object... parameters )
```
Johannes-Schneider, line 64: Usage example:

```java
ServiceBindingDestinationOptions.forService(ServiceIdentifier.IDENTITY_AUTHENTICATION)
        .withOption(BtpServiceOptions.withGenericOption("IasOptions", "withTargetUri", "https://my-target.uri"))
        .build();
```

StefanHenke: @Johannes-Schneider : Ideally we could get rid of the first parameter "IasOptions". When calling `ServiceBindingDestinationOptions.forService` we don´t know whether the binding is of type `xsuaa` or `ias`. Hence, we also don´t know whether to put `IasOptions` or something else. Couldn´t you derive this internally based on the binding type?
Our vision is that the parameter can be configured declarative for example in `application.yaml` and CAP can route the parameters to Cloud SDK without understanding that the binding is `xsuaa` or `ias` based. 

Something like:
```
RemoteServiceConfig config = ...;
Map<String, String> options = config.getOptions();
ServiceBinding binding = findBindingByName(config.getBindingName()); 
ServiceBindingDestinationOptions.forService(binding)
        .withOption(BtpServiceOptions.withGenericOption("withTargetUri", options.get("targetUri"))
        .build();
```

Johannes-Schneider: Hi @beckermarc and @StefanHenke,

thanks for your feedback.
I definitely understand where you are coming from and also see the value of the target state you are describing.

Actually, I also thought about how we can get there.
**However**, from my perspective, it makes sense to achieve that target state (i.e. some `withTargetUri` option that is shared between XSUAA and IAS) in two (or maybe three) steps.

1. Define the new "generic" API for the use case we are actually supporting OOTB right now (which is just the IAS case).
2. Extend our implementation with an implementation for XSUAA communication, similar to what we already have for IAS.
3. Identify the options that are common for IAS and XSUAA (e.g. the `targetUri`) and unify the options accordingly. 

From my perspective, we could also swap the order of steps `1.` and `2.`, but I would still like to have separate PRs for them.
This is because (as of now) we are having a very strict approach for our API: Each `OptionsEnhancer` is meant to be for **exactly one** service type. That is: We want the semantic of each `OptionsEnhancer` to be define very precisely.
If we were to add a "generic" `TargetUriOptionsEnhancer`, I fear, we would lose that sharp semantic.
For example, imagine we had that option, and some user would combine it with a service that is neither IAS nor XSUAA like so:

```java
ServiceBindingDestinationOptions.forService(ServiceIdentifier.BUSINESS_RULES)
        .withOption(BtpServiceOptions.BusinessRulesOptions.AUTHORING_API)
        .withOption(BtpServiceOptions.TargetUri.of("https://foo.bar"))
```

What is that supposed to mean? Which of the two options should take precedence?

To avoid these questions, I'd suggest to implement something like `AuthenticationServiceOptions` (which is defined to have common options for IAS and XSUAA, but for no other service) in step `3.`.

Personally, I think it is not too far stretched to have all of these changes implemented for version `5.9.0`, which is scheduled for next Thursday.

What do you think?
Does that make sense to you?

beckermarc: Yes totally fine with doing this in separate steps. I just wanted to make sure to emphasize @StefanHenke point again, that having something like a `IasOptions` String is a little bit in the way to our target picture.

> If we were to add a "generic" TargetUriOptionsEnhancer, I fear, we would lose that sharp semantic.

I think you could have some common option enhancers that are reused by multiple service binding destination loaders and make them "generic" in that sense. I don't think you should make it "generic" in the sense that you in your framework code handle that option automatically for all loaders. Each loader could still explicitly document which options it supports, so your BusinessRulesOptions example above would probably simply ignore that option I'd suggest.

From that perspective I'd say that a naming like `AuthenticationServiceOptions` would make the scope to narrow for me, as it semantically would restrict reusing such a target URI in a completely different scenario not tight to authentication (just some random thought: e.g. a proxy service that adds additional resilience to outgoing calls in a sidecar approach somehow)

Johannes-Schneider:
As aligned in our online sync today, we are putting this PR on ice.
Instead, we will be focussing on implementing OOTB for XSUAA communication and introduce some shared OptionsEnhancer that both IAS and XSUAA implementations consider for the target URI.

**DO NOT DELETE THIS BRANCH**

PR #413, opened 2024-04-26
feat: OAuth2Property Supplier for XSUAA Itself

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#435.

This PR adds a new implementation of our `OAuth2PropertySupplier`, which is capable of parsing service bindings to the XSUAA service itself.
Therefore, it behaves similarly to our IAS logic.

Additionally, this PR introduces a common "Target Uri" `OptionsEnhancer`, which is respected by **both** our existing **and** the new XSUAA property supplier:

```java
ServiceBindingDestinationOptions.forService(ServiceIdentifier.of("xsuaa"))
        .withOption(BtpServiceOptions.AuthenticationServiceOptions.withTargetUri("https://my-service.com"))
        .build();

ServiceBindingDestinationOptions.forService(ServiceIdentifier.IDENTITY_AUTHENTICATION)
        .withOption(BtpServiceOptions.AuthenticationServiceOptions.withTargetUri("https://my-service.com"))
        .build();
```

### Feature scope:

 
- [x] Implement new `OAuth2PropertySupplier` 
- [x] Introduce a shared "Target Uri" Option
  - [x] Deprecate the existing IAS-specific option
  - [x] Respect the new option in existing IAS logic
  - [x] Respect the new option in new XSUAA logic

## Definition of Done

- [x] Documentation updated → [Documentation PR](https://github.com/SAP/cloud-sdk/pull/1802)
- [x] Release notes updated


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
```diff
@@ -134,11 +197,22 @@ public static OptionsEnhancer<?> withTargetUri( @Nonnull final String targetUri
...
+            logTargetUriDeprecationWarning();
+            return AuthenticationServiceOptions.withTargetUri(targetUri);
+        }
+
+        private static void logTargetUriDeprecationWarning()
+        {
+            log.warn("""
+                The method '{}.withTargetUri' is deprecated and will be removed in a future version of the Cloud SDK. \
+                Use '{}.withTargetUri' instead.
+                """, IasOptions.class, AuthenticationServiceOptions.class);
```
beckermarc, line 215: Do we really need this warning in the application logs?

beckermarc (approved):
Looks good 👍

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
Johannes-Schneider: As our entire API is currently still in `@Beta`, I'd like to use that as a chance to remove these methods still in Cloud SDK 5.
Therefore, I want users to be extra aware that something will break in the not-too-far future.

Of course, we have to align whether we actually want to break our API (even a Beta one) internally.
If we decide **against** that approach, I'd get rid of the log.

MatKuhr:
> ... both resolvers are present. Which one would take precedence? 

I thought about this as well, the CAP one will take precedence. We prepend any resolver that is manually registered ([source](https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceBindingDestinationLoader.java#L85)).

I think we are safe as long as the options enhancer used matches the resolver used. Which I'm not sure if we can guarantee that from CAP side..?

beckermarc:
> I think we are safe as long as the options enhancer used matches the resolver used. Which I'm not sure if we can guarantee that from CAP side..?

If custom resolvers are always prepended and take precedence CAPs own usage of your APIs to create destinations from XSUAA bindings will always succeed. However the new approach outlined here will not work for other parties as long as CAP doesn't remove its own resolver. They would need to use the CAP OptionEnhancer during that time, which they are not aware of however. I think this is acceptable, as CAP will anyway switch to Cloud SDK 5 very soon.

StefanHenke:
> > I think we are safe as long as the options enhancer used matches the resolver used. Which I'm not sure if we can guarantee that from CAP side..?
> 
> If custom resolvers are always prepended and take precedence CAPs own usage of your APIs to create destinations from XSUAA bindings will always succeed. However the new approach outlined here will not work for other parties as long as CAP doesn't remove its own resolver. They would need to use the CAP OptionEnhancer during that time, which they are not aware of however. I think this is acceptable, as CAP will anyway switch to Cloud SDK 5 very soon.

Alternatively, we could register our resolver only if the one from Cloud SDK is not available on the classpath. In that case, we wouldn´t break customer code. If we add both uri enhancers anyway, this should work.

beckermarc:
> Alternatively, we could register our resolver only if the one from Cloud SDK is not available on the classpath.

Yes, that would be possible. We can anyway only add the new option enhancer in that case as well.
But I guess we don't really need to overthink this, as all of these issues will be removed, if we increase our minimum Cloud SDK version :)

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -98,6 +103,43 @@ static List<OAuth2PropertySupplierResolver> getDefaultServiceResolvers()
...
+        @Nonnull
+        @Override
+        public URI getServiceUri()
+        {
+            final Option<URI> maybeTargetUri = options.getOption(TargetUri.class);
+            if( maybeTargetUri.isDefined() ) {
+                return maybeTargetUri.get();
+            }
+
+            // we need to get the token uri here because it respects the different properties for either
```
MatKuhr, line 122: I'm thinking if we really want to support this case. If we do, I am not sure if it's safe to return the certurl? 
Because that would require us adding the KeyStore which we currently don't do for XSUAA, right? Also, I'm not 100% sure if accessing other XSUAA endpoints is allowed under the certurl. Intuitively it should be, but who knows..

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
```diff
@@ -98,6 +103,43 @@ static List<OAuth2PropertySupplierResolver> getDefaultServiceResolvers()
...
+            // certificate (X509) or client secret authentication.
+            // so it either chooses the "url" or the "certurl" property.
+            // as we are overriding the method (see below), we need to use the super implementation to not get the
+            // "/oauth/token" appended to the URL.
+            return super.getTokenUri();
+        }
+
+        @Nonnull
+        @Override
+        public URI getTokenUri()
```
MatKuhr, line 132: Why do we need this implementation? Can't we let this be handled by the `OAuth2Service.Builder` like for the other XSUAA based services?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
Johannes-Schneider: > Because that would require us adding the KeyStore which we currently don't do for XSUAA, right?

Good point, that's indeed true.

---

> Also, I'm not 100% sure if accessing other XSUAA endpoints is allowed under the certurl

No clue.
I guess lets just use the regular `url` in any case and wait until someone complains? 😅

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServicePropertySuppliers.java
Johannes-Schneider: Yes, we can - my bad 🤦‍♂️

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
MatKuhr: I am wondering if this is worth deprecating. It would arguably be a bit more convenient for users who are using this API directly, if we keep this option, because it is a bit more discoverable, and we can offer more explicit Javadoc here. Since it's just the API it doesn't add that much bloat to our code, so maybe worth just keeping?

Johannes-Schneider: > It would arguably be a bit more convenient for users who are using this API directly, if we keep this option, because it is a bit more discoverable

I think that is a fair statement.
**However**, the API would be somewhat inconsistent if we had such a discoverable API for IAS but not for XSUAA, right?
So we would need to introduce an `XsuaaOptions` factory class.
I'm afraid that we are making our API very prone to becoming bloated in the future, if we were to pursue that sort of mentality.

Also, as this is "expert" API, I wouldn't be concerned with convenience and ease of discoverability too much, to be honest.
I'd expect this option to be documented (of course) and assume that expert users are able to find and follow our documentation.

MatKuhr: I'm not sure I would consider this expert API, since this is expected to be used by any user who wants to do app-2-app communication with XSUAA or IAS, which is a rather basic use case?

Anyhow, for now I was thinking to not add a `XsuaaOptions`, since we don't have further options for XSUAA, and only keep the IAS one because there we have more options. But on the other hand, chances are we might add some XSUAA specific options in the future, and then it would become a bit more API bloat, or otherwise would be inconsistent.

No strong opinion honestly. I like the small benefit of more intuitive naming, discoverability and chance to have dedicated API docs, but those are rather minor benefits to accept code/api duplication, so I am happy with w/e you prefer 😄 

BTW, we could also straight up delete the IAS option now, because otherwise we keep the deprecation bloat in the SDK and need yet another "update this later" PR.. We probably don't have that many users yet, right? Or would this break the current CAP usage?

beckermarc: > Or would this break the current CAP usage?

Yep, this would unfortunately now break the (just released) CAP usage, where we internally call this API now, in one specific scenario, when Cloud SDK 5 is used.

MatKuhr (approved):

PR #417, opened 2024-04-29
refactor: Dependabot Auto Merge GH Workflow

Johannes-Schneider:
This PR refactors our automated PR merging for Dependabot PRs in a way that the workflow is triggered on a schedule (every Monday, after Dependabot might have opened new PRs) instead of triggering right after the PR creation.

The old approach has shown to be non-functioning as the workflow was triggered by an external user (i.e. Dependabot), which (we assume) prevented the workflow from accessing any our secrets.
As a consequence, the workflow always failed.

This updated approach does not run into that issue as the workflow is not triggered by an external user.
So we can safely access all of our secrets.

The updated workflow has been tested here: [Workflow Run](https://github.com/SAP/cloud-sdk-java/actions/runs/8878953514/job/24375619632)


MatKuhr (approved):
LGTM 😄

PR #420, opened 2024-05-02
fix: Derive current `Principal` from `Basic` Authentication Header

Johannes-Schneider:
## Context
 

SAP/cloud-sdk-java-backlog#440.

This PR fixes a regression introduced in Cloud SDK 5.0, where we silently (and unintentionally) dropped support for deriving the current tenant from a `Basic` Authentication header, if neither a JWT nor a OIDC token are present.

### Feature scope:

 
- [x] Consider `Basic` authentication header as a last fallback 

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/security/src/test/java/com/sap/cloud/sdk/cloudplatform/security/principal/DefaultPrincipalFacadeTest.java
```diff
@@ -0,0 +1,29 @@
...
+    {
+        final String basicAuthValue = "username:pass";
+        final String basicAuthHeaderValue = "Basic " + Base64.getEncoder().encodeToString(basicAuthValue.getBytes());
+
+        RequestHeaderAccessor
+            .executeWithHeaderContainer(Map.of(HttpHeaders.AUTHORIZATION, basicAuthHeaderValue), () -> {
+                final PrincipalFacade sut = new DefaultPrincipalFacade();
+                final Principal principal = sut.tryGetCurrentPrincipal().get();
+
+                assertEquals("username", principal.getPrincipalId());
```
MatKuhr, line 26:
```suggestion
                assertEquals("username", principal.getPrincipalId());
                assertEquals("username", PrincipalAccessor.getCurrentPrincipal().getPrincipalId());
```

cloudplatform/security/src/test/java/com/sap/cloud/sdk/cloudplatform/security/principal/PrincipalThreadContextListenerTest.java
```diff
@@ -0,0 +1,22 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Collection;
+
+import org.junit.jupiter.api.Test;
+
+import com.sap.cloud.sdk.cloudplatform.thread.ThreadContextListener;
+import com.sap.cloud.sdk.cloudplatform.util.FacadeLocator;
+
+class PrincipalThreadContextListenerTest
```
MatKuhr, line 12: How about adding this [here](https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/thread/ThreadContextTest.java#L74) instead?

release_notes.md
```diff
@@ -29,4 +29,4 @@
 
 ### 🐛 Fixed Issues
 
-- 
+- Fix a regression that was introduced with the SAP Cloud SDK 5.0 release where the principal would no longer be derived from a `Basic` authorization header, in cases where neither a JWT nor an OIDC was present.
```
MatKuhr, line 32:
```suggestion
- Fix a regression that was introduced with the SAP Cloud SDK 5.0 release where the principal would no longer be derived from a `Basic` authorization header, in cases where neither a JWT nor an OIDC token was present.
```

MatKuhr (approved):
lgtm with some minor comments on test code

cloudplatform/security/src/test/java/com/sap/cloud/sdk/cloudplatform/security/principal/DefaultPrincipalFacadeTest.java
```diff
@@ -23,7 +23,7 @@ void testExtractPrincipalFromBasicAuthHeader()
                 final PrincipalFacade sut = new DefaultPrincipalFacade();
                 final Principal principal = sut.tryGetCurrentPrincipal().get();
 
-                assertEquals("username", principal.getPrincipalId());
+                assertThat(principal.getPrincipalId()).isEqualTo("username");
```
MatKuhr, line 26: What is a bit tricky is that this doesn't cover if the `sut` actually works when invoked by the listener during the thread context creation. Subtle difference, but potentially a relevant difference? 

That was my idea behind asserting on the result of `PrincipalAccessor.getCurrentPrincipal().getPrincipalId()` 😄 

First I thought maybe there could be an issue if the principal is accessed before the basic credentials are computed and stored in the thread context, but turns out this wouldn't be a problem, the computation would just happen twice.

MatKuhr (approved):

cloudplatform/security/src/test/java/com/sap/cloud/sdk/cloudplatform/security/principal/DefaultPrincipalFacadeTest.java
Johannes-Schneider: > What is a bit tricky is that this doesn't cover if the sut actually works when invoked by the listener during the thread context creation. Subtle difference, but potentially a relevant difference?

hmm... good point.
However, since there is just the single method (`tryGetCurrentPrincipal`) in the implementation, I wouldn't know how it would make a difference **when** the method is being called.
(Of course, assuming that the headers are set correctly, which should always be the case since the headers are set within our servlet filter)

The other concern might be about the `ThreadContextListener` implementation, which is sort of the counter-part to the facade.
I improved these tests a bit to make sure the listener is working as expected.

With that, I think we should be good 👍

MatKuhr (approved):

PR #437, opened 2024-05-16
Add comment about `ConnectionManager` cardinality in `AbstractHttpClientFactory`

newtork:
I was wondering why this was part of the `HttpClientWrapper`, when I realized - most likely we do not and cannot reuse connection manager between multiple instances of `HttpClient`:

https://github.com/SAP/cloud-sdk-java/blob/f0dd44016892ef1f1fbd758b83c84d00948f9854/cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/HttpClientWrapper.java#L38-L48

To address this in the source code, I want to add a comment.


MatKuhr (commented):
this is probably true for the other things as well. But regardless, good to have 👍🏻

MatKuhr (approved):

PR #453, opened 2024-05-27
feat: Support ZTIS with XSUAA

MatKuhr:
## Context

This PR improves the credential type detection to deal with a lowercase variant of `x509_attested`, which will be used by XSUAA.

- [x] E2E tested

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
@@ -111,7 +111,7 @@ public class DefaultOAuth2PropertySupplier implements OAuth2PropertySupplier
     @Nonnull
     public URI getTokenUri()
     {
-        final String tokenUrlProperty = switch(getCredentialType()) {
+        final String tokenUrlProperty = switch( getCredentialType() ) {
             case X509, X509_GENERATED, X509_PROVIDED, X509_ATTESTED -> "certurl";
             case BINDING_SECRET, INSTANCE_SECRET -> "url";
         };
@@ -129,7 +129,7 @@ public class DefaultOAuth2PropertySupplier implements OAuth2PropertySupplier
             case X509_ATTESTED -> getZtisIdentity(clientid);
             case BINDING_SECRET, INSTANCE_SECRET -> getSecretIdentity(clientid);
             case X509_PROVIDED -> throw new DestinationAccessException(
-                "Credential type X509_PROVIDED is not supported. Please use X509_ATTESTED instead.");
+                "Credential type X509_PROVIDED is not supported. Please use X509_GENERATED or X509_ATTESTED instead.");
         };
     }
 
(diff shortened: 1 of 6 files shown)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
```diff
@@ -134,22 +145,15 @@ protected List<String> getOAuthPropertyPath()
     }
 
     @Nonnull
-    ClientIdentity getCertificateIdentity()
+    private ClientIdentity getCertificateIdentity( @Nonnull final String clientid )
     {
-        final String clientid = getOAuthCredentialOrThrow(String.class, "clientid");
-
-        final Option<String> exactCredentialType = getOAuthCredential(String.class, "credential-type");
```
MatKuhr, line 141: see [the impact](https://github.com/SAP/cloud-sdk-java/pull/453/files#diff-00bbe36107fd3c1daad27fb5a5194bcf955ce907a4fcd01db8d72d4dbb83e4d5R11) of this refactoring. I figured since the feature is very new, SAP internal only and probably not used a lot yet we can accept this

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultOAuth2PropertySupplier.java
```diff
@@ -111,15 +111,26 @@ public URI getServiceUri()
...
     }
 
     @Override
     @Nonnull
     public ClientIdentity getClientIdentity()
     {
-        return getCredentialType() == CredentialType.X509 ? getCertificateIdentity() : getSecretIdentity();
+        final String clientid = getOAuthCredentialOrThrow(String.class, "clientid");
+
+        return switch( getCredentialType() ) {
```
MatKuhr, line 127: In preparation of removing `SecurityLibWorkarounds` via https://github.com/SAP/cloud-sdk-java-backlog/issues/428

Alternatively, one could apply [this change](https://github.com/SAP/cloud-sdk-java/pull/453/files#diff-5b17f73003c42fdff01a047bd9bf07d362d5be7ee56b27e1cefed7d22b4bed9aR39-R41) standalone in this PR.

release_notes.md
```diff
@@ -8,11 +8,11 @@
 
 ### 🔧 Compatibility Notes
 
-- 
+- Using the `X509_ATTESTED` credential type now requires a version >= `3.4.0` of the [BTP Security Library](https://github.com/SAP/cloud-security-services-integration-library).
```
newtork, line 11: I wonder what happens at runtime, or how "hard" this requirement is (or needs to be).

MatKuhr: We would log a warning that we don't recongise `X509_ATTESTED`, then try to apply `BINDING_SECRET` but fail with an IllegalSomethingException, because the binding doesn't contain a secret.

newtork: I meant: in these lines we are referencing the enums at runtime directly.
image

Rendering `3.4.0` the new lower bound for (transitive) security library dependency version

newtork (approved):
Please address /  take note of my comment here
https://github.com/SAP/cloud-sdk-java/pull/453#discussion_r1625579737

Looks good to me.

release_notes.md
MatKuhr: Good point!

I tried out what happens in a demo application. Turns out: This seems to be fine. I compiled the SDK with version 3.5.0 and then used it in the app with 3.3.5. Nothing bad happens 😄

```
CredentialType type = SecurityLibWorkarounds.getCredentialType("X509_GENERATED");
System.out.println(type); // x509
System.out.println(List.of(CredentialType.values())); // [x509, instance-secret, binding-secret]
```

Where I changed `SecurityLibWorkarounds.getCredentialType` to also include the switch statement, just to quickly run this test.

PR #458, opened 2024-06-03
fix: [IAS] ZTIS for Subscriber Tenants when using IAS

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#450.

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
@@ -111,7 +111,10 @@ class OAuth2Service
 
         final DefaultHttpDestination destination =
             DefaultHttpDestination
-                .builder(tokenUri)
+                // Giving an empty URL here as a workaround
+                // If we were to give the token URL here we can't change the subdomain later
+                // But the subdomain represents the tenant in case of IAS, so we have to change the subdomain per-tenant
+                .builder("")
                 .name("oauth-destination-ztis-" + identity.getId().hashCode())
                 .keyStore(((ZtisClientIdentity) identity).getKeyStore())
                 .build();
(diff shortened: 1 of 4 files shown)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -111,7 +111,10 @@ private OAuth2TokenService createTokenService( @Nonnull final CacheKey ignored )
 
         final DefaultHttpDestination destination =
             DefaultHttpDestination
-                .builder(tokenUri)
+                // Giving an empty URL here as a workaround
+                // If we were to give the token URL here we can't change the subdomain later
+                // But the subdomain represents the tenant in case of IAS, so we have to change the subdomain per-tenant
+                .builder("")
```
MatKuhr, line 117: not very pretty, but very effective 😅

In the long term we could consider replacing our implementation with the one from security lib. They recently added support [here](https://github.com/SAP/cloud-security-services-integration-library/pull/1528). But I figured that's too much work for a quick fix 😬

cloudplatform/connectivity-oauth/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2ServiceTest.java
```diff
@@ -366,14 +370,20 @@ void testCacheIsRegistered()
 
     @Test
     void testZeroTrustClientIdentity()
+        throws KeyStoreException,
+            CertificateException,
+            IOException,
+            NoSuchAlgorithmException
     {
-        ClientIdentity identity = new ZtisClientIdentity("id", mock(KeyStore.class));
+        final KeyStore ks = KeyStore.getInstance("JKS");
```
MatKuhr, line 378: Http Client Cache tries to compare the certificates which doesn't work with a mock

newtork: Args... this is unfortunate. We could only drop the throws declarations, for fewer LoCs.

newtork (approved):
LGTM

PR #462, opened 2024-06-06
Deprecate non functional IAS option

MatKuhr:
## Context

SAP/cloud-sdk-java-backlog#454

IAS doesn't allow for sending `clientid:client-id:apptid:tenant-id`. We might have built this based on some older documentation from IAS that in the meanwhile got updated. Or I made a copy/paste mistake when copying the requirements for the parameters into our ticket 😅

Anyhow, deprecating the non-functional option.

## Definition of Done

- [x] [Documentation updated](https://github.com/SAP/cloud-sdk/pull/1828)
- [x] Release notes updated

### DIFF
--- a/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
+++ b/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java
@@ -220,7 +220,7 @@ public final class BtpServiceOptions
          * Creates an {@link OptionsEnhancer} that instructs an IAS-based destination to use the given application
          * provider name when performing token retrievals. This is needed in <b>App-To-App</b> communication scenarios.
          * <p>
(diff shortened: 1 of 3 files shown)


CharlesDuboisSAP (approved):
Why don't you remove it all together since it's `@Beta`

MatKuhr:
> Why don't you remove it all together since it's @Beta

Yeah I was considering that too, but figured I don't want to spend more time thinking if this could somehow break something.  It should probably be fine, since it's non-functional currently, but breaking API changes are always a bit dangerous. No strong opinion though TBH.

I could also already prepare a PR and add it to the [September BLI](https://github.com/SAP/cloud-sdk-java-backlog/issues/428) 😄

CharlesDuboisSAP:
> > Why don't you remove it all together since it's @beta
> 
> Yeah I was considering that too, but figured I don't want to spend more time thinking if this could somehow break something. It should probably be fine, since it's non-functional currently, but breaking API changes are always a bit dangerous. No strong opinion though TBH.
> 
> I could also already prepare a PR and add it to the [September BLI](https://github.com/SAP/cloud-sdk-java-backlog/issues/428) 😄

Please do

PR #467, opened 2024-06-10
Javadoc fix

CharlesDuboisSAP:
Javadoc change that we forgot for v5


MatKuhr (commented):
huh, I thought we did a global search for this stuff. A quick search reveals 4 more findings for me: https://github.com/search?q=repo:SAP/cloud-sdk-java%20scpcf&type=code

Maybe fix them as well while we are on it? looks trivial from a quick look ;)

cloudplatform/connectivity-destination-service/src/main/resources/META-INF/services/com.sap.cloud.sdk.cloudplatform.connectivity.DestinationFacade
```diff
@@ -1 +0,0 @@
-com.sap.cloud.sdk.cloudplatform.connectivity.ScpCfDestinationFacade
```
newtork, line 1: F*ck, we didn't notice this? Guess there wasn't a big warning.

newtork (approved):

PR #469, opened 2024-06-12
(Legacy) Remove confusing comment

newtork:
Remove confusing inline code comment


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ServiceBindingDestinationOptions.java
```diff
@@ -70,7 +70,6 @@ public final class ServiceBindingDestinationOptions
     @Nonnull
     public static Builder forService( @Nonnull final ServiceIdentifier identifier )
     {
-        // Move this out of here once we fix that currently the actual ServiceBindingAccessor instance is on ScpCf platform
```
newtork, line 73: > Move this out of here once we fix that
> currently the actual ServiceBindingAccessor instance is on ScpCf platform

This comment came from deprecated SDK v4 when we still had a platform class. The idea was to offer a central reference to `ServiceBindingAccessor`. It would've been easy to mock for testing, troubleshooting and for workarounds. However we never established "a central platform" class again in SDK v5. Therefore this comment longer makes sense.

CharlesDuboisSAP (approved):

PR #488, opened 2024-06-28
[chore] Ignore Json assert update

KavithaSiva:
The latest patch version updates breaks our build as it requires Java 21. 
For more details: https://github.com/skyscreamer/JSONassert/issues/190


KavithaSiva:
New patch version `1.53` fixes the issue, so we don't require an exclusion.

PR #497, opened 2024-07-09
Fragment Destinations are compatible with Change Detection

CharlesDuboisSAP:
# WiP

TODO: remove the fragment property from the cache key.


MatKuhr:
This doesn't work, as there will can be arbitrary properties that are new or changed via the fragment. In fact, there should always at least be one more, otherwise the fragment would be completely empty and thus not useful.

Without knowing which properties were changed by the fragment, and without knowing if the fragment itself changed, we cannot perform change detection.

MatKuhr:
So, URL is just one example, you can change or add any property with a fragment 😉

PR #506, opened 2024-07-23
Add comment to DefaultHttpClientFactory

newtork:


cloudplatform/connectivity-apache-httpclient4/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpClientFactory.java
```diff
@@ -61,9 +61,9 @@ public class DefaultHttpClientFactory extends AbstractHttpClientFactory
     protected RequestConfig.Builder getRequestConfigBuilder( @Nullable final HttpDestinationProperties destination )
     {
         return super.getRequestConfigBuilder(destination)
-            .setConnectTimeout(getTimeoutMilliseconds())
-            .setConnectionRequestTimeout(getTimeoutMilliseconds())
-            .setSocketTimeout(getTimeoutMilliseconds());
+            .setConnectTimeout(getTimeoutMilliseconds()) // until a connection is established
```
MatKuhr, line 64: probably good to also add to http client 5 😉 
did you find a definitive source for this information? I searched for it in the past, but only found somewhat vague statements..

newtork: I got it from here
https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html

<details><summary>HttpClient5 has a time-to-live setting apparently, that we do not use.</summary>



</details>

PR #509, opened 2024-07-26
Fix automated testing of sample modules

newtork:
While experimenting with the code generator, we found our sample modules are not being tested, nor compiled.
- [x] Add dedicated pipeline step.
  **Explanation:** For usual unit tests we rely on already compiled classes. Solely running `surefire:test` will neither compile the test code nor invoke the code generation on those sample modules. And I didn't feel comfortable manipulating the regular test stage.
- [x] Update formatted code and update unit tests


.github/workflows/continuous-integration.yaml
```diff
@@ -223,6 +223,14 @@ jobs:
           echo "[DEBUG] Running Maven with arguments: $MVN_ARGS"
           mvn $MVN_ARGS
 
+      - name: "Run Sample Modules Test"
+        run: |
+          MVN_MODULES="-pl :odata-api-sample,:odata-v4-api-sample,:openapi-api-sample"
```
MatKuhr, line 228: running this here has the problem that it runs after the local changes check, thus formtting is not enforced

.github/workflows/continuous-integration.yaml
```diff
@@ -223,6 +223,14 @@ jobs:
...
+          MVN_ARGS="${{ env.MVN_SINGLE_THREADED_ARGS }} org.jacoco:jacoco-maven-plugin:prepare-agent test org.jacoco:jacoco-maven-plugin:report --fail-at-end $MVN_MODULES -Dsdk.code.scanning"
```
MatKuhr, line 229: I am surprised this doesn't push our coverage down

.github/workflows/continuous-integration.yaml
newtork: Local-changes-check runs between "install" and "Upload (Release Artifact|SDK M2|SDK TARGET)". I feel hesitant to put the sample code there. Maybe we could put another check here, after the tests. However there's a chance pipeline will fail because of formatted test could polluting the directory. (Is this a true/false-positive then?) ^^

.github/workflows/continuous-integration.yaml
newtork: Strange. I removed the jacoco references from the command line. Not necessary (presumably)

MatKuhr: Did that change the coverage values at all?

.github/workflows/continuous-integration.yaml
MatKuhr: Not sure how it could pollute, since there are no changes if properly formatted?

Anyhow, why not let the module build during the normal build phase?

newtork: > Not sure how it could pollute, since there are no changes if properly formatted?

My concern is pollution of the release. But on the other hand we should trust the release toggles right? 🤔 

> Anyhow, why not let the module build during the normal build phase?

I don't remember what was the motivation. My guess is order of compilation; maybe the sample modules were build before the plugins completed.

I can check what happens if I change it.

newtork: It works!

I changed something in `generator` (and `generator-plugin`), and expected `sample` to pick it up while respecting the build order:

```
> mvn clean install -pl :openapi-parent
[INFO] Reactor Build Order:
[INFO] 
[INFO] Data Model - OpenAPI Services - Parent POM                         [pom]
[INFO] Data Model - OpenAPI Services - Core                               [jar]
[INFO] Data Model - OpenAPI Services - Generator                          [jar]
[INFO] Data Model - OpenAPI Services - Generator Maven Plugin    [maven-plugin]
[INFO] Data Model - OpenAPI Services - Sample API                         [jar]
```

And it worked, afterwards the `sample` module updated to mirror the changes in `generator`. I'm changing the POMs accordingly and revert the workflows.

.github/workflows/continuous-integration.yaml
newtork: I changed the scenario from plugin config exclusion to plugin skipping.
This way the coverage stayed good.

CharlesDuboisSAP (approved):

PR #517, opened 2024-07-30
Fix JDK22 compiler warnings

newtork:
Let's make the SAP Cloud SDK compatible.

I'm using `temurin-22.0.1`

I think most of the `this-escape` warnings are too difficult to fix, without breaking API and behavior.


MatKuhr (approved):
LGTM

PR #523, opened 2024-08-05
Fix Staging Deploy in Release workflow

CharlesDuboisSAP:
Fix the broken Perform Release

### DIFF
--- a/.github/workflows/perform-release.yml
+++ b/.github/workflows/perform-release.yml
@@ -139,7 +139,7 @@ jobs:
         env:
           PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
 
-      - name: "Perform Release"
+      - name: "Deploy Locally"
         run: >
           mvn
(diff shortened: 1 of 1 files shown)


.github/workflows/perform-release.yml
```diff
@@ -149,7 +149,13 @@ jobs:
           -Dgpg.passphrase="$GPG_PASSPHRASE"
           -Dgpg.keyname="$MAVEN_CENTRAL_USER"
           deploy
+        env:
+          MAVEN_CENTRAL_USER: ${{ secrets.MAVEN_CENTRAL_USER }}
+          MAVEN_CENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
```
newtork, line 154: The env var `MAVEN_CENTRAL_PASSWORD` seems to be redundant.
But we had it inside in `main` already. So, no action required.

newtork (approved):

PR #529, opened 2024-08-08
[OpenAPI Generator] When deleting files, only consider specific packages

newtork:
~It would be good if we checked for any unexpected files before blindly deleting user-provided paths.~
To stay safe, we should only delete what was being generated, according to the configuration provided.

\* Also I want us to retain "hidden" files (with `.`-prefix), to keep custom files like `.gitignore` or `.openapi-generator-ignore`)

### DIFF
--- a/datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGenerator.java
+++ b/datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGenerator.java
@@ -4,6 +4,8 @@
 
 package com.sap.cloud.sdk.datamodel.openapi.generator;
 
+import static org.apache.commons.io.filefilter.TrueFileFilter.TRUE;
+
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
@@ -17,7 +19,6 @@ import java.util.function.Consumer;
 import javax.annotation.Nonnull;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.filefilter.TrueFileFilter;
 import org.apache.commons.io.function.IOConsumer;
 import org.openapitools.codegen.ClientOptInput;
 import org.openapitools.codegen.DefaultGenerator;
@@ -121,9 +122,9 @@ public class DataModelGenerator
             log.info("Deleting output directory \"{}\".", outputDirectory.getAbsolutePath());
 
             // create set of directories that can be deleted
-            HashSet<File> allowedDirs = new HashSet<>();
+            final var allowedDirs = new HashSet<File>();
             allowedDirs.add(outputDirectory);
-            Consumer<String> addDirs = pckg -> {
+            final Consumer<String> addDirs = pckg -> {
                 File d = outputDirectory;
                 for( final String ns : pckg.split("\\.") ) {
                     d = new File(d, ns);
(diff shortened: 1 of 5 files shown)


datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGenerator.java
```diff
@@ -115,7 +120,36 @@ private void cleanOutputDirectoryIfRequested( final GenerationConfiguration gene
...
+                for( final String ns : pckg.split("\\.") ) {
+                    d = new File(d, ns);
+                    if( !d.exists() || !d.isDirectory() ) {
+                        return;
+                    }
+                    allowedDirs.add(d);
+                }
+            };
+            addDirs.accept(generationConfiguration.getModelPackage());
+            addDirs.accept(generationConfiguration.getApiPackage());
```
MatKuhr, line 138: Here I am afraid users might change the naming and then be surprised that not everything is cleaned up..

newtork: After talking, we came to the conclusion. The additional effort of manually deleting local directories after changing the _package_ properties is acceptable. We consider this to be an edge case, which justifies manual checks anyway.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGenerator.java
```diff
@@ -114,8 +116,19 @@ private void cleanOutputDirectoryIfRequested( final GenerationConfiguration gene
         if( generationConfiguration.deleteOutputDirectory()
             && outputDirectory.exists()
             && outputDirectory.isDirectory() ) {
-            log.info("Deleting output directory \"{}\".", outputDirectory.getAbsolutePath());
-            FileUtils.cleanDirectory(outputDirectory);
+            log.info("Cleaning generated folders in output directory \"{}\".", outputDirectory.getAbsolutePath());
+
+            final var pckModel = generationConfiguration.getModelPackage();
+            final var dirModel = Arrays.stream(pckModel.split("\\.")).reduce(outputDirectory, File::new, ( a, b ) -> a);
```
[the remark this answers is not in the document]
MatKuhr: nevermind, as discussed, only relevant for parallel stream execution

MatKuhr (dismissed):

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGenerator.java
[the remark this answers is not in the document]
newtork: As discussed, not an oversight. But also not good coding practice.
Especially in this risky API we better avoid ambiguous code.

I'm changing the code.
(We also considered throwing an exception in case of unexpected behavior)

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGenerator.java
```diff
@@ -144,22 +151,21 @@ private void assertTemplatesAvailableOnClasspath(
...
-
-        if( configuration.getApiPackage() == null || configuration.getApiPackage().isEmpty() ) {
-            throw new IllegalArgumentException("API package is null or empty.");
+        if( config.getOutputDirectory() == null || config.getOutputDirectory().isEmpty() ) {
+            throw new IllegalArgumentException("Output directory is null or empty.");
         }
 
-        if( configuration.getModelPackage() == null || configuration.getModelPackage().isEmpty() ) {
-            throw new IllegalArgumentException("Model package is null or empty.");
+        final var packagePattern = Pattern.compile("[a-z_$][a-z0-9_$]+(\\.[a-z0-9_$]+)*");
```
MatKuhr, line 163: Hmm with this I am a bit scared if people have non-english symbols in their package names, e.g. german umlaute? From a quick google search it seems like this is possible and allowed..

Still, I think it's fair to enforce this pattern, but then we might want to make the exception more explicit and add a release note

newtork: I forgot to comment in advance: This is according to java language spec (if I translated it correctly)
https://stackoverflow.com/a/23251187

newtork: After finding out (late to the game).. that you are 100% right about unicode support in class- and package names, I revisited the rule and changed from "strict" regex to "lenient" predicate.

MatKuhr (approved):

PR #535, opened 2024-08-12
fix: gRPC Version

MatKuhr:
## Context

The approach of only using `dependencyManagement` doesn't work, as this only affects the version within the module itself, not the version when using the module.

Unfortunately, we can only override the version by referencing transitive dependencies explicitly. We could add it to the main BOM, but that only fixes it when the BOM is used by customers. 

## Definition of Done

- [x] [Tests cover the scope above](https://sdk-v2.cpe.c.eu-de-1.cloud.sap/job/end-to-end-tests/view/Cloud%20SDK%20v5/job/v5-scp-cf-spring-ias/465/)
- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-ztis/pom.xml
+++ b/cloudplatform/connectivity-ztis/pom.xml
@@ -87,6 +87,22 @@
 			<version>0.8.6</version>
 			<scope>runtime</scope>
 		</dependency>
+		<!-- Override grpc versions from spiffe -->
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-netty-shaded</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-protobuf</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-stub</artifactId>
+			<scope>runtime</scope>
+		</dependency>
 		<dependency>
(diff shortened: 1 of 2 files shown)


cloudplatform/connectivity-ztis/pom.xml
```diff
@@ -87,6 +87,22 @@
 			<version>0.8.6</version>
```
CharlesDuboisSAP, line 87: `java-spiffe-core` also uses `grpc`
image

cloudplatform/connectivity-ztis/pom.xml
```diff
@@ -87,6 +87,22 @@
...
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-protobuf</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-stub</artifactId>
+			<scope>runtime</scope>
+		</dependency>
```
CharlesDuboisSAP, line 105: I thought only `grpc-netty-shaded` would need to be excluded, why also `grpc-protobuf` and `grpc-stub`?
image

cloudplatform/connectivity-ztis/pom.xml
MatKuhr: the versions 1.62.0 will be omitted by maven in favor of 1.66.0

cloudplatform/connectivity-ztis/pom.xml
MatKuhr: no need to exclude, we only need to force the version

PR #555, opened 2024-08-20
fix: Destination Service Error Handling

MatKuhr:
## Context

This fixes a bug that would cause HTTP connections to stay open in case the destination service returns an error with payload. Which is the case for e.g. 404 responses.

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAdapter.java
+++ b/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAdapter.java
@@ -22,7 +22,9 @@ import org.apache.http.HttpStatus;
 import org.apache.http.StatusLine;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpUriRequest;
+import org.slf4j.event.Level;
 
+import com.google.common.base.Strings;
 import com.sap.cloud.environment.servicebinding.api.DefaultServiceBindingAccessor;
 import com.sap.cloud.environment.servicebinding.api.ServiceBinding;
 import com.sap.cloud.environment.servicebinding.api.ServiceIdentifier;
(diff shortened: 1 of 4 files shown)


cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAdapter.java
```diff
@@ -148,35 +150,37 @@ private static String handleResponse( final HttpUriRequest request, final HttpRe
...
-                        .format(
-                            "Failed to get destinations: destination service returned HTTP status %s (%S) at '%s'.,",
-                            statusCode,
-                            reasonPhrase,
-                            requestUri));
-            }
+        Try<String> maybeBody = Try.of(() -> HttpEntityUtil.getResponseBody(response));
+        String logMessage = "Destination service returned HTTP status %s (%s)";
+        if( maybeBody.isFailure() ) {
+            final var ex =
```
MatKuhr, line 156: (used `var` for better formatting)

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAdapterTest.java
```diff
@@ -370,6 +373,39 @@ void getDestinationServiceProviderTenantShouldThrowForMissingId()
...
+        assertThatThrownBy(
+            () -> adapterToTest
+                .getConfigurationAsJson(
+                    "/dest",
+                    DestinationRetrievalStrategy.withoutToken(TECHNICAL_USER_CURRENT_TENANT)))
+            .describedAs("A 404 should produce a DestinationNotFoundException")
+            .isInstanceOf(DestinationNotFoundException.class)
+            .hasMessageContaining("Destination could not be found");
+
+        // invoke adapter 200 times, ensuring the response content is consumed
```
MatKuhr, line 397: I'd loved to use something more sophisticated, e.g. `WiremockNetworkTrafficListener`, but didn't find a good way to make that work. We can't listen for closed connections, since we keep the TCP connection open (on purpose of course).

I manually tested this fails, if we don't consume the response.

newtork: For now I changed the test to have an explicit check.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAdapter.java
```diff
@@ -148,35 +150,37 @@ private static String handleResponse( final HttpUriRequest request, final HttpRe
...
-                throw new DestinationAccessException(
-                    String
-                        .format(
-                            "Failed to get destinations: destination service returned HTTP status %s (%S) at '%s'.,",
-                            statusCode,
-                            reasonPhrase,
-                            requestUri));
-            }
+        Try<String> maybeBody = Try.of(() -> HttpEntityUtil.getResponseBody(response));
+        String logMessage = "Destination service returned HTTP status %s (%s)";
```
newtork, line 154: (Comment)
I'm not a fan of the dynamic string construction for log- and exception-messages. I know the format string `%s` is different from the log string template `{}` but I'm not sure the solution is the "best" practice yet.
It makes the code below complicated and during troubleshooting, we have to think twice, how a messages maps a corresponding piece of code.

newtork (approved):

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceAdapter.java
MatKuhr: Same, I felt this was a tradeoff between complex code vs. code duplication. Maybe at some point we switch to a more high-level HTTP client, where we don't have to re-implement this response handling every time 😄

PR #559, opened 2024-08-23
[DwC] Limit DwC Headers when Connecting to Megaclite

MatKuhr:
## Context

https://github.tools.sap/deploy-with-confidence/issues/issues/2552#issuecomment-7439955

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DwcHeaderProvider.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DwcHeaderProvider.java
@@ -3,6 +3,7 @@
  */
 package com.sap.cloud.sdk.cloudplatform.connectivity;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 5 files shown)


cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DwcHeaderProvider.java
```diff
@@ -26,6 +27,41 @@ public class DwcHeaderProvider implements DestinationHeaderProvider
...
+     * DwC headers are required.
+     *
+     * @return A header provider configured to only forward dwc headers required for destination service access.
+     */
+    static DwcHeaderProvider limitedHeaderProviderForDestinationAccess()
+    {
+        return new DwcHeaderProvider(
+            List
+                .of(
+                    "dwc-tenant",
```
MatKuhr, line 57: cross checked this list with [the repo](https://github.tools.sap/deploy-with-confidence/megaclite/search?q=dwcHeaders) and w/ Niklas

newtork: I think this is stretching the definition of singleton a bit far ^^

**Proposal:** keep using the `getInstance()` and define a destination property, which -when active- reduces the header count(?)

newtork: > The class never was a singleton

True, the constructor was public before.

newtork, line 57: (Major/Question)

I'm hesitant on this unmodifiable allow-list. When someone decides they'll need another `dwc-*` header propagated, that'll be impossible with the current approach, right :/

MatKuhr: Correct. In general, if we decide to limit the headers when accessing the destination or connectivity service this is the risk we are taking. The argument for why this would be acceptable is that, unless the megaclite implementation changes, this should not happen. And because using megaclite as proxy for the destination service is considered legacy and not recommended, changes should be rare.

newtork: > using megaclite as proxy for the destination service is considered legacy and not recommended
> changes should be rare

okay, this reduces my concerns.

newtork (approved):

newtork (approved):

PR #565, opened 2024-08-28
fix: [DevOps] Improve Test Reliability

MatKuhr:
## Context

This PR changes OData headers and parameters to follow a deterministic order.


newtork (approved):
This may affect customer unit tests as well, but I think this is acceptable.

Issue #557, opened 2024-08-22
DwC Auth Token not available (DwC + IAS)


jingweiz2017:
## Issue Description
This is an issue between DwC and SAP Cloud SDK's integration for "Principal Propagation" scenario when DwC is using IAS for authentication.

I think originally SAP Cloud SDK had approuter in mind. So it would expect an "Authorization" Header in the incoming request header. For DwC, it keep this dwc-ias-jwt token in the header in case needed. This dwc-ias-jwt token is the one I should use as "Authorization" header which later on can be passed onto remote service as user_token(x-user-token) for token exchange and authorization(principal propagation scenario). But SDK is not able to retrieve this dwc auth token and continue with the rest of the work.

After some research, I find this issue looks a lot like https://github.com/SAP/cloud-sdk-java/pull/337. However, I think that fix was targeting DwC xsuaa token key (dwc-jwt). For the case of DwC+IAS, the key would be "dwc-ias-jwt".

I have prepared a PR to enhance https://github.com/SAP/cloud-sdk-java/pull/337 so that dwc-ias-jwt will be covered as well.

Important information:
- Your code
```
public HttpDestination getBusinessLoggingDestination() {
    Try<Destination> tryGetDestination = DestinationAccessor.tryGetDestination(BUSINESS_LOGGING);

    if (tryGetDestination.isSuccess()) {
       HttpDestination httpDestination = tryGetDestination.get().asHttp();
       logger.debug("business log url: {}",httpDestination.getUri());
       return httpDestination;
    }

    try {
      var options = ServiceBindingDestinationOptions
        .forService(ServiceIdentifier.of(BUSINESS_LOGGING))
        .withOption(BtpServiceOptions.BusinessLoggingOptions.WRITE_API)
        .onBehalfOf(OnBehalfOf.NAMED_USER_CURRENT_TENANT)
        .build();
      HttpDestination destination = ServiceBindingDestinationLoader.defaultLoaderChain().getDestination(options);
      DestinationAccessor.prependDestinationLoader(
        (name, option) -> name.equals(BUSINESS_LOGGING) ?
        Try.of(() -> destination) :
        Try.failure(new DestinationNotFoundException(name))
      );
      return destination;
    } catch (Exception e) {
      String message = "Failed to get destination for business-logging caused by: " + e.getMessage();
      logger.error(message);
      throw new IllegalStateException(message, e);
    }
  }
```
- Expected outcome
The token exchange will be conducted successfully.
[...]

PR #568, opened 2024-08-29
fix: issue #557: DwC Auth Token not available (DwC + IAS)

jingweiz2017:
This PR is for fixing the issue SAP#557. The code change is mainly an enhancement on top of SAP#337 to include dwc-ias-jwt in header as auth token as well. dwc-ias-jwt is the key used for auth token when dwc is integrated with ias for authentication.

## Context
https://github.com/SAP/cloud-sdk-java/issues/557

The scenario is to propagate user principal from DwC+IAS to an xsuaa based reused service. 
Under such case, the authorization token attached to DwC headers used the key "dwc_ias_jwt". It is different from what approuter would attach (which is Authorization) or DwC+XSUAA would attach(which is dwc_jwt).

This fix is built on top of the fix [#337](https://github.com/SAP/cloud-sdk-java/pull/337). The fix #337 handles auth token retrieving for dwc_jwt for the DwC+xsuaa setup. This fix enhanced that auth token retrieving logic by also looking for "dwc_ias_jwt" header in case of DwC+IAS setup.

## Definition of Done
- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -150,7 +150,9 @@ public class DwcHeaderUtils
         return doGetNonEmptyDwcHeaderValue(container, key);
     }
 
-    private static String doGetNonEmptyDwcHeaderValue( @Nonnull final RequestHeaderContainer container, final String key )
+    private static
+        String
+        doGetNonEmptyDwcHeaderValue( @Nonnull final RequestHeaderContainer container, final String key )
     {
         return container
             .getHeaderValues(key)
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -126,12 +126,12 @@ public class DwcHeaderUtils
                 .tryGetHeaderContainer()
                 .getOrElseThrow(e -> new DwcHeaderNotFoundException("Unable to get current request headers.", e));
 
-        if(!container.containsHeader(DWC_JWT_HEADER) && !container.containsHeader(DWC_IAS_JWT_HEADER)) {
+        if( !container.containsHeader(DWC_JWT_HEADER) && !container.containsHeader(DWC_IAS_JWT_HEADER) ) {
             throw new DwcHeaderNotFoundException(
                 "Unable to find the " + DWC_JWT_HEADER + " or " + DWC_IAS_JWT_HEADER + " in header.");
         }
 
-        if(container.containsHeader(DWC_IAS_JWT_HEADER)) {
+        if( container.containsHeader(DWC_IAS_JWT_HEADER) ) {
             return doGetNonEmptyDwcHeaderValue(container, DWC_IAS_JWT_HEADER);
         }
 
--- a/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
+++ b/cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
@@ -121,21 +121,21 @@ public class DwcHeaderUtils
     @Nonnull
     public static String getDwcJwtOrThrow()
     {
-        final String errMsg = "Unable to find the " + DWC_JWT_HEADER + " or " + DWC_IAS_JWT_HEADER + " header value.";
         final RequestHeaderContainer container =
             RequestHeaderAccessor
                 .tryGetHeaderContainer()
-                .getOrElseThrow(e -> new DwcHeaderNotFoundException(errMsg, e));
+                .getOrElseThrow(e -> new DwcHeaderNotFoundException("Unable to get current request headers.", e));
 
-        if( !container.containsHeader(DWC_JWT_HEADER) && !container.containsHeader(DWC_IAS_JWT_HEADER) ) {
-            throw new DwcHeaderNotFoundException(errMsg);
+        if(!container.containsHeader(DWC_JWT_HEADER) && !container.containsHeader(DWC_IAS_JWT_HEADER)) {
+            throw new DwcHeaderNotFoundException(
+                "Unable to find the " + DWC_JWT_HEADER + " or " + DWC_IAS_JWT_HEADER + " in header.");
         }
 
-        if( container.containsHeader(DWC_IAS_JWT_HEADER) ) {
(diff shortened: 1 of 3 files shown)


cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
```diff
@@ -117,6 +121,20 @@ public static String getDwcPrincipalIdOrThrow()
...
+            RequestHeaderAccessor
+                .tryGetHeaderContainer()
+                .getOrElseThrow(e -> new DwcHeaderNotFoundException(errMsg, e));
+
+        if( !container.containsHeader(DWC_JWT_HEADER) && !container.containsHeader(DWC_IAS_JWT_HEADER) ) {
+            throw new DwcHeaderNotFoundException(errMsg);
+        }
+
+        if( container.containsHeader(DWC_IAS_JWT_HEADER) ) {
+            return getNonEmptyDwcHeaderValue(DWC_IAS_JWT_HEADER);
```
newtork, line 135: Thanks for your contribution!

Just a minor thing: The call to `RequestHeaderAccessor` is running up to 3x, when it could be 1x.

Maybe sth. like the following:
```java
    public static String getDwcJwtOrThrow()
    {
        final RequestHeaderContainer container =
            RequestHeaderAccessor
                .tryGetHeaderContainer()
                .getOrElseThrow(e -> new DwcHeaderNotFoundException("Failed to get current request headers.", e));

        return Stream
            .concat(
                container.getHeaderValues(DWC_IAS_JWT_HEADER).stream(),
                container.getHeaderValues(DWC_JWT_HEADER).stream())
            .filter(val -> !Strings.isNullOrEmpty(val))
            .findFirst()
            .orElseThrow(
                () -> new DwcHeaderNotFoundException(
                    "Unable to find the " + DWC_JWT_HEADER + " or " + DWC_IAS_JWT_HEADER + " header value."));
    }
```

newtork:
(You committed with two distinct users, please make sure both of them signed the "Contributor License Agreement")

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
jingweiz2017: Thanks for the review. I will make the call to RequestHeaderAccessor goes down to 1 time. 

However, we may want to identify a key of either DWC_JWT_HEADER or DWC_IAS_JWT_HEADER need to presented in the header. And when one of them is presented, its associated value should not be null/empty. 

For the string concatenation, due to DWC_JWT_HEADER and DWC_IAS_JWT_HEADER are static values, compiler will optimize the string concatenation as static string when compiled. But, it is more obvious for the usage of the error message in the sample code. I will take it.

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
```diff
@@ -117,7 +121,21 @@ public static String getDwcPrincipalIdOrThrow()
...
+                .tryGetHeaderContainer()
+                .getOrElseThrow(e -> new DwcHeaderNotFoundException("Unable to get current request headers.", e));
+
+        if(!container.containsHeader(DWC_JWT_HEADER) && !container.containsHeader(DWC_IAS_JWT_HEADER)) {
+            throw new DwcHeaderNotFoundException(
+                "Unable to find the " + DWC_JWT_HEADER + " or " + DWC_IAS_JWT_HEADER + " in header.");
+        }
+
+        if(container.containsHeader(DWC_IAS_JWT_HEADER)) {
+            return doGetNonEmptyDwcHeaderValue(container, DWC_IAS_JWT_HEADER);
```
MatKuhr, line 135: I think preferring the IAS JWT over the XSUAA JWT is in general the right choice. I am afraid there is a small chance this breaks existing use cases, where both XSUAA and IAS JWTs are present and the XSUAA JWT is used for something where the IAS JWT doesn't work equally.

But in most cases the IAS JWT should work equally and thus I would simply accept the risk.

cloudplatform/connectivity-dwc/src/test/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacadeTest.java
```diff
@@ -28,11 +29,22 @@ void testFacadeIsPickedUpAutomatically()
...
+        this.doTestSuccessfulAuthTokenRetrieval(DWC_JWT_HEADER);
+    }
+
+    @Test
+    void testSuccessfulIasAuthTokenRetrieval()
+    {
+        this.doTestSuccessfulAuthTokenRetrieval(DWC_IAS_JWT_HEADER);
+    }
+
+    void doTestSuccessfulAuthTokenRetrieval(String dwcHeaderKey)
```
MatKuhr, line 42: having a test that covers the IAS token taking precedence would be nice

MatKuhr (commented):
Thanks for the contribution! mostly LGTM, a bit of additional testing would be nice from my perspective

cloudplatform/connectivity-dwc/src/test/java/com/sap/cloud/sdk/cloudplatform/security/DwcAuthTokenFacadeTest.java
RockyZzzzz: added.

MatKuhr (dismissed):

MatKuhr (dismissed):

cloudplatform/connectivity-dwc/src/main/java/com/sap/cloud/sdk/cloudplatform/DwcHeaderUtils.java
```diff
@@ -129,6 +147,11 @@ private static String getNonEmptyDwcHeaderValue( @Nonnull final String key )
                 .tryGetHeaderContainer()
                 .getOrElseThrow(e -> new DwcHeaderNotFoundException("Unable to read the " + key + " header value.", e));
 
+        return doGetNonEmptyDwcHeaderValue(container, key);
+    }
+
+    private static String doGetNonEmptyDwcHeaderValue( @Nonnull RequestHeaderContainer container, final String key )
```
MatKuhr, line 153:
```suggestion
    private static String doGetNonEmptyDwcHeaderValue( @Nonnull final RequestHeaderContainer container, final String key )
```

MatKuhr (approved):

MatKuhr (approved):

jingweiz2017:
Thanks for your effort on this one. @MatKuhr & @newtork

MatKuhr:
@jingweiz2017 you are welcome! We released [version 5.12.0](https://sap.github.io/cloud-sdk/docs/java/release-notes#5120---september-03-2024) today and the improvement is included, so you should be able to make use of it, finally. Thanks again for your contribution, much appreciated 🥳

PR #629, opened 2024-11-05
OpenAPI: Added AI SDK Constructors

CharlesDuboisSAP:
## Context

Better UX for AI SDK generated clients.

### Feature scope:
 
- [x] Added an `aiSdkConstructor` option to replace all constructors with more convenient ones for AI SDK users

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


datamodel/openapi/openapi-generator/src/test/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGeneratorIntegrationTest.java
```diff
@@ -135,7 +150,7 @@ void generateDataModelForComparison( final TestCase testCase )
         final Path outputDirectory = getComparisonDirectory(testCase);
 
         assertThat(inputDirectory).exists().isReadable().isDirectory();
-        assertThat(inputDirectory).exists().isReadable().isDirectory();
+        assertThat(outputDirectory).exists().isReadable().isDirectory();
```
newtork, line 153: :o

datamodel/openapi/openapi-generator/src/main/resources/openapi-generator/mustache-templates/libraries/resttemplate/api.mustache
```diff
@@ -61,6 +66,27 @@ public class {{classname}} extends AbstractOpenApiService {
...
     }
+{{/aiSdkConstructor}}
+{{#aiSdkConstructor}}
+
+    /**
+     * Instantiates this API class to invoke operations on the {{appName}}
+     */
+    public {{classname}}()
+    {
+         super(new AiCoreService().client());
```
newtork, line 77: (Comment)

I'm a little concerned about adding **untested** and unmaintained code from [another project](https://github.com/SAP/ai-sdk-java) here. While I like the constructors, I would've preferred the following code to be supplied declaratively as generator property, like...

```xml
<noArgConstructor>new AiCoreService().client()</noArgConstructor>
<singleArgConstructorParam>@Nonnull final AiCoreDestination aiCoreDestination</singleArgConstructorParam>
<singleArgConstructor>aiCoreDestination.client()</singleArgConstructor>
```

But I see, that's not an urgent topic to discuss right now.

newtork (approved):

datamodel/openapi/openapi-generator/src/main/resources/openapi-generator/mustache-templates/libraries/resttemplate/api.mustache
MatKuhr: Yeah me too, in particular the workflow of changing related code is a bit scuffed. E.g. if we change `AiCoreService` to `AiCoreService2`:

1. Create PR in AI SDK to update Code
2. Create PR in Cloud SDK to update generator
3. Release Cloud SDK
4. Update Cloud SDK in AI SDK
5. Merge the original PR

But agreed, as long as we don't have a large amount of such interdependencies we can live with it for now

datamodel/openapi/openapi-generator/src/test/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGeneratorIntegrationTest.java
```diff
@@ -84,6 +97,7 @@ private enum TestCase
         String methodBuilder = null;
         String methodBuild = null;
         String constructorVisibility = null;
+        String aiSdkConstructor = null;
```
MatKuhr, line 100: why is this String and used with `"true"` and `null`?

```suggestion
        boolean aiSdkConstructor = false;
```

CharlesDuboisSAP: `String` and not `boolean` because it is a maven property.
`null` instead of `"false"` because the `true/false` is a lie, the real activation of this maven property is `absent/present` , no matter the value of the property.

PR #634, opened 2024-11-08
Proper API to change TLS upgrade flag and store it on instance level, non-static

newtork:


cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultApacheHttpClient5Factory.java
```diff
@@ -175,16 +170,22 @@ private RequestConfig getRequestConfig( @Nullable final HttpDestinationPropertie
...
+            case INTERNET:
+                if( destination == null ) {
+                    return true;
+                }
+                if( destination.getTlsVersion().isDefined() ) {
+                    return false;
+                }
+                return !destination.getProxyType().contains(ProxyType.ON_PREMISE);
+            default:
+                throw new IllegalStateException("Unknown TLS upgrade setting: " + tlsUpgrade);
```
MatKuhr, line 187: Thanks to Java 17, can be simplified to:

```suggestion
        return switch( tlsUpgrade ) {
            case ENABLED -> true;
            case DISABLED -> false;
            case INTERNET -> {
                if( destination == null ) {
                    yield true;
                }
                if( destination.getTlsVersion().isDefined() ) {
                    yield false;
                }
                yield !destination.getProxyType().contains(ProxyType.ON_PREMISE);
            }
        };
```

MatKuhr (approved):
LGTM, minor suggestions

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultApacheHttpClient5Factory.java
newtork: yield... :o

PR #636, opened 2024-11-11
Stabilized ServiceBinding classes

CharlesDuboisSAP:
Related ticket:
https://github.com/SAP/cloud-sdk-java-backlog/issues/461


MatKuhr (approved):
LGTM. There are arguably a few more classes related to service bindings, namely the dwc and oauth classes. I think it's safe to stabilize DwC. For OAuth, I would only update `OAuth2PropertySupplier` and `OAuth2ServiceBindingDestinationLoader`. The rest is rather niche and maybe doesn't need to be stable right now..

MatKuhr (approved):

PR #652, opened 2024-11-28
Update Code Generator for AI SDK

MatKuhr:
## Context

Supports this PR: SAP/ai-sdk-java#201

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


datamodel/openapi/openapi-generator/src/main/resources/openapi-generator/mustache-templates/libraries/resttemplate/api.mustache
```diff
@@ -74,17 +73,17 @@ public class {{classname}} extends AbstractOpenApiService {
      */
     public {{classname}}()
     {
-         super(new AiCoreService().client());
+         super(new AiCoreService().getApiClient());
```
newtork, line 76: (Hindsight)

This would not have been necessary if we had considered parameterized constructors:
https://github.com/SAP/cloud-sdk-java/pull/629#discussion_r1830676355

Why not do this now?

MatKuhr: true, on the other hand I'd hope we don't have to do frequent changes to this, and if we do, there is the risk the parameters don't exactly fit what we need 😅

No strong opinion though, we also don't have time pressure to implement this, can always be done later on if we like

CharlesDuboisSAP (approved):

Issue #659, opened 2024-12-07
HttpClient doesn't configure the Spring RestTemplate properly.


hdva2502:
### Describe the Bug

Hi colleagues,

when use the 
ApacheHttpClient5Accessor.getHttpClient(destination)
feature the retrieve a httpclient from an http destination (internet and onPremise).

Then we use the retrieved http client to create a Spring RestTemplate like this
new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient));

This RestTemplate is then used for API calls, also http HEAD to the backend to check connectivity.
See also the issues you recently resolved: [#632](https://github.com/SAP/cloud-sdk-java/issues/632), [#602](https://github.com/SAP/cloud-sdk-java/issues/602).

The http client is configured correctly (see first screenshot), however the request is not configured corretly (see second screenshot)

If you look deeper into HttpComponentsClientHttpRequestFactory => [createRequest](https://github.com/spring-projects/spring-framework/blob/ac5c8adb9830939e2329f1e16727c522a172c7c8/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.java#L209) you see that the relevant method is [createRequestConfig](https://github.com/spring-projects/spring-framework/blob/ac5c8adb9830939e2329f1e16727c522a172c7c8/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.java#L250) where your HttpClientWrapper is not used for configuration ... because it doesn't implement Configurable ... however the HttpClient you wrap does.

So imho it would help if ApacheHttpClient5Wrapper would implement Configurable and would delegate the getConfig to the inner httpClient.

What do you think?

Thanks in advance and best regards,
Andreas

### Steps to Reproduce

see above

### Expected Behavior

see above

### Screenshots

![image](https://github.com/user-attachments/assets/4c08a02a-14fc-4d12-94b5-2be80496b208)
![image](https://github.com/user-attachments/assets/f26de4e2-5b1f-4ada-b8b5-c021fcb58c62)

### Used Versions

- Java: 17
- SAP Cloud SDK version: 5.14.0
- Spring Boot: 3.3.6

### Code Examples

```java
// Your code here
```

### Log File

<details><summary>Log file</summary>
...
</details>

### Affected Development Phase

Production

### Impact

Blocked

CharlesDuboisSAP:
Hello @hdva2502 

I'm currently investigating the issue.
[...]

PR #662, opened 2024-12-10
Fix: TLS Upgrade config propagates to RestTemplate

CharlesDuboisSAP:
## Context

#659

### Feature scope:
 
- [x] ApacheHttpClient5Wrapper RequestConfig propagates to RestTemplate

## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultApacheHttpClient5Factory.java
+++ b/cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultApacheHttpClient5Factory.java
@@ -82,7 +82,7 @@ class DefaultApacheHttpClient5Factory implements ApacheHttpClient5Factory
             return httpClient;
         }
 
-        return new ApacheHttpClient5Wrapper(httpClient, destination);
+        return new ApacheHttpClient5Wrapper(httpClient, destination, getRequestConfig(destination));
     }
 
     @Nonnull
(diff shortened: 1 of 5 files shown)


cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultApacheHttpClient5Factory.java
```diff
@@ -82,7 +82,7 @@ public HttpClient createHttpClient( @Nullable final HttpDestinationProperties de
             return httpClient;
         }
 
-        return new ApacheHttpClient5Wrapper(httpClient, destination);
+        return new ApacheHttpClient5Wrapper(httpClient, destination, getRequestConfig(destination));
```
MatKuhr, line 85: this causes `getRequestConfig` to be computed twice, right? Is there not a way we can get the already built config from the `httpClient` object? (mostly to prevent this thing is modified elsewhere without changes being reflected here)

CharlesDuboisSAP: fixed

MatKuhr (approved):
LGTM

We could consider to apply the same change to the v4 Apache client. However, RestTemplate requires the v5 Apache client now, so we don't have the same test case. So we would have to find something else to test or just leave it untested. So maybe we just leave it as-is until we see a use case.

PR #685, opened 2025-01-06
Fix copyright header `2024` -> `2025`

newtork:


jjtang1985 (changes requested):
I would even remove the copyright headers as they are not needed.

PR #695, opened 2025-01-15
fix: [OpenAPI] Fix Enum Discriminator

MatKuhr:
## Context

Fixes a compilation issue in case `enum` is used together with a discriminator. 

Background: This is a frequent use case, as people want to explicitly limit the allowed values for the discriminator. In OpenAPI 3.0, this is done by declaring an enum with just one value.

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorIntegrationTest/oneof-interfaces-enabled/output/test/OneOfWithDiscriminator.java
```diff
@@ -51,6 +51,5 @@
 })
 
 public interface OneOfWithDiscriminator  {
-    String getSodaType();
```
newtork, line 54: (Minor)

Now thinking about it again, we could also solve it via...?
```diff
- String getSodaType();
+ Object getSodaType();
```

Question is whether we want users to cast objects based on `instanceof` or `getProperty()`.

newtork (dismissed):
It's a breaking change, but since it's behind a feature toggle, presumably that's acceptable.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorIntegrationTest/oneof-interfaces-enabled/output/test/OneOfWithDiscriminator.java
MatKuhr: Good idea. While I think in practice it's easier to `f instanceof Foo foo`, this leaves the other option open. Updated the PR.

newtork (approved):

Issue #704, opened 2025-01-28
Non-changing update of an entity value of type BigDecimal leads to: itemToUpdate.getChangedFields().size() >0


itmstr01:
### Describe the Bug

I read an Item-Object of a Bom. The value of componentQuantity is "**1**" as a BigDecimal.
I updated the value of componentQuantity with "**1.0**", wich should be mathematically the same!
But now the expression **itemToUpdate.getChangedFields().size() >0** is true!
The reason is this behavior: **new BigDecimal("1.0").equals(new BigDecimal("1")) == false**
Sadly the VdmObject.getChangedFields() uses the equals-method for delta-detection.
With BigDecimals you have to use "compareTo() == 0" for equality-testing.
This behavior can lead to performance issues and huge network traffic.

### Steps to Reproduce

See bug description

### Expected Behavior

**itemToUpdate.getChangedFields().size() == 0**

### Used Versions

- Java and Maven version via `mvn --version`: 
```
Apache Maven 3.9.9 (8e8579a9e76f7d015ee5ec7bfcdc97d260186937)
Maven home: C:\Program Files\JetBrains\IntelliJ IDEA 2024.1.4\plugins\maven\lib\maven3
Java version: 17.0.13, vendor: Amazon.com Inc., runtime: C:\Java\CorrettoJdk17.0.13_11
Default locale: de_DE, platform encoding: UTF-8
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
```
- SAP Cloud SDK version: 5.15.0
- Spring Boot or CAP version: no
- <details><summary>Dependency tree via `mvn dependency:tree`</summary>
    
    ```
    not needed because of obvious behavior.
    ```
  </details>

### Code Examples

```
not needed because of obvious behavior.
```

### Log File

<details><summary>Log file</summary>
...
</details>

### Affected Development Phase

Release

### Impact

Impaired

### Timeline

Needed end of Feb.

Jonas-Isr:
Hi @itmstr01,

what is your use case for this? Can't you do the check/conversion before the update?

Best,
Jonas

itmstr01:
First of all: Can you guarantee that the BigDecimal value within the entity always contains no decimal places if it contains an integer value? (and no extra zeros on fractional values)  That would be the prerequisite for solving the problem on our side.

[...]

PR #714, opened 2025-02-10
Fix OData `VDMObject.getChangedFields()` for `BigDecimal`

CharlesDuboisSAP:
## Context
 

#704.

A `BigDecimal` field which is updated from `1` to `1.0` will not be considered as changed anymore.

### Feature scope:

 
- [x] OData v2
- [x] OData v4
- [x] Drive-by Java 23 compilation fixes

## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/datamodel/odata-v4/odata-v4-generator/src/main/java/com/sap/cloud/sdk/datamodel/odatav4/generator/PreparedEntityBluePrint.java
+++ b/datamodel/odata-v4/odata-v4-generator/src/main/java/com/sap/cloud/sdk/datamodel/odatav4/generator/PreparedEntityBluePrint.java
@@ -24,27 +24,6 @@ final class PreparedEntityBluePrint
     private final JDefinedClass entityClass;
 
     /**
-     * The more concrete sub-interface of the {@link com.sap.cloud.sdk.datamodel.odata.helper.EntitySelectable
-     * EntitySelectable} interface for this entity.
-     */
-    /*@Nullable // in case of "POJO only"
-    private final JDefinedClass selectableInterface;*/
-
-    /**
-     * The concrete sub-class of the {@link com.sap.cloud.sdk.datamodel.odata.helper.EntityLink EntityLink} class for
-     * this entity used for one-to-many links.
-     */
-    /*@Nullable // in case of "POJO only"
-    private final JDefinedClass entityOneToManyLinkClass;
-    */
-    /**
-     * The concrete sub-class of the {@link com.sap.cloud.sdk.datamodel.odata.helper.EntityLink EntityLink} class for
-     * this entity used for one-to–one links.
-     */
-    /*@Nullable // in case of "POJO only"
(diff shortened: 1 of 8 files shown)


datamodel/odata-v4/odata-v4-generator/src/main/java/com/sap/cloud/sdk/datamodel/odatav4/generator/PreparedEntityBluePrint.java
```diff
@@ -23,27 +23,6 @@ final class PreparedEntityBluePrint
...
-    /*@Nullable // in case of "POJO only"
-    private final JDefinedClass entityOneToManyLinkClass;
-    */
-    /**
-     * The concrete sub-class of the {@link com.sap.cloud.sdk.datamodel.odata.helper.EntityLink EntityLink} class for
-     * this entity used for one-to–one links.
-     */
-    /*@Nullable // in case of "POJO only"
-    private final JDefinedClass entityOneToOneLinkClass;*/
-
```
CharlesDuboisSAP, line 46: Drive-by Java 23 fixes

datamodel/odata/odata-generator/src/main/java/com/sap/cloud/sdk/datamodel/odata/generator/DeprecationUtils.java
```diff
@@ -56,16 +56,6 @@ static void createBasicServiceInterfaceField(
...
-    /**
-     * Adds javadoc and annotations regarding states such as deprecation if necessary.
-     *
-     * @param affectedClass
-     *            The class to add information to.
-     * @param service
-     *            The service to take the status information from.
-     *
-     */
-
```
CharlesDuboisSAP, line 68: Drive-by Java 23 fixes

pom.xml
```diff
@@ -759,6 +759,12 @@
 							<arg>-Xlint:all</arg>
 							<arg>-Xlint:-processing</arg>
 						</compilerArgs>
+						<annotationProcessorPaths>
+							<annotationProcessorPath>
+								<groupId>org.projectlombok</groupId>
+								<artifactId>lombok</artifactId>
+							</annotationProcessorPath>
+						</annotationProcessorPaths>
```
CharlesDuboisSAP, line 767: Drive-by Java 23 fixes

MatKuhr (changes requested):
We can't make this change as this might break existing use cases. There is no gaurantee servers will treat `1.0` and `1` equally

CharlesDuboisSAP:
> We can't make this change as this might break existing use cases. There is no gaurantee servers will treat `1.0` and `1` equally

I don't mind this behaviour change. Differentiating between 1 and 1.0 is not a real use-case anyway. Plus as the customer said it will cut on bandwidth usage.

PR #717, opened 2025-02-11
Update dependabot.yaml

CharlesDuboisSAP:


newtork (approved):
I assumed `day: [...]` would automatically imply `interval: weekly`

PR #735, opened 2025-02-26
feat: OpenAPI Generator to drop the redundant `isIsBoolean()` method prefix

newtork:
## Context
 

We encountered this in AI SDK, it would be nice to fix. The generated method name is a little embarrassing.

image

### Feature scope:

* Feature toggle (default: disabled)
* Drop the double "is"-prefix.
  * There may be side effects though.

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


Jonas-Isr (approved):
LGTM

PR #827, opened 2025-06-05
feat: Extend error message for 401 errors

Jonas-Isr:
## Context

SAP/cloud-sdk-java-backlog#474

There have been user issues where the underlying problem is that the SaaS dependencies were not setup correctly, resulting in a 401 - Bad Credentials error when the user was doing a OAuth2 token request.

This PR extends the relevant error message in this case to include a hint that the reason for the 401 might be a wrong setup of the SaaS registry.

### Feature scope:

 
- [x] error message is extended in the relevant case

## Definition of Done

- [x] Documentation updated: https://github.com/SAP/cloud-sdk/pull/2116
- [ ] ~Release notes updated~

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
@@ -205,7 +205,7 @@ class OAuth2Service
         if( e instanceof OAuth2ServiceException
             && ((OAuth2ServiceException) e).getHttpStatusCode().equals(401)
             && tenant != null ) {
-            String extension;
+            final String extension;
             if( serviceIdentifier != null ) {
                 extension =
                     " In case you are accessing a multi-tenant BTP service on behalf of a subscriber tenant, ensure that the service instance (here, of the "
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
@@ -36,6 +36,7 @@ import com.sap.cloud.security.client.HttpClientFactory;
 import com.sap.cloud.security.config.ClientIdentity;
 import com.sap.cloud.security.token.Token;
 import com.sap.cloud.security.xsuaa.client.DefaultOAuth2TokenService;
+import com.sap.cloud.security.xsuaa.client.OAuth2ServiceException;
 import com.sap.cloud.security.xsuaa.client.OAuth2TokenResponse;
 import com.sap.cloud.security.xsuaa.client.OAuth2TokenService;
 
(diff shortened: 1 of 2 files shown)


cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -190,7 +190,17 @@ private OAuth2TokenResponse executeClientCredentialsFlow( @Nullable final Tenant
...
-            .getOrElseThrow(e -> new TokenRequestFailedException("Failed to resolve access token.", e));
+            .getOrElseThrow(e -> buildException(e, tenant));
+    }
+
+    private TokenRequestFailedException buildException(@Nonnull final Throwable e, @Nullable final Tenant tenant )
+    {
+        String msg = "Failed to resolve access token.";
+        //        In case where tenant is subscriber, and we get 401 error, add hint to error message.
+        if( e.getMessage().contains("Http status code 401") && tenant != null ) {
+            msg += " This might be due to missing or wrongly set up dependencies in your SaaS registry.";
```
MatKuhr, line 201:
```suggestion
            msg += "In case you are accessing a multi-tenant BTP service, ensure that the service instance is declared as dependency to SaaS Provisioning Service or Subscription Manager (SMS) and subscribed for the current tenant.";
```

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -190,7 +190,17 @@ private OAuth2TokenResponse executeClientCredentialsFlow( @Nullable final Tenant
...
                         false))
-            .getOrElseThrow(e -> new TokenRequestFailedException("Failed to resolve access token.", e));
+            .getOrElseThrow(e -> buildException(e, tenant));
+    }
+
+    private TokenRequestFailedException buildException(@Nonnull final Throwable e, @Nullable final Tenant tenant )
+    {
+        String msg = "Failed to resolve access token.";
+        //        In case where tenant is subscriber, and we get 401 error, add hint to error message.
+        if( e.getMessage().contains("Http status code 401") && tenant != null ) {
```
MatKuhr, line 200: this check is rather fragile, don't we have anything better?
If not, maybe ignore case and check for `contains("401") && contains("status code")`

the message may easily change to `code: 401` and break this check

MatKuhr, line 200: please double check that tenant != null is really sufficient. We need to check that tenant != provider tenant.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
MatKuhr: Keep in mind: we use this code for all OAuth requests, only some of which may target BTP services.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
Jonas-Isr: Yes, that is something where I am still not 100% sure as well. Maybe we can have a short look at this together next week?

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -88,6 +91,8 @@ class OAuth2Service
     @Nonnull
     @Getter( AccessLevel.PACKAGE )
     private final ResilienceConfiguration resilienceConfiguration;
+    @Nullable
+    private ServiceIdentifier serviceIdentifier;
```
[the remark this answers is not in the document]
Jonas-Isr: I decided to not include the serviceIdentifier information in the error message anymore. The cost-benefit ratio is not positive in my opinion given the different arguments against it and the little improvement it provides for the user.

MatKuhr (approved):
lgtm

PR #808, opened 2025-05-19
chore: Migrate to commons-configuration2

CharlesDuboisSAP:
## Context
 

https://github.com/SAP/cloud-sdk-java/issues/807

We previously used [commons-configuration](https://mvnrepository.com/artifact/commons-configuration/commons-configuration) but the new version is at [commons-configuration2](https://mvnrepository.com/artifact/org.apache.commons/commons-configuration2)

### Feature scope:

 
- [ ] Migrate to commons-configuration2

## Definition of Done

- [ ] Functionality scope stated & covered
- [ ] Tests cover the scope above
- [ ] ~Error handling created / updated & covered by the tests above~
- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

PR #848, opened 2025-07-01
fix: Compile with Java 23 (language level 17)

CharlesDuboisSAP:
Fixed changes from #808 that added a compilation warning in Java 23
```
[WARNING] ServiceNameMappings.java:[51,29] possible 'this' escape before subclass is fully initialized
[WARNING] ServiceNameMappings.java:[114,30] previous possible 'this' escape happens here via invocation
```


Jonas-Isr (dismissed):
LGTM

PR #850, opened 2025-07-01
fix: (odata-client) Move disable-buffer toggle from result object to request builder

newtork:
## Context
 

SAP/cloud-sdk-java-backlog#ISSUENUMBER.

### Feature scope:

 
- [ ] Task1 
- [ ] Task2 
  - [ ] SubTask1 

## Definition of Done

- [ ] Functionality scope stated & covered
- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestBatch.java
+++ b/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestBatch.java
@@ -93,6 +93,7 @@ public class ODataRequestBatch extends ODataRequestGeneric
         this.uuidProvider = uuidProvider;
         this.batchUuid = uuidProvider.get();
         this.headers.remove(HttpHeaders.ACCEPT); // batch request does not require Accept header
+        this.bufferStrategy = ResponseBufferStrategy.DISABLED;
     }
 
     @Nonnull
--- a/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestGeneric.java
+++ b/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestGeneric.java
@@ -89,6 +89,12 @@ public abstract class ODataRequestGeneric implements ODataRequestExecutable
     @Setter
     protected CsrfTokenRetriever csrfTokenRetriever;
 
+    /**
+     * The response buffer strategy to use for this request.
+     */
+    @Nonnull
+    ResponseBufferStrategy bufferStrategy = ResponseBufferStrategy.ENABLED;
+
     ODataRequestGeneric(
         @Nonnull final String servicePath,
         @Nonnull final ODataResourcePath resourcePath,
@@ -141,6 +147,14 @@ public abstract class ODataRequestGeneric implements ODataRequestExecutable
     public void addListener( @Nonnull final ODataRequestListener listener )
     {
         listeners.add(listener);
+    }
+
+    /**
+     * Disable pre-buffering of http response entity.
+     */
+    public void disableHttpResponseBuffering()
+    {
+        bufferStrategy = ResponseBufferStrategy.DISABLED;
     }
 
     /**
@@ -229,6 +243,7 @@ public abstract class ODataRequestGeneric implements ODataRequestExecutable
     {
         return Try
             .ofSupplier(httpOperation)
+            .map(bufferStrategy.getHandler())
             .map(response -> new ODataRequestResultGeneric(this, response, httpClient))
             .andThenTry(ODataHealthyResponseValidator::requireHealthyResponse);
     }
--- a/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultGeneric.java
+++ b/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultGeneric.java
@@ -22,9 +22,7 @@ import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.StatusLine;
 import org.apache.http.client.HttpClient;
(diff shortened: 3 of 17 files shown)


datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultFactory.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Strategy that buffers the response by creating a copy of it.
+     */
+    ODataRequestResultFactory WITH_BUFFER = ( oDataRequest, httpResponse, httpClient ) -> {
+        final BasicHttpResponse copy = new BasicHttpResponse(httpResponse.getStatusLine());
+        Option.of(httpResponse.getLocale()).peek(copy::setLocale);
+        Option.of(httpResponse.getAllHeaders()).peek(copy::setHeaders);
+        Option
+            .of(httpResponse.getEntity())
+            .peek(entity -> Try.run(() -> copy.setEntity(new BufferedHttpEntity(entity))));
```
MatKuhr, line 34: This currently drops an important warning log, please add it back in. It's highly likely that any IOException here will cause a failure later, and without the log the exception may be lost.

Unfortunately, logging is always annoying in interfaces, so maybe better to move this to class, e.g. with an inner interface?

datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestRead.java
```diff
@@ -134,4 +135,15 @@ public ODataRequestResultGeneric execute( @Nonnull final HttpClient httpClient )
                 : tryExecuteWithCsrfToken(httpClient, request::requestGet);
         return result.get();
     }
+
+    /**
+     * Disable pre-buffering of http response entity.
+     */
+    @Beta
+    @Nonnull
+    public ODataRequestResultResource.Executable withoutResponseBuffering()
```
MatKuhr, line 144: any reason not to offer this on the parent class? 

Ah damn it, now that I think about it I guess the issue is that this would then also appear on `batch` where the return type wouldn't match? 🫠

Maybe this is doable with overriding the return type in the batch case with the more specific Multipart result?

datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultFactory.java
```diff
@@ -0,0 +1,42 @@
...
+{
+    /**
+     * Strategy that does not buffer the response.
+     */
+    ODataRequestResultFactory WITHOUT_BUFFER = ODataRequestResultResource::new;
+
+    /**
+     * Strategy that buffers the response by creating a copy of it.
+     */
+    ODataRequestResultFactory WITH_BUFFER = ( oDataRequest, httpResponse, httpClient ) -> {
```
MatKuhr, line 28: previously, the implementation was `synchronized`. Are we confident this is no longer needed?

datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestRead.java
```diff
@@ -134,4 +135,15 @@ public ODataRequestResultGeneric execute( @Nonnull final HttpClient httpClient )
...
+    {
+        requestResultFactory = ODataRequestResultFactory.WITHOUT_BUFFER;
+        return httpClient -> (ODataRequestResultResource) this.execute(httpClient);
```
MatKuhr, line 147: This effectively means that `withoutResponseBuffering()` will be the last method before executing, but I think that is an acceptable small limitation

MatKuhr (commented):
productive code largely LGTM, didn't look at tests yet. I like the non-breaking introduction of `AutoClosable`, even though it comes at the cost of some new API.

If we want to reduce new API a bit, we could consider adding `executeOpen()` to `ODataRequestExecutable`. That would have the downside that this is also on Batch, where then both `execute` and `executeOpen` behave the same (since changing `execute` to now start buffering would be a breaking change 🙁)

But maybe we can somewhere make clearer, that this solution not only buffers, but eagerly buffers. It's good for understanding the importance of having both, as you pointed out 😉

MatKuhr (commented):
tests LGTM 👍🏻

rpanackal (commented):
I have reviewed low level api changes. Overall looks good to me. 

Same concern as @MatKuhr on logging failed buffering of http entity.

That said, I haven't previously looked at this module deeply. So not sure if I missed anything.

datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultFactory.java
newtork: Thanks for pointing it out!
I missed this regression.

datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/request/ODataRequestResultFactory.java
```diff
@@ -29,9 +32,15 @@ interface ODataRequestResultFactory
         final BasicHttpResponse copy = new BasicHttpResponse(httpResponse.getStatusLine());
         Option.of(httpResponse.getLocale()).peek(copy::setLocale);
         Option.of(httpResponse.getAllHeaders()).peek(copy::setHeaders);
+
+        final Logger log = getLogger(ODataRequestResultFactory.class);
         Option
             .of(httpResponse.getEntity())
-            .peek(entity -> Try.run(() -> copy.setEntity(new BufferedHttpEntity(entity))));
+            .onEmpty(() -> log.warn("HTTP response entity is empty."))
```
MatKuhr, line 39: Wouldn't this produce a warning on any 204 response, i.e. over-eager?

PR #851, opened 2025-07-03
Introducing Transparent Proxy Builder

isevdalinosap:
## Context
 

This PR adds support for consuming systems in Cloud SDK applications leveraging the Transparent Proxy for new destination features and on-premise connectivity.

### Feature scope:

 
- [x] Implement TransparentProxyDestination class 

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


cloudplatform/cloudplatform-connectivity/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,151 @@
...
+                .subjectTokenType("subjectTokenType")
+                .actorToken("actorToken")
+                .actorTokenType("actorTokenType")
+                .redirectUri("redirectUri")
+                .codeVerifier("codeVerifier")
+                .chainName("chainName")
+                .chainVarSubjectToken("chainVarSubjectToken")
+                .chainVarSubjectTokenType("chainVarSubjectTokenType")
+                .chainVarSamlProviderDestinationName("chainVarSamlProviderDestName")
+                .property(DestinationProperty.URI, "https://example.com")
```
newtork, line 125: (Question)

I don't understand this. I thought destinations referenced via "Transparent Proxy" would rely on relative URIs not absolute addresses.

newtork (commented):
For further review purpose please find the official documentation on _Transparent Proxy_
https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/using-transparent-proxy

Please find our current explanation in _SAP Cloud SDK_
https://sap.github.io/cloud-sdk/docs/java/environments/kubernetes-kyma#executing-requests

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
...
+    public URI getUri()
+    {
+        return URI.create(baseProperties.get(DestinationProperty.URI).get());
+    }
+
+    @Nonnull
+    @Override
+    public Collection<Header> getHeaders( @Nonnull URI requestUri )
+    {
+        return customHeaders;
```
MatKuhr, line 92: We have to evaluate how this should interact with [static header providers](https://sap.github.io/cloud-sdk/docs/java/features/connectivity/http-destinations#about-headerproviders). Probably, those need to be loaded and evaluated.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
...
+    private TransparentProxyDestination(
+        @Nonnull final DestinationProperties baseProperties,
+        @Nullable final List<Header> customHeaders,
+        @Nonnull final ComplexDestinationPropertyFactory destinationPropertyFactory )
+    {
+        this.baseProperties = baseProperties;
+        this.customHeaders =
+            customHeaders != null ? ImmutableList.<Header> builder().addAll(customHeaders).build() : ImmutableList.of();
+
+        cachedProxyConfiguration = destinationPropertyFactory.getProxyConfiguration(baseProperties);
```
MatKuhr, line 77: I'm not sure if we should set the transparent proxy as a proxy in the HTTP client, or if we should leave it empty.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
...
+        return null;
+    }
+
+    @Nullable
+    @Override
+    public AuthenticationType getAuthenticationType()
+    {
+        return null;
+    }
+
```
MatKuhr, line 142: These values will likely depend on how connections to the transparent proxy will be secured, and thus might need to be implemented.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
...
+    public Option<String> getTrustStorePassword()
+    {
+        return null;
+    }
+
+    /**
+     * Builder class to allow for easy creation of an immutable {@code TransparentProxyDestination} instance.
+     */
+    @Accessors( fluent = true, chain = true )
+    public static class Builder
```
MatKuhr, line 168: As Alex pointed out, the builder should enforce required properties

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
...
+         * @return A fully instantiated {@code TransparentProxyDestination}.
+         */
+        @Nonnull
+        public TransparentProxyDestination build()
+        {
+            if( !builder.get(DestinationProperty.URI).isDefined() ) {
+                this.property(DestinationProperty.URI, TRANSPARENT_PROXY_GATEWAY);
+            }
+
+            TenantAccessor.tryGetCurrentTenant().onSuccess(tenant -> {
```
MatKuhr, line 582: I think we have two options on how to handle tenancy:

1. Determine the tenant upon building the destination
2. Determine the tenant when using the destination

Currently, strategy (1) is applied, but I think (2) better fits the Cloud SDK APIs.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
...
+                if( !containsHeader(TENANT_ID_HEADER_KEY) && !tenant.getTenantId().isEmpty() ) {
+                    headers.add(new Header(TENANT_ID_HEADER_KEY, tenant.getTenantId()));
+                }
+
+                final DefaultTenant defaultTenant = (DefaultTenant) tenant;
```
MatKuhr, line 587: casting requires an instance of check

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,604 @@
+package com.sap.cloud.sdk.cloudplatform.connectivity;
```
MatKuhr, line 1: Since this functionality is still specific to the destination service, I think this would be more suitable under `connectivity-destination-service` module.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
isevdalinosap: I believe from transparent proxy point of view only destination name is required. The url is designed to be defaulted and the namespace is embedded into the url. I have added enforcement of the destination name.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
MatKuhr: I think it's up to the user how to authenticate the connection to TP. Currently, TP is only running in K8S, where this is typically done with sidecars. So in that case, users are unlikely to need this functionality. However, TP is also coming to CF, and users may decide to authenticate differently, or want to use some of these methods for other purposes.

For now, I think it's okay to not implement actual logic, leaving these cases unsupported until users request it. 

However, returning `null` when the method is declared `Option<..>` should never be done, so instead, please return `Option.none()` and change to `@Nonnull`

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
isevdalinosap: Agreed, moved all the logic into `connectivity-destination-service` module.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
isevdalinosap: Changed all the methods in question to return `Option.none()` and be annotated with `@Nonnull`.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+    @Override
+    public Option<String> getKeyStorePassword()
+    {
+        return Option.none();
+    }
+
+    @Override
+    public boolean isTrustingAllCertificates()
+    {
+        return true;
```
MatKuhr, line 192:
```suggestion
        return false;
```

I think this should default to false? In the (admittedly unlikely) case the transparent proxy is exposed under a `https` URL, we should not trust all certificates

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+
+import io.vavr.control.Option;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+import lombok.experimental.Delegate;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Immutable implementation of the {@link HttpDestination} interface.
```
MatKuhr, line 28: It'd be nice to have a bit of context about the transparent proxy here, maybe with a link to help.sap.com for more information

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+        /**
+         * Adds the given key-value pair to the destination to be created. This will overwrite any property already
+         * assigned to the key.
+         *
+         * @param key
+         *            The key to assign a property for.
+         * @param value
+         *            The property value to be assigned.
+         * @return This builder.
+         * @since 5.0.0
```
MatKuhr, line 291:
```suggestion
         * @since 5.22.0
```

and apply the same everywhere for consistency

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+    }
+
+    boolean containsHeader( final Collection<Header> headers, final String headerName )
+    {
+        return headers.stream().anyMatch(h -> h.getName().equalsIgnoreCase(headerName));
+    }
+
+    private List<Header> getHeadersFromHeaderProviders( @Nonnull final URI requestUri )
+    {
+        final List<DestinationHeaderProvider> aggregatedHeaderProviders = new ArrayList<>();
```
MatKuhr, line 137: Not ideal to duplicate this code from the `DefaultHttpDestination`. While the two implementations are in separate Maven modules, they share the same Java package, so I think using a package-private helper method for this may be possible?

@newtork any opinions? I'm just worried we'd forget to update this code if we ever change `DefaultHttpDestination`...

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+         */
+        @Nonnull
+        public B tenantId( @Nonnull final String tenantId )
+        {
+            return header(new Header(TENANT_ID_HEADER_KEY, tenantId));
+        }
+
+        /**
+         * Adds a token service tenant header to the destination. See
+         * https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/technical-user-propagation
```
MatKuhr, line 393: Is this link really correct? Isn't this rather to control placeholders in token URLs? I.e. to be used with `tokenServiceURLType = common`?

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+         * @return This builder.
+         */
+        @Nonnull
+        public B header( @Nonnull final String headerName, @Nonnull final String headerValue )
+        {
+            return header(new Header(headerName, headerValue));
+        }
+
+        /**
+         * Adds a tenant subdomain header to the destination. See
```
MatKuhr, line 364: For the tenant and authorization headers it'd be good to mention here that they are set automatically by the Cloud SDK per-request, if left unset here.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+    private static final String SUBJECT_TOKEN_TYPE_HEADER_KEY = "x-subject-token-type";
+    private static final String ACTOR_TOKEN_HEADER_KEY = "x-actor-token";
+    private static final String ACTOR_TOKEN_TYPE_HEADER_KEY = "x-actor-token-type";
+    private static final String REDIRECT_URI_HEADER_KEY = "x-redirect-uri";
+    private static final String CODE_VERIFIER_HEADER_KEY = "x-code-verifier";
+    private static final String CHAIN_NAME_HEADER_KEY = "x-chain-name";
+    private static final String CHAIN_VAR_SUBJECT_TOKEN_HEADER_KEY = "x-chain-var-subjectToken";
+    private static final String CHAIN_VAR_SUBJECT_TOKEN_TYPE_HEADER_KEY = "x-chain-var-subjectTokenType";
+    private static final String CHAIN_VAR_SAML_PROVIDER_DESTINATION_NAME_HEADER_KEY =
+        "x-chain-var-samlProviderDestinationName";
```
MatKuhr, line 53: Instead of duplicating, let's rather make the existing values package-private

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+                .codeVerifier("codeVerifier")
+                .chainName("chainName")
+                .chainVarSubjectToken("chainVarSubjectToken")
+                .chainVarSubjectTokenType("chainVarSubjectTokenType")
+                .chainVarSamlProviderDestinationName("chainVarSamlProviderDestName")
+                .build();
+
+        assertThat(destination.getHeaders(URI.create(VALID_URI.toString())))
+            .extracting(Header::getName)
+            .contains(
```
MatKuhr, line 122: This and the next test don't have strong assertions. Maybe split into "headers depending on static vs. dynamic" and "headers set by methods independent of static/dynamic".  The tenant / auth header values are also tested separately below, so maybe not necessary to cover them here?

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+    @Test
+    void testBuildAddsTenantIdHeaderIfTenantPresent()
+    {
+        Tenant tenant = new DefaultTenant(TEST_TENANT_ID, TEST_TENANT_SUBDOMAIN);
+        TransparentProxyDestination destination =
+            TransparentProxyDestination.dynamicDestination(TEST_DEST_NAME, TRANSPARENT_PROXY_GATEWAY).build();
+        assertNotNull(destination);
+        TenantAccessor.executeWithTenant(tenant, () -> {
+            assertThat(destination.getHeaders(URI.create(TRANSPARENT_PROXY_GATEWAY))).anyMatch(header -> {
```
MatKuhr, line 197: why not just `.contains(new Header(..))`?

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+                CHAIN_VAR_SAML_PROVIDER_DESTINATION_NAME_HEADER_KEY);
+    }
+
+    @Test
+    void testBuildAddsTenantIdHeaderIfTenantPresent()
+    {
+        Tenant tenant = new DefaultTenant(TEST_TENANT_ID, TEST_TENANT_SUBDOMAIN);
+        TransparentProxyDestination destination =
+            TransparentProxyDestination.dynamicDestination(TEST_DEST_NAME, TRANSPARENT_PROXY_GATEWAY).build();
+        assertNotNull(destination);
```
MatKuhr, line 195:
```suggestion
```

Since it's annotated nonnull anyway 😉

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+        TenantAccessor.executeWithTenant(tenant, () -> {
+            assertThat(destination.getHeaders(URI.create(TRANSPARENT_PROXY_GATEWAY))).anyMatch(header -> {
+                if( !header.getName().equalsIgnoreCase(TENANT_ID_HEADER_KEY) )
+                    return false;
+                assertNotNull(header.getValue());
+                return header.getValue().equals(TEST_TENANT_ID);
+            });
+        });
```
MatKuhr, line 203: It would be good to also test the case, where no tenant is present in the context and no tenant has been set manually

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
isevdalinosap: Done

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -0,0 +1,671 @@
...
+    public Collection<Header> getHeaders( @Nonnull final URI requestUri )
+    {
+        final Collection<Header> allHeaders = new ArrayList<>();
+
+        allHeaders.addAll(customHeaders);
+        allHeaders.addAll(getHeadersFromHeaderProviders(requestUri));
+
+        // Automatically add tenant headers if not already present
+        TenantAccessor.tryGetCurrentTenant().onSuccess(tenant -> {
+            if( !containsHeader(allHeaders, TENANT_ID_HEADER_KEY) && !tenant.getTenantId().isEmpty() ) {
```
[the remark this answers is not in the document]
isevdalinosap: Done

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+                REDIRECT_URI_HEADER_KEY,
+                CODE_VERIFIER_HEADER_KEY,
+                CHAIN_NAME_HEADER_KEY,
+                CHAIN_VAR_SUBJECT_TOKEN_HEADER_KEY,
+                CHAIN_VAR_SUBJECT_TOKEN_TYPE_HEADER_KEY,
+                CHAIN_VAR_SAML_PROVIDER_DESTINATION_NAME_HEADER_KEY);
+    }
+
+    @Test
+    void testBuildAddsTenantIdHeaderIfTenantPresent()
```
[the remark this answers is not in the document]
isevdalinosap: Done

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
isevdalinosap: Done

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
```diff
@@ -0,0 +1,253 @@
...
+    @Test
+    void testBuildThrowsExceptionWhenDestinationNameMissing()
+    {
+        Assertions
+            .assertThatThrownBy(
+                () -> TransparentProxyDestination.dynamicDestination("", TRANSPARENT_PROXY_GATEWAY).build())
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessageContaining(
+                "The 'destinationName' property is required for dynamic destinations but was not set.");
+    }
```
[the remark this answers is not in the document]
isevdalinosap: Done

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
isevdalinosap: The transparent proxy accepts x-token-service-tenant header and then sends it to the destination service as x-tenant when `tokenServiceURLType = common`. The links added are just examples where it is used. I can't find super specific documentation about this in the connectivity area but i will add some more explanations in the java doc.

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
isevdalinosap: Done

cloudplatform/connectivity-destination-service/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestinationTest.java
isevdalinosap: I don't quite understand what you mean here. These tests are just to validate that the dynamic and static destination can be build with the respective header shortcuts.

PR #860, opened 2025-07-17
feat: [Connectivity] Introduce OAuth2Option for token cache parameters 

rpanackal:
## Context
 

SAP/cloud-sdk-java-backlog#[#296](https://github.com/SAP/ai-sdk-java-backlog/issues/296).

Please provide a short description of what your change does and why it is needed.

### Feature scope:

 
- [ ] Add a new OAuth2Option called TokenCacheParameters that allows to configure token cache duration, expiration delta and cache su

## Definition of Done

- [x] Functionality scope stated & covered
- [x] Tests cover the scope above
- [ ] ~Error handling created / updated & covered by the tests above~
- [ ] ~Documentation updated~
- [x] Release notes updated

PR #861, opened 2025-07-17
fix: [Connectivity] Change default token cache duration to 1hr

rpanackal:
## Context
 

SAP/cloud-sdk-java-backlog#[296](https://github.com/SAP/ai-sdk-java-backlog/issues/296).

Relax the OAuth2 token cache duration from 10mins to 1hr

Related PR adding convenience 
- https://github.com/SAP/cloud-sdk-java/pull/860

## Definition of Done

- [ ] ~Tests cover the scope above~
- [ ] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
+++ b/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
@@ -220,7 +220,7 @@ class OAuth2Service
             // the IAS property supplier will have set this to the provider ID by default
             // we have to override it here to match the current tenant, if the current tenant is defined
             additionalParameters.put("app_tid", tenantId);
-            if (onBehalfOf == OnBehalfOf.NAMED_USER_CURRENT_TENANT) {
+            if( onBehalfOf == OnBehalfOf.NAMED_USER_CURRENT_TENANT ) {
                 // workaround until a fix is provided by IAS
                 additionalParameters.put("refresh_token", "0");
             }


release_notes.md
```diff
@@ -22,3 +22,4 @@
 
 - OData v2 and OData v4: Fixes eager HTTP response evaluation for _Create_, _Update_, and _Delete_ request builders in convenience APIs.
   Previous change of `5.20.0` may have resulted in the HTTP connection being left open after the request was executed.
+- Relax OAuth2 token cache duration to 1hr to avoid unnecessary token refreshes.
```
newtork, line 25: Should we indicate that the cache is for the HTTP layer only?


```suggestion
- Relax OAuth2 token query-response cache duration to 1hr to avoid unnecessary token refreshes.
```

newtork (commented):
Looks good! Only question regarding release note.
Is this PR in conflict with https://github.com/SAP/cloud-sdk-java/pull/860?

release_notes.md
MatKuhr: I feel like this might be more confusing for users, in the end what matters is that our token caching is more efficient now

newtork: It was confusing to us when we looked into it; realizing the parameter was affecting the token request cache.

It was not obvious that the cache entry is invalidated automatically when the token is expired (including the `tokenExpirationDelta` buffer). That means even if the token was valid for another 30s, a new one would be queried.

It may be obvious for the three of us today, but this is hardly documented anywhere - including javadoc or inline-comments.

cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java
```diff
@@ -220,6 +220,10 @@ private void setAppTidInCaseOfIAS( @Nullable final String tenantId )
             // the IAS property supplier will have set this to the provider ID by default
             // we have to override it here to match the current tenant, if the current tenant is defined
             additionalParameters.put("app_tid", tenantId);
+            if (onBehalfOf == OnBehalfOf.NAMED_USER_CURRENT_TENANT) {
+                // workaround until a fix is provided by IAS
+                additionalParameters.put("refresh_token", "0");
```
newtork, line 225: (Comment)

This looks like a drive-by change. I would be okay with leaving it in, but please in the future consider a dedicated PR.

MatKuhr: It's not drive-by, because increasing the cache duration would increase an existing risk without this additional setting. See https://jira.tools.sap/browse/SECREQ-5220 for more context

release_notes.md
MatKuhr: What else would the parameter be affecting? From a user point of view, there is only one cache, namely the token cache, and for users it doesn't matter what exactly is cached there (if it's only the token or some additional metadata).

> It was not obvious that the cache entry is invalidated automatically when the token is expired

It must be that way, otherwise the cache may yield expired tokens which can't be the expected behavior.

Also, the token expiry is typically included in the token itself. That's why to me, `query-response` is meaningless in this context.

rpanackal: I am for now keeping the release notes entry as is for the sake of simplicity.

But, I have updated all the related javadoc to include the information that the cache entries are token response for future reference.

rpanackal: To be clear, the javadoc lies in the other PR where `TokenCacheParameters` is introduced.

PR #874, opened 2025-08-01
chore: [BlackDuck] Fix CVE-2024-7254 again

newtork:
Blackduck is failing today for dependency `protobuf-java`.
https://github.com/SAP/cloud-sdk-java/actions/runs/16661849685/job/47180048733

* We are using fixed version in `connectivity-ztis` with `3.25.8:runtime`.
* Maven detects flawed version in `connectivity-oauth` with `3.25.5:runtime (optional)`

See more available versions and their vulnerabilities here:
https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java?p=3

The problem was identified today.
The symptom can be found in older versions.

Options:
1️⃣ We ignore the BlackDuck finding, mark it as false-positive (is it?)
2️⃣ **We change from implicit `<dependencyManagement>` to explicit `<dependency>` (+ exclusion)**
3️⃣ ~Move dependency management to parent pom~ **Does not work for users of the SDK**

### DIFF
--- a/cloudplatform/connectivity-ztis/pom.xml
+++ b/cloudplatform/connectivity-ztis/pom.xml
@@ -32,6 +32,11 @@
 		<dependencies>
 			<!-- resolve inconsistent versions coming from spiffe -> io.grpc -->
 			<dependency>
+				<groupId>com.google.protobuf</groupId>
+				<artifactId>protobuf-java</artifactId>
+				<version>3.25.8</version>
+			</dependency>
+			<dependency>
 				<groupId>io.grpc</groupId>
 				<artifactId>grpc-bom</artifactId>
 				<version>1.73.0</version>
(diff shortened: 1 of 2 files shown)


cloudplatform/connectivity-ztis/pom.xml
```diff
@@ -88,6 +88,19 @@
 			<groupId>io.grpc</groupId>
 			<artifactId>grpc-protobuf</artifactId>
 			<scope>runtime</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>com.google.protobuf</groupId>
+					<artifactId>protobuf-java</artifactId>
+				</exclusion>
+			</exclusions>
```
MatKuhr, line 96: this exclusion effectively doesn't do anything

Issue #885, opened 2025-08-10
EntityMetadata generates code that does not compile.


kevinm416:
### Describe the Bug

I'm generating odata-v4 for Microsoft Dynamics 365 using `com.sap.cloud.sdk.datamodel:odata-v4-generator:5.21.0`. The EntityMetadata.java file that it generates does not compile because it redefines a variable:
```
            if (cloudSdkValues.containsKey("properties")) {
                final Object value = cloudSdkValues.remove("properties");
                if (value instanceof Iterable) {
                    final LinkedList<entityMetadataField> properties = new LinkedList<entityMetadataField>();
                    for (Object properties: ((Iterable<?> ) value)) { // <----- Redefines final properties variable
                        if (properties instanceof Map) {
                            final entityMetadataField item = new entityMetadataField();
                            @SuppressWarnings("unchecked")
                            final Map<java.lang.String, Object> inputMap = ((Map<java.lang.String, Object> ) value);
                            item.fromMap(inputMap);
                            properties.add(item);
                        }
                    }
                    setProperties(properties);
                }
                if ((value == null)&&(getProperties()!= null)) {
                    setProperties(null);
                }
            }
```

### Steps to Reproduce

1. Use edmx from Microsoft Dynamics 365:
  ```
<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
    <edmx:DataServices>
        <Schema Namespace="Microsoft.NAV" xmlns="http://docs.oasis-open.org/odata/ns/edm">
            <EntityType Name="entityMetadata">
                <Key>
                    <PropertyRef Name="entityName" />
                </Key>
                <Property Name="entityName" Type="Edm.String" Nullable="false" />
                <Property Name="entitySetName" Type="Edm.String" />
                <Property Name="entityCaptions" Type="Collection(Microsoft.NAV.entityMetadataLabel)" />
                <Property Name="entitySetCaptions" Type="Collection(Microsoft.NAV.entityMetadataLabel)" />
                <Property Name="properties" Type="Collection(Microsoft.NAV.entityMetadataField)" />
                <Property Name="actions" Type="Collection(Microsoft.NAV.entityMetadataAction)" />
                <Property Name="enumMembers" Type="Collection(Microsoft.NAV.entityMetadataEnumMember)" />
            </EntityType>
  ```
2. Generate using SAP's odata-v4-generator
```
[...]

PR #886, opened 2025-08-11
fix: [ODatav4] Support for property name `properties`

CharlesDuboisSAP:
## Context
 

- https://github.com/SAP/cloud-sdk-java/issues/885

Generating an OData v4 client with a property name `properties` would create a collision with an internal variable.

### Feature scope:

 
- [x] Renamed internal `properties´ variable to cloudSdkProperties

## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -21,4 +21,5 @@
 ### 🐛 Fixed Issues
 
 - [ODatav4] Fixed an issue when generating clients.
-  - Property names: `value`, `ìtem` and `properties` are now allowed.
+  - Property names: `value`, `item` and `properties` are now allowed.
+  


datamodel/odata-v4/odata-v4-generator/src/main/java/com/sap/cloud/sdk/datamodel/odatav4/generator/NamespaceClassGenerator.java
```diff
@@ -721,7 +721,7 @@ else if( isPrimitive ) {
                 final JBlock ifValueIsList = ifFoundBody._if(value._instanceof(typeIterable))._then();
                 final JVar listInst = ifValueIsList.decl(JMod.FINAL, listType, javaFieldName, JExpr._new(listType));
                 final JExpression iter = JExpr.cast(typeIterable.narrow(codeModel.wildcard()), value);
-                final JForEach forEach = ifValueIsList.forEach(typeObject, "properties", iter);
+                final JForEach forEach = ifValueIsList.forEach(typeObject, "cloudSdkProperties", iter);
                 final JBlock isMap = forEach.body()._if(forEach.var()._instanceof(codeModel.ref(Map.class)))._then();
                 final JVar item = isMap.decl(JMod.FINAL, javaType, "item", JExpr._new(javaType));
```
newtork, line 726: While I approve the change, could we also fix the other occurrences as well? E.g. define a static common variable name prefix, and prepend it to the commands similar to `.forEach`, `.decl`.

We already did something similar for `cloudSdkValue` it seems.

PR #903, opened 2025-08-20
Seamless handling of Cross-level consumption of destination-fragment pairs

panayotmarinov:
## Context
 

SAP/cloud-sdk-java-backlog#ISSUENUMBER.

### Feature scope:

 
- [ ] Task1 
- [ ] Task2 
  - [ ] SubTask1 

## Definition of Done

- [ ] Functionality scope stated & covered
- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -673,6 +675,34 @@ public DynamicBuilder fragmentName( @Nonnull final String fragmentName )
             return header(new Header(FRAGMENT_NAME_HEADER_KEY, fragmentName));
         }
 
+        /**
+         * Sets the destination level for the dynamic destination. See
+         * https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/dynamic-lookup-of-destinations
```
MatKuhr, line 680: I don't see any docs here about the level?

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
```diff
@@ -673,6 +675,34 @@ public DynamicBuilder fragmentName( @Nonnull final String fragmentName )
...
+         *
+         * @param destinationLevel
+         *            The level of the destination to use.
+         * @return This builder instance for method chaining.
+         */
+        @Nonnull
+        public DynamicBuilder destinationLevel( @Nonnull final String destinationLevel )
```
MatKuhr, line 687: I assume this is either "subaccount" or "instance". For such cases, we usually use enums and, in fact, just introduced one: https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationServiceOptionsAugmenter.java#L139

~But I also assume the two `provider` options don't apply to transparent Proxy? If it's just two options and not 4, then you could also consider to default to `subaccount` and just add one method without args to explicitly change it to `.instanceLevel()`~ From the test code below it looks like provider is also supported, then let's reuse the enum.

What is the behavior of TP if nothing is given? I assume subaccount?

MatKuhr (commented):
please also update the PR description and sign the CLA 😉

MatKuhr:
>  applied there with the next transparent proxy release.

Got it, I now marked this PR as draft and we can merge it once the release is available. @panayotmarinov kindly ping us when the release is ready and the PR has been finalized (merge conflicts, PR description, etc.). Thanks!

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/TransparentProxyDestination.java
MatKuhr: > the "default level" is "instnace ".

for the /v2 consume API yes, for the /v1 find destination it will fallback to subaccount. So I assume TP uses /v2 under the hood?

Because the Cloud SDK behaves according to /v1 by default, and only if cross-level is enabled, it uses /v2...

MatKuhr (approved):
@panayotmarinov if not provided, what is the default for the two? Maybe you can clarify by adding that info to the Javadoc. Otherwise LGTM!

PR #942, opened 2025-09-24
fix: CircuitBreaker stack-trace

CharlesDuboisSAP:
## Context
 

SAP/cloud-sdk-java-backlog#476.

- previously: `CircuitBreaker is OPEN and does not permit further calls`
- now: `Failed to get destination ... Caused by: Destination could not be found for path destination-configuration/v1/destinations/xxxx.`

### Feature scope:

 
- [x] Restored stack-trace for circuit breaker exception

⚠️ Note that there is a bug in the circuit breaker. **All destination calls use the same circuit breaker.**
## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated


Jonas-Isr (approved):
LGTM

Issue #970, opened 2025-10-17
Repeated "Connection pool shut down" errors after updating the Cloud SDK version


melwinlobo18:
### Describe the Bug

We upgraded the Cloud SDK from version `5.18` to `5.22`.

Along with this:
* Java Development Kit (JDK) upgrade from JDK 17 to JDK 21.
* The CAP Java library was updated from version 3 to version 4.

Post this upgrade, we see a lot of **5XX** issues with the requests in our production landscape, with the following stack trace:
```
Exception:
java.lang.IllegalStateException

Message:
Connection pool shut down

Stacktrace:
org.apache.http.util.Asserts.check(Asserts.java:34)
org.apache.http.impl.conn.PoolingHttpClientConnectionManager.requestConnection(PoolingHttpClientConnectionManager.java:269)
org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:176)
org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:186)
org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:89)
org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185)
org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:83)
org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:108)
com.sap.cloud.sdk.cloudplatform.connectivity.HttpClientWrapper.execute(HttpClientWrapper.java:145)
com.sap.cloud.sdk.cloudplatform.connectivity.HttpClientWrapper.execute(HttpClientWrapper.java:32)
com.sap.cds.services.utils.lib.tools.impl.ServiceCallImpl.execute(ServiceCallImpl.java:217)
...
```

* The CPU and memory usage are completely normal. 
* The issues are intermittent. A few network calls fail intermittently, occurring at random intervals throughout the day.

### Steps to Reproduce

NA

### Expected Behavior

Requests are processed normally without the `connection pool shut down`

### Used Versions

- Java and Maven version via `mvn --version`: Java version: 21.0.8, Apache Maven 3.9.4
- SAP Cloud SDK version: 5.22
- Spring Boot or CAP version: CAP Java v4.3.0

<details><summary>Dependency tree via <code>mvn dependency:tree</code></summary>
    
```
[...]

PR #976, opened 2025-10-21
feature: HttpClient health-check in default cache implementation

newtork:
## Context
 

Related issue
https://github.com/SAP/cloud-sdk-java/issues/970

Adding a health check that uses reflection to check boolean flag on connection pool whether it's shutdown or not.

### Feature scope:

 
- [ ] Confirmed to be working by user
- [ ] No unexpected behavior change from E2E tests, etc.

## Definition of Done

- [x] ~Documentation updated~
- [ ] Release notes updated


newtork:
We will likely not merge the check testing for symptoms. Instead we will try and fix the cause.

PR #1004, opened 2025-11-17
Add test covering the default retry behavior of HTTP client 5

MatKuhr:
## Context

See https://hc.apache.org/httpcomponents-client-5.5.x/current/httpclient5/apidocs/org/apache/hc/client5/http/impl/DefaultHttpRequestRetryStrategy.html

Adding a test just to verify & document the current state


newtork (approved):
Thanks for providing a test for current behavior

PR #1010, opened 2025-11-19
feat: [OpenAPI] Type and Import mapping support

rpanackal:
## Context
 
- https://github.com/SAP/ai-sdk-java-backlog/issues/239

When OpenAPI generator find the following schema in input spec file, particularly the mention `string+binary`, it maps the response to type `File`. 

```yaml
responses:
  '200':
    description: Successful response
    content:
      application/octet-stream:
        schema:
          type: string
          format: binary
```

Unfortunately, while invoking any API method that returns `File` type, we will run into the following error. (Find the bug report in linked BLI)
 
```
UnknownContentType Could not extract response: no suitable HttpMessageConverter found for response type [class java.io.File] and content type [application/octet-stream]
```

The issue above suggests that the mapped type `File` is simply unsupported and should be changed. We have had similar issues in the past which could have simply been resolved with a different type mapping. So instead of a per-type fix, we can resort to a more generic solution. To do so, we can take advantage of the `<typeMappings>` and `<importMappings>` features of OpenAPI generator. They are can be set via the OpenAPI generator CLI or through the `pom.xml `.

- `<typeMappings>` : allow you to override OpenAPI type with some custom type
- `<importMappings>` : allows you to replace the file imports for any type with another qualified path. Often used in tandem with defined `<typeMappings>` to correct the import statements.

**Problem**:

Our OpenAPI generator wrapper and corresponding maven plugin, does not accept `<typeMappings>` and `<importMappings>`. The goal of this PR is to propagate their configuration down to the OpenAPI generator.

### Feature scope:

 
- [x] Accept `<typeMappings>`
- [x] Accept `<importMappings>`
- [x] Test the mojo and propagation of the mappings. 
- [x] Add a test in `openapi-api-sample` to demo fix for related bug report 

## Definition of Done

- [ ] ~Documentation updated~
- [ ] Release notes updated

### DIFF
--- a/datamodel/openapi/openapi-api-sample/src/test/java/com/sap/cloud/sdk/datamodel/openapi/sample/api/DeserializationTest.java
+++ b/datamodel/openapi/openapi-api-sample/src/test/java/com/sap/cloud/sdk/datamodel/openapi/sample/api/DeserializationTest.java
@@ -130,7 +130,7 @@ class DeserializationTest
                             .withHeader("Content-Type", "application/octet-stream")
                             .withBody(binaryData)));
 
-        byte[] result = sut.sodasDownloadIdGet(1L);
+        final byte[] result = sut.sodasDownloadIdGet(1L);
         assertThat(result).isNotNull();
         assertThat(result).isEqualTo(binaryData);
     }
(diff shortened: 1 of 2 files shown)


newtork:
competitive PR
https://github.com/SAP/ai-sdk-java/pull/660

PR #1017, opened 2025-11-27
fix: Improve Circuit breaker behaviour

Jonas-Isr:
## Context

SAP/cloud-sdk-java-backlog#495

This PR improves the circuit breaker logic by prepending an additional check to `tryGetDestination`. When a single destination is called, it is first checked whether this destination appears in the result of a get all destination call. If the destination is not present, no call to the destination is made. Thus the circuit breaker does not open when non-existing destination(s) are called multiple times.

This new behaviour is implemented as opt-in to not introduce breaking changes.

### Feature scope:

- [x] Add additional check
- [x] add feature switch to turn on new additional check

## Definition of Done

- [x] Documentation updated
    - https://github.com/SAP/cloud-sdk/pull/2327 
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
+++ b/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
@@ -457,8 +457,8 @@ public class DestinationService implements DestinationLoader
 
         /**
          * Disables checking if a destination exists before trying to call it directly when invoking
-         * {@link #tryGetDestination}. All available destinations can be found with {@link
-         * #getAllDestinationProperties}.
+         * {@link #tryGetDestination}. All available destinations can be found with
+         * {@link #getAllDestinationProperties}.
          *
          * @since 5.25.0
          */


cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
```diff
@@ -443,6 +446,11 @@ static boolean isChangeDetectionEnabled()
             return changeDetectionEnabled;
         }
 
+        static boolean isGetAllDocumentsPrepended()
```
newtork: (Minor)

You already have lombok setter.
You could just use a lombok getter similarly.

rpanackal (commented):
(Question)

So the idea was instead of directly augmenting resilience config cache key, we introduce a (opt-in) check for the existence of a destination, essentially avoid tripping the circuit breaker in the first place.

But, I wonder what if a destination call fails for a different reason than `DestinationAccessException`? Is that possible? If so, I don't see how the current approach would resolve the issue?

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
```diff
@@ -443,6 +455,28 @@ static boolean isChangeDetectionEnabled()
...
+        /**
+         * Enables checking if a destination exists before trying to call it directly when invoking
+         * {@link #tryGetDestination}.
+         *
+         * @since 5.25.0
+         */
+        public static void enablePreLookupCheck()
+        {
+            preLookupCheckEnabled = true;
+        }
```
MatKuhr: Why have this if enabled by default? Users should never switch this on/off back-and-forth, so one method should be enough, no?

```suggestion
```

PR #1048, opened 2025-12-30
refactor: [OpenAPI] Part 1: Refactor Apache related files and templates

rpanackal:
## Context
 

SAP/cloud-sdk-java-backlog#464.

The parent PR contains all files as they would be if we purely relied on generation via OpenAPI code generator with commonly generated files packages with `openapi-core` module.

The scope of this PR is to reduce public API, achieve feature parity with resttemplate based generation and refactoring code for readability.

### Feature scope:

 
- [x] Carefully removed 11 unnecessary class files
  - All `auth/*.java` files - We support authentication via destination
  - `ApiException` removed in favour existing `OpenApiRequestException` (modified)
  - `ServerConfiguration`, `ServerVariable`, `StringUtil` removed as we do not support single client instance pointing to multi urls
  - apache/RFC3339*.java classes removed in favor of reusing `apiClient/RFC3339DateFormat`
- [x] Reduce public api in `BaseApi`

#### Improvements in `ApiClient`

- [x] Public api removed
    - `defaultHeaderMaps`: headers expected to be set request level, destination header provider or custom `ClosableHttpClient`
    - `defaultCookieMap`: no cookie support
    - `servers`, `serverIndex`, `serverVariables`: only single `baseUrl` support
    -  `lastStatusCode`, `lastResponseHeaders`: Removed due to feature engineering concerns. To compensate, new api introduced  in `OpenApiRequestException` and new `OpenApiResponse`.
    - `debugging` flag, `connectionTimeout`: useless and to be configured elsewhere.
    - `cookieParams` and `authNames` that sets request level configuration for cookies and authorization (but unused) are removed from `invokeAPI` signature
- [x] All api not used by generated classes are now set to the strictest access possible.
- [x] `@Beta` on jackson exposing api 
- [x] All public api that doesn't rely/modify on internal state is now `static`
- [x] Move out all response handling logic into new package private class `DefaultApiResponseHandler`
- [x] ctors replaced/hidden with static factory methods
- [x] No-args ctor introduced (deviation from resttemplate)
- [x] Make ApiClient a **almost** `@Value `class
- [x] Fix url building handle egde case "basePath/" + "/v2" = "basePath//v2"
- [x] Vulnerability fix to remove serializing payload onto `Exception` messages

### Feature Parity with resttemplate (template modifications)
- [x] Object mapper configuration
- [x] Nullability annotation and `final` on local vars, parameters, returns as appropriate.
- [x] Javadoc improvements
- [x] Destination based ctor and ai sdk ctor support 
- [x] Method overloading per operation (get, post, etc) for optional and required params
- [x] Vendor extension support `x-sap-cloud-sdk-operation-name`, `x-return-nullable`
- [x] Support `apache/OpenApiResponse` instead of `void` as return type for remote calls. Existing `core/OpenApiResponse` has Spring components in public contract. 😿 
 
## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -49,6 +49,9 @@ updates:
         versions: ['3.6.1', '3.6.2']
       - dependency-name: 'com.sap.cloud.security:env'
         versions: ['3.6.1', '3.6.2']
+      # needs Java 21
+      - dependency-name: 'com.puppycrawl.tools:checkstyle'
+        versions: [ ">=13.0.0" ]
 
   # archetype updates
   # Dependabot seems to be unable to handle those, so this is disabled for now
--- a/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommand.java
+++ b/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/GetOrComputeSingleDestinationCommand.java
@@ -8,13 +8,13 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import com.github.benmanes.caffeine.cache.Cache;
-import com.google.common.base.Functions;
 import com.sap.cloud.sdk.cloudplatform.cache.CacheKey;
 import com.sap.cloud.sdk.cloudplatform.connectivity.exception.DestinationAccessException;
 import com.sap.cloud.sdk.cloudplatform.security.principal.exception.PrincipalAccessException;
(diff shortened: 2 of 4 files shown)


datamodel/openapi/openapi-core/src/main/java/com/sap/cloud/sdk/services/openapi/apache/ApiClient.java
```diff
@@ -12,444 +12,120 @@
...
-     * @return API client
-     */
-    public ApiClient setConnectTimeout( int connectionTimeout )
+    public static ApiClient create( @Nonnull final Destination destination )
     {
-        this.connectionTimeout = connectionTimeout;
-        return this;
-    }
+        return fromHttpClient((CloseableHttpClient) ApacheHttpClient5Accessor.getHttpClient(destination))
+            .withBasePath(destination.asHttp().getUri().toString());
```
newtork: (Minor)

I'm not sure whether the following is unnecessary: `withBasePath(destination.asHttp().getUri().toString())` 

unit tests will show.

datamodel/openapi/openapi-generator/src/main/resources/openapi-generator/mustache-templates/libraries/apache-httpclient/api.mustache
```diff
@@ -24,222 +28,155 @@ import java.util.Locale;
...
     }
 
-    public {{classname}}(ApiClient apiClient) {
+    /**
+    * Instantiates this API class to invoke operations on the {{appName}} based on a given {@link ApiClient}.
+    *
+    * @param apiClient
+    *            ApiClient to invoke the API on
+    */
+    @Beta
```
rpanackal: Remove `@Beta` annotation

PR #1056, opened 2026-01-05
feat: Add pre lookup check to `DestinationService.tryGetDestination`

Jonas-Isr:
BLI:
- https://github.com/orgs/SAP/projects/62/views/7?pane=issue&itemId=140717964&issue=SAP%7Ccloud-sdk-java-backlog%7C495

Prerequisite:
- https://github.com/SAP/cloud-sdk-java/pull/1055

## Context

This is a feature that we already [merged at one point](https://github.com/SAP/cloud-sdk-java/pull/1017) and then later removed again because we found problems that we were not able to fix and test before release. (The underlying problem we found and wanted to fix first is in the PR linked above.)

This PR prepends an additional check to `tryGetDestination`: when a single destination is called, it is first checked whether this destination appears in the result of a get all destination call. If the destination is not present, no call to the destination is made. The goal of this feature is to improve the circuit breaker, as the circuit breaker now does not open unnecessarily when non-existing destination(s) are called multiple times.

### Feature scope:
 
- [x] add new lookup to `DestinationService.tryGetDestination`
- [x] add tests

## Definition of Done

- [x] Documentation updated
    - https://github.com/SAP/cloud-sdk/pull/2327 
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
+++ b/cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
@@ -135,25 +135,25 @@ public class DestinationService implements DestinationLoader
 
     Option<Exception> validateDestinationLookup( final String destinationName, final DestinationOptions options )
     {
-        final Option<Exception> VALID_LOOKUP = Option.none();
+        final Option<Exception> validLookup = Option.none();
         if( !Cache.isEnabled() ) {
-            return VALID_LOOKUP;
+            return validLookup;
         }
         if( !Cache.preLookupCheckEnabled ) {
-            return VALID_LOOKUP;
+            return validLookup;
         }
         if( Cache.isUsingExperimentalFeatures(options) ) {
             final String msg =
                 "Using pre-lookup check together with either fragments, cross-level options, or custom headers might lead to unexpected behaviour. Pre-lookup check is skipped.";
             log.warn(msg);
-            return VALID_LOOKUP;
+            return validLookup;
         }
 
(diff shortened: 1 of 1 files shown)


Jonas-Isr:
Regarding a test showing the failure we saw in the e2e tests in December: 

I added an additional test that is green in the current implementation but will fail if the retrieval strategy is not correctly forwarded in the pre-lookup check of the `DestinationService`. In that case the destination is not found even though it exists, because the a different retrieval strategy is used for the pre-lookup check than was specified in the `DestinationOptions`. 

To see this, simply delete `retrievalStrategy` in line 397 of `DestinationService` and the test will fail. This is what happened in the previous implementation we merged to main in December.

cloudplatform/connectivity-destination-service/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DestinationService.java
```diff
@@ -122,17 +126,41 @@ static ResilienceConfiguration createResilienceConfiguration(
         Try<Destination>
         tryGetDestination( @Nonnull final String destinationName, @Nonnull final DestinationOptions options )
     {
-        if (Cache.preLookupCheckEnabled) {
-            if (Cache.isUsingExperimentalFeatures(options)) {
-                log.warn("Using pre-lookup check together with either fragments, cross-level options, or custom headers might lead to unexpected behaviour. Pre-lookup check is skipped.");
-            } else if (!preLookupCheckSuccessful(destinationName, options)) {
-                final String msg = "Destination %s was not found among the destinations of the current tenant.";
-                return Try.failure(new DestinationNotFoundException(destinationName, String.format(msg, destinationName)));
```
newtork: (Minor)
I found the hard-coded "destinations of current tenant" misleading for cases like "always-provider". Therefore I moved the code around a little and introduced the notion of "validating destination lookup". Maybe "intercept" would've been a better word(?)

PR #1070, opened 2026-01-16
feat: [OpenAPI] `ResponseMetadataListener` with case insensitive header map

rpanackal:
## Context
 

SAP/cloud-sdk-java-backlog[#464](https://github.com/SAP/cloud-sdk-java-backlog/issues/464).

### Feature scope:
 
- [x] Introduce a `ResponseMetadataListener` that accepts `OpenApiResponse` object which includes the metadata
- [x] Make case insensitive key match for headers using TreeMap

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/datamodel/openapi/openapi-core/src/main/java/com/sap/cloud/sdk/services/openapi/apache/ApiClient.java
+++ b/datamodel/openapi/openapi-core/src/main/java/com/sap/cloud/sdk/services/openapi/apache/ApiClient.java
@@ -93,6 +93,7 @@ public class ApiClient
     @Nullable
     private final String tempFolderPath;
 
+    @With( onMethod_ = @Beta )
     @Nullable
     private final OpenApiResponseListener openApiResponseListener;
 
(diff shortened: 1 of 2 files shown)


datamodel/openapi/openapi-core/src/main/java/com/sap/cloud/sdk/services/openapi/apache/ResponseMetadataListener.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Listener for receiving metadata about HTTP responses.
+ *
+ * @since 5.25.0
+ */
+@FunctionalInterface
+public interface ResponseMetadataListener
```
newtork: (Minor/Nitpick)

`ResponseMetadataListener` isn't very easy name.
What would you think of `OpenApiResponseListener`?

Jonas-Isr (dismissed):
LGTM

datamodel/openapi/openapi-core/src/main/java/com/sap/cloud/sdk/services/openapi/apache/ApiClient.java
```diff
@@ -94,7 +94,7 @@ public class ApiClient
     private final String tempFolderPath;
 
     @With
-    private final ResponseMetadataListener responseMetadataListener;
+    private final OpenApiResponseListener openApiResponseListener;
```
newtork: (Minor)

Could we add the `@Beta` annotation here?

While I'm okay with current state, I feel like having options in the future would be nice:
* rename it, e.g. to `withResponseListener`
* extend interface API, e.g. with HttpResponse reference, settings for error handling

PR #1110, opened 2026-03-04
feat: [OpenAPI] GZIP encoding

CharlesDuboisSAP:
## Context
 

SAP/cloud-sdk-java-backlog#363.

**Smaller payload.**
Corresponding AI SDK PR:
- https://github.com/SAP/ai-sdk-java/pull/779

### Feature scope:

 
- [x] If `Content-Encoding` is set to `gzip` -> compress the request payload

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/datamodel/openapi/openapi-core-apache/src/main/java/com/sap/cloud/sdk/services/openapi/apache/apiclient/ApiClient.java
+++ b/datamodel/openapi/openapi-core-apache/src/main/java/com/sap/cloud/sdk/services/openapi/apache/apiclient/ApiClient.java
@@ -14,6 +14,7 @@ package com.sap.cloud.sdk.services.openapi.apache.apiclient;
 
 import static com.sap.cloud.sdk.services.openapi.apache.apiclient.DefaultApiResponseHandler.isJsonMime;
 import static lombok.AccessLevel.PRIVATE;
+import static org.apache.hc.core5.http.HttpHeaders.CONTENT_ENCODING;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -492,7 +493,7 @@ public class ApiClient
         if( body != null || !formParams.isEmpty() ) {
             if( isBodyAllowed(Method.valueOf(method)) ) {
                 // Add entity if we have content and a valid method
-                builder.setEntity(serialize(body, formParams, contentTypeObj, headerParams.get("Content-Encoding")));
+                builder.setEntity(serialize(body, formParams, contentTypeObj, headerParams));
             } else {
                 throw new OpenApiRequestException("method " + method + " does not support a request body");
             }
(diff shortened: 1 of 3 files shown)


datamodel/openapi/openapi-core-apache/src/main/java/com/sap/cloud/sdk/services/openapi/apache/apiclient/ApiClient.java
```diff
@@ -578,4 +585,16 @@ public <T> T invokeAPI(
             throw new OpenApiRequestException(e);
         }
     }
+
+    private HttpEntity serializeGzip( final Object body, final ContentType contentType )
```
ZhongpinWang: [pp] Some personal preference (we use pp in js team) 🙂 It could be better if this method can be merged with the already existing `serialize()` method by passing an additional param `contentEncoding` since it is the responsibility of the `serialize()` method to consider if data needs to be encoded / compressed. 

Then maybe refactor the `serialize()` method as it now has 55 lines with a big if-else.

ZhongpinWang (commented):
Some clean code ideas. Bear my first Java PR review with some old JS habits. 😁

Jonas-Isr (commented):
I still feel like we should add some minimal (vibe-coded?) unit tests. I don't think the argument that this is tested through AI SDK is strong enough: If we ever discontinue the feature in AI SDK this will not be tested anymore and potentially break without us noticing.

datamodel/openapi/openapi-core-apache/src/main/java/com/sap/cloud/sdk/services/openapi/apache/apiclient/ApiClient.java
```diff
@@ -578,4 +510,112 @@ public <T> T invokeAPI(
...
+    private HttpEntity serializeJson(
+        @Nullable final Object body,
+        @Nonnull final ContentType contentType,
+        @Nullable final String contentEncoding )
+        throws OpenApiRequestException
+    {
+        if( "gzip".equals(contentEncoding) ) {
+            val outputStream = new ByteArrayOutputStream();
+            try( val gzip = new GZIPOutputStream(outputStream) ) {
+                gzip.write(objectMapper.writeValueAsBytes(body));
```
newtork: (Comment)

I'm not a huge fan of logic:
```
gzip header exists -> compress body
```

Normally you would go with:
```
compressed body exists -> add gzip header
```

But I understand this is like a workaround, piggy-pack feature-toggle.

rpanackal (approved):
The case insensitive "Content-Encoding" check may need to be more robust. We can improve that later as needed

LGTM.

Issue #1120, opened 2026-03-14
OData V4 CountRequestBuilder.toRequest() is stateful and mutates the resource path


Louis-Stegra:
### Describe the Bug

Calling OData V4 `CountRequestBuilder.toRequest()` multiple times on the same builder mutates the underlying `ODataResourcePath` and appends `/$count` repeatedly.

## Observed behavior

Given a `CountRequestBuilder` for a collection such as:

```text
/People
```

the first call to `toRequest()` produces:

```text
/People/$count
```

but subsequent calls on the same builder can produce:

```text
/People/$count/$count
/People/$count/$count/$count
```

This happens because `ODataRequestCount` mutates the `ODataResourcePath` instance it receives.

## Root cause

In `ODataRequestCount`:

```java
super(servicePath, resourcePath.addSegment("$count"), encodedQuery, protocol);
```

`ODataResourcePath.addSegment(...)` is mutating, so any shared path object is modified in place.

This makes `toRequest()` have side effects, which is surprising for SDK request builders and breaks our integration that inspect a request before executing it.

## Proposed fix

Treat the incoming `ODataResourcePath` as immutable from `ODataRequestCount`'s:
- copy it before mutating it when constructing a new ODataRequestCount instance 

There seems to be a similar setup in ODataRequestFunction.

I have drafted a PR with the suggested fix, feel free to disregard if you consider a different fix, or if you don't believe this is a bug. 

### Steps to Reproduce

```java
import com.sap.cloud.sdk.datamodel.odatav4.core.CountRequestBuilder;
import com.sap.cloud.sdk.datamodel.odatav4.referenceservice.namespaces.trippin.Person;
import com.sap.cloud.sdk.datamodel.odatav4.referenceservice.namespaces.trippin.Trip;
import com.sap.cloud.sdk.datamodel.odatav4.referenceservice.services.DefaultTrippinService;

public class Repro {
    public static void main(String[] args) {
        Person person = Person.builder().userName("russellwhyte").build();

        CountRequestBuilder<Trip> countBuilder =
            new DefaultTrippinService()
                .forEntity(person)
                .navigateTo(Person.TO_TRIPS)
                .count();

        System.out.println(countBuilder.toRequest().getRelativeUri());
        System.out.println(countBuilder.toRequest().getRelativeUri());
    }
}
```
Current behaviour, outputs:

/TripPinServiceRW/People('russellwhyte')/Trips/$count
[...]

PR #1121, opened 2026-03-14
fix: avoid mutating resource paths in ODataRequestCount

Louis-Stegra:
## Context
 

Bug report: #1120 

### Feature scope:

`CountRequestBuilder.toRequest()` for OData V4 had the side effect of mutating the shared `ODataResourcePath` when appending `$count` to it. Reusing the same builder could therefore produce invalid paths such as `$count/$count`. This change makes `ODataRequestCount` copy the incoming resource path before appending extra segments, so repeated `toRequest()` calls remain stable and do not corrupt later execution.
  
That fix was inspired by a similar implementation in `ODataRequestFunction`

  ### Feature scopes:

  - [x] Prevent ODataRequestCount constructors from mutating the provided ODataResourcePath
  - [x] Add a low-level regression test proving ODataRequestCount does not modify the original resource path
  - [x] Add a V4 builder-level regression test proving repeated CountRequestBuilder.toRequest() calls return the same URI
 
## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/expression/ODataResourcePath.java
+++ b/datamodel/odata-client/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/expression/ODataResourcePath.java
@@ -1,6 +1,7 @@
 package com.sap.cloud.sdk.datamodel.odata.client.expression;
 
 import java.util.ArrayList;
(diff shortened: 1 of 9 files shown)


newtork (approved):
looks good, thanks for the contribution!

PR #1129, opened 2026-03-27
feat: [OpenAPI] Customize apache request object before invocation

newtork:
Allows for runtime customization of request builder object before the `ApiClient` invocation.

Related:
* https://github.com/SAP/ai-sdk-java/pull/792
* https://github.com/SAP/ai-sdk-java-backlog/issues/379

Example:

```diff
  static RptClient forDestination(@Nonnull final Destination destination) {
    final var apiClient = ApiClient.create(destination)
        .withObjectMapper(getDefaultObjectMapper())
+       .withRequestCustomizer(request -> {
+         if(request.getPath().contains("/predict")) {
+           return request.addHeader("Accept-Encoding", "gzip");
+         }
+         return request;
+       });
    return new RptClient(new DefaultApi(apiClient));
  }
```

New methods:
```diff
+ ApiClient#withRequestCustomizer( UnaryOperator<ClassicRequestBuilder> )
```


Jonas-Isr (dismissed):
I like this! It looks pretty cool and should give us a clean solution to the GZIP problem.

Only one minor comment. Also would be interested in what the others think.

rpanackal (approved):
I tried different approaches that is more focused at modifying `BaseApi` class to achieve the same. But its either breaking, requires modifications to generated code or a lot of new public.

So, I think this is a great way to do this.

newtork:
Looks like we're preferring https://github.com/SAP/cloud-sdk-java/pull/1130 over this

PR #1135, opened 2026-04-01
fix ZTIS certificate caching

mpern:
## Context

Fixes the following issues:
- https://sap.stackenterprise.co/questions/90517
- https://sap.stackenterprise.co/questions/90606
- https://github.com/SAP/cloud-sdk-java/issues/1134

### Feature scope:

refresh the cached ZTIS keystore if certificate is close to expiration.
otherwise requests fail if the backend is up for more than 7 days.

## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated


MatKuhr:
@mpern we've now added a change that should allow the SDK to automatically detect rotations for the OAuth2 token flows: #1142 

While not 100% confirmed this will resolve all rotation issues, it should be an improvement. Thus, closing this PR for now.

Still, the current logic relies on comparing the `KeyStore` object references, which isn't ideal. If you like, once the fix is confirmed to work, please feel free to update this PR and re-open to change that aspect specifically.

PR #1153, opened 2026-04-20
feat: expose slowCallDurationThreshold in CircuitBreakerConfiguration

### LINKED ISSUE #1148: Expose slowCallDurationThreshold and slowCallRateThreshold in CircuitBreakerConfiguration
### Describe the Problem

`ResilienceConfiguration.CircuitBreakerConfiguration` currently exposes only a subset of the properties available in Resilience4j's `CircuitBreakerConfig`. Notably missing are:

- **`slowCallDurationThreshold`** (Resilience4j default: 60 seconds)
- **`slowCallRateThreshold`** (Resilience4j default: 100%)

This is problematic for applications that perform long-running operations (e.g., large SAP OData batch imports that routinely take 2–10 minutes per call). With the default 60-second slow-call threshold, these calls are classified as "slow", and once the slow-call rate exceeds the threshold, the circuit breaker transitions to OPEN — **even though all calls succeed**.

Since `CircuitBreakerConfig` is immutable after construction and `DefaultCircuitBreakerProvider` does not set these properties from the SDK configuration, there is **no clean way** to adjust the slow-call behavior without resorting to reflection hacks.

### Propose a Solution

Add two optional properties to `CircuitBreakerConfiguration`:

```java
@NoArgsConstructor(staticName = "of")
@Accessors(fluent = true)
@EqualsAndHashCode
@Getter
@Setter
public static final class CircuitBreakerConfiguration {
    // ...existing fields...

    @Nonnull
    private Duration slowCallDurationThreshold = Duration.ofSeconds(60); // Resilience4j default

    private float slowCallRateThreshold = 100f; // Resilience4j default
}
```

Then wire them through in `DefaultCircuitBreakerProvider.getCircuitBreaker()`:

```java
CircuitBreakerConfig.custom()
    .failureRateThreshold(config.failureRateThreshold())
    .waitDurationInOpenState(config.waitDuration())
    .slidingWindowType(CircuitBreakerConfig.SlidingWindowType.COUNT_BASED)
    .slidingWindowSize(config.closedBufferSize())
    .minimumNumberOfCalls(config.closedBufferSize())
    .permittedNumberOfCallsInHalfOpenState(config.halfOpenBufferSize())
    .slowCallDurationThreshold(config.slowCallDurationThreshold())     // NEW
    .slowCallRateThreshold(config.slowCallRateThreshold())             // NEW
    .build();
```

This is fully backwards-compatible since the proposed defaults match Resilience4j's existing defaults.

### Describe Alternatives

The only current workaround is using **reflection** to modify the `CircuitBreakerConfig` after construction:

```java
// Step 1: Access DefaultCircuitBreakerProvider via reflection on Resilience4jDecorationStrategy.decorators (protected field)
Field decoratorsField = Resilience4jDecorationStrategy.class.getDeclaredField("decorators");
decoratorsField.setAccessible(true);
List<GenericDecorator> decorators = (List<GenericDecorator>) decoratorsField.get(strategy);
DefaultCircuitBreakerProvider cbProvider = decorators.stream()
    .filter(DefaultCircuitBreakerProvider.class::isInstance)
    .map(DefaultCircuitBreakerProvider.class::cast)
    .findFirst().orElseThrow();

// Step 2: Obtain CircuitBreaker instance and mutate the immutable config via reflection
CircuitBreaker cb = cbProvider.getCircuitBreaker(myResilienceConfiguration);
Field field = cb.getCircuitBreakerConfig().getClass().getDeclaredField("slowCallDurationThreshold");
field.setAccessible(true);
field.set(cb.getCircuitBreakerConfig(), Duration.ofMinutes(5));
```

This approach is fragile because:
1. It accesses `protected` internals of `Resilience4jDecorationStrategy`
2. It mutates an immutable Resilience4j config object via reflection
3. It can break silently on any minor version update of either the SAP Cloud SDK or Resilience4j

### Affected Development Phase

Production

### Impact

Inconvenience

### Timeline

No immediate deadline — this is a long-standing workaround we'd like to eliminate. Happy to contribute a PR if the team agrees with the approach.

dira-AUR:
## Summary

Exposes the `slowCallDurationThreshold` and `slowCallRateThreshold` properties from resilience4j's `CircuitBreakerConfig` in the SAP Cloud SDK's `ResilienceConfiguration.CircuitBreakerConfiguration` wrapper.

## Motivation
Currently, users who need to configure slow-call detection for their circuit breakers must resort to reflection to access the underlying resilience4j configuration, which is fragile and not officially supported.
The resilience4j `CircuitBreakerConfig` supports:
- `slowCallDurationThreshold` — duration above which a call is considered slow
- `slowCallRateThreshold` — percentage of slow calls that triggers the circuit breaker to open
These are common production requirements, especially when integrating with external systems (e.g., SAP OData services) that may degrade gradually rather than fail outright.

## Changes
- Added `slowCallDurationThreshold(Duration)` and `slowCallRateThreshold(int)` to `CircuitBreakerConfiguration`
- These values are forwarded to the underlying resilience4j `CircuitBreakerConfig.Builder`
- Closes https://github.com/SAP/cloud-sdk-java/issues/1148

### DIFF
--- a/cloudplatform/resilience-api/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceConfiguration.java
+++ b/cloudplatform/resilience-api/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceConfiguration.java
@@ -429,6 +429,18 @@ public class ResilienceConfiguration
         public static final Duration DEFAULT_WAIT_DURATION = Duration.ofSeconds(10);
 
         /**
+         * The default duration threshold above which calls are considered as slow.
+         * Matches the Resilience4j default of 60 seconds.
+         */
+        public static final Duration DEFAULT_SLOW_CALL_DURATION_THRESHOLD = Duration.ofSeconds(60);
+
+        /**
+         * The default slow call rate threshold (as percentage within [0, 100]).
+         * Matches the Resilience4j default of 100%.
+         */
+        public static final float DEFAULT_SLOW_CALL_RATE_THRESHOLD = 100;
+
+        /**
          * Flag to indicate active CircuitBreakerConfiguration.
          */
         @Getter( AccessLevel.NONE )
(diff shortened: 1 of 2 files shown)


newtork (commented):
Looks good!

Please format javadoc and add `@since` tag
https://github.com/dira-AUR/cloud-sdk-java/compare/main...newtork:patch-1 (after pulling and merging latest changes from `main`)

rpanackal (approved):
LGTM and thanks for the contribution :)

PR #1155, opened 2026-04-21
fix: Upgrade minor netty version (transitive dependency patch)

newtork:
Currently blackduck scan is failing for 
* CVE-2026-33870 (7.5)
* CVE-2026-33871 (7.5)

Let's fix that by forcing a transitive dependency patch.


Jonas-Isr (approved):
Looks good

PR #1156, opened 2026-04-22
fix: Declare dependency updates explicitly while leveraging Netty BOM

newtork:
Follow up #1155


Jonas-Isr (approved):
I approve of this PR

PR #1160, opened 2026-04-27
fix: [OpenAPI] Pin file upload parameter to `File` type

rpanackal:
## Context
 

- https://github.com/SAP/ai-sdk-java-backlog/issues/391

On file uploads, the remote service may expect (here Prompt Registry) `Content-Disposition` header to contain the `filename`. At the moment, setting `<typeMappings>File=byte[]<typeMappings>` can replace all occurrence of `File` type with `byte[]`.

This means on upload operations, we end up losing filename information, critical for correct successful response.

### Feature scope:

 
- [x] Pin the parameter type to `File` in upload operation 
  - [x] Unmodifiable with `<typeMappings>File=byte[]<typeMappings>` override. Now only influence return type on download operations.

**This is a breaking change.**

## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -4,19 +4,20 @@
 
 ### 🚧 Known Issues
 
-- 
+-
 
 ### 🔧 Compatibility Notes
 
-- 
+- [OpenAPI] In Apache HttpClient generator, file upload parameters are now pinned to type `java.io.File`, unmodifiable with `<typeMappings>` in generator configuration.
 
 ### ✨ New Functionality
 
-- 
+-
 
 ### 📈 Improvements
 
-- Enable Retry behavior for HttpClient5 instances, to mirror legacy behvior for HttpClient4: `1` retry with `1s` delay for response codes `429` (Too Many Requests) and `503` (Service Unavailable).
+- Enable Retry behavior for HttpClient5 instances, to mirror legacy behvior for HttpClient4: `1` retry with `1s` delay
+  for response codes `429` (Too Many Requests) and `503` (Service Unavailable).
 
 ### 🐛 Fixed Issues
 
--- a/datamodel/openapi/openapi-api-apache-sample/src/main/java/com/sap/cloud/sdk/datamodel/openapi/apache/sodastore/api/SodasApi.java
+++ b/datamodel/openapi/openapi-api-apache-sample/src/main/java/com/sap/cloud/sdk/datamodel/openapi/apache/sodastore/api/SodasApi.java
@@ -4,6 +4,7 @@
 
 package com.sap.cloud.sdk.datamodel.openapi.apache.sodastore.api;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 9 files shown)


rpanackal (commented):
~Here is a viable alternate proposal due to the existence of `<typeMappings>File=byte[]<typeMappings>`~


~**Current:** We pin upload type and allows `typeMappings` to choose download type~
 
~**Alternate:** We pin download type to `byte[]` and allow `typeMappings` to override upload type~

~I don't see a particular winner. So, any ideas are welcomes~

rpanackal (commented):
**Why not simply keep upload parameter and download return type as `File`?**

From earlier discussions, we want to avoid placing the expectation on user to handle downloaded files properly.
[Related Comment](https://github.com/SAP/ai-sdk-java/pull/660/changes#r2549235996)

rpanackal:
> Here is a viable alternate proposal due to the existence of `<typeMappings>File=byte[]<typeMappings>`
> 
> **Current:** We pin upload type and allows `typeMappings` to choose download type
> 
> **Alternate:** We pin download type to `byte[]` and allow `typeMappings` to override upload type
> 
> I don't see a particular winner. So, any ideas are welcomes

Nevermind. Pinning return type to `byte[]` in the alternate design is not feasible. There is no available `isFile` or `isReturnFile` flag on define a conditional logic in mustache file.

release_notes.md
```diff
@@ -4,19 +4,20 @@
 
 ### 🚧 Known Issues
 
-- 
+-
 
 ### 🔧 Compatibility Notes
 
-- 
+- [OpenAPI] In Apache HttpClient generator, file upload parameters are now pinned to type `java.io.File`, unmodifiable with `<typeMappings>` in generator configuration.
```
newtork, line 11: (Major/Question)

Are you saying this will no longer have an effect?

(`openapi-api-apache-sample/pom.xml`)
```xml
<typeMappings>File=byte[]</typeMappings>
```

Sounds like a regression to me.

rpanackal: Setting `<typeMappings>` is effective for download operations to change return type from `File` to `byte[]`. But not in upload operation.

The default behaviour has always been (is still) `File` type on method parameter and return type. In prompt registry, we overrid this behaviour using `<typeMappings> `. So, for our regular user, nothing changes. Still, advanced users are affected.

If that is not acceptable, we will have to go with introducing a new feature flag.

newtork: is type-mappings for... only?
- (a) return types
- (b) pojo field value types

---

In case we had existing users, It's unfortunate the ones with `byte[]` would now require a creation of temporary file, handle I/O exceptions during OpenAPI execution and delete the file (full life-cycle).

newtork: Interesting, the legacy generator would generate `byte[]` API-arguments too.
But the code compilation would fail like:

image


```
no suitable constructor found for FileSystemResource(byte[])
```

newtork (approved):
I'm a little sad that `byte[]` would not work.

But considering this is (likely?) not a regression for legacy users, I think we can take the hit.

release_notes.md
rpanackal: > is type-mappings for... only?
> 
> (a) return types
> (b) pojo field value types

`typeMappings` is for replacing all occurrence of a some standard types in mustache

> In case we had existing users, It's unfortunate the ones with byte[] would now require a creation of temporary file, handle I/O exceptions during OpenAPI execution and delete the file (full life-cycle).

So, without a file, there is no filename. Unless make one up.

PR #1169, opened 2026-05-08
feat: Add ODataClient for Apache HttpClient5

Jonas-Isr:
## Context
SAP/cloud-sdk-java-backlog#504

This PR adds a new module to the Cloud SDK: `datamodel/odata-client-apache-httpclient5`. This aims to enable users to use Apaches HttpClient v5 in conjunction with our OData functionalities. I tried to stick as closely to the v4 implementation as possible while respecting the design choices of the new HttpClient v5.

This PR does **not** consider the high-level OData v2 and v4 modules. This will be handled in the future.

### Comments regarding review

The first commit of this branch only features the copying of the v4 version of the module. If you want to see the changes that are made compared to our v4 version, look at the changes between the current state of the branch compared to the state after the first commit.

Currently, nothing is marked beta. It might be reasonable to mark (parts of) the new code as such. I am happy to discuss this in the review.

I am also happy to do pair reviews.

### Feature scope:
 
- [x] Add new module for HttpClient v5 (by copying over the old v4 module and adapting it to v5)
- [x] Tests

#### CSRF Token Handling

HC4: `ODataRequestGeneric.tryExecuteWithCsrfToken()` issued a HEAD request to the service path before every mutating request, using DefaultCsrfTokenRetriever.

HC5: Moved to `CsrfTokenInterceptor`, an `HttpRequestInterceptor` registered in `DefaultApacheHttpClient5Factory`. It intercepts mutating requests at the HTTP client level before they are sent. Thus, all calls with `ODataRequestGeneric.tryExecute()` use it. `tryExecuteWithCsrfToken()` therefor does not exist. (This is a breaking change for users that update to HC5. There is of course also the option to keep the method and simply forward this to `tryExecute` but I felt this is even more confusing since then it looks like the two methods should do different things. Happy to discuss this.)

Why: HC5's HttpClient interface no longer exposes an easy way to issue a pre-request side-effect at the OData layer. The interceptor mechanism is the idiomatic HC5 way to inject cross-cutting behavior. It also means CSRF handling applies to all HTTP clients built by the factory, not just OData requests. If this is not wanted, we can add a flag to the builder of the `DefaultApacheHttpClient5Factory` to make this opt-in/opt-out.

Notes: 
 - The interceptor sends HEAD to the service root (derived by stripping the last path segment) rather than the full resource URI, matching HC4's behaviour of scoping the token to the service path.
 - Compared to the HC4 version, if no CSRF token header is returned, there is no `CsrfTokenRetrievalException` thrown. Instead, the problem is only logged (at WARN level). This is a necessary evil since we use HC5's interceptor pattern for CSRF tokens and throwing the exception there is not possible.

#### HTTP Execution Method

HC4: `ODataHttpRequest.requestResource()` uses `HttpClient.execute(httpRequest)`. This [keeps the connection open](https://hc.apache.org/httpcomponents-client-4.5.x/quickstart.html).

HC5: `ODataHttpRequest.requestResource()` uses `HttpClient.executeOpen(null, httpRequest, null)`, which also keeps the connection open. It is explicitly closed in `ODataRequestResultFactory.WITH_BUFFER`.

Why: The HC5 `execute(request, responseHandler)` pattern closes the response after the handler returns. Since the OData layer needs to keep the response open for cases using `ODataRequestResultFactory.WITHOUT_BUFFER` (e.g. for streaming/deserialization), `executeOpen` is the only way to do this in HC5. **This is not a change compared to HC4 since `execute()` in HC4 also did not close the response.** So even though the method name changes from execute to executeOpen, the behaviour is the same.
                                                            
#### Connection Timeout Exception

HC4 uses `ConnectionPoolTimeoutException` from org.apache.http.conn.

HC5: uses `ConnectionRequestTimeoutException` from org.apache.hc.core5.http.

Why: HC5 renamed and restructured exception types. The exception is caught in `ODataHttpRequest` to produce a descriptive `ODataConnectionException` about connection leaks.
                                                            
#### Content-Type on JSON Entities

HC4: `ContentType.APPLICATION_JSON` had no charset and thus `getContentType()` returned "application/json".

HC5: `ContentType.APPLICATION_JSON` includes `charset=UTF-8` and this `getContentType()` would return "application/json; charset=UTF-8".

Fix in `ODataRequestUpdate`: `ComparableHttpEntity` now uses `ContentType.create("application/json")` (no charset). `ODataRequestCreate` sets "application/json" as a literal string. Both avoid the unwanted charset suffix that would appear in batch request bodies.
                                                            
#### UriRequestMerger interface

HC4: `removeDuplicateQueryParameters` was an instance method in `ODataRequestResultGeneric` that cast `httpClient` to `UriQueryMerger`. `UriQueryMerger` was implemented by `HttpClientWrapper` and exposed the destination-contributed query parameters (base URL params + URL.queries... properties) via `mergeRequestUri(URI.create(""))`.

HC5: i tried to do it in the same way. I added `UriQueryMerger` as a public interface in the `connectivity-apache-httpclient5` module, implemented by `ApacheHttpClient5Wrapper`. `ODataRequestResultGeneric` casts `httpClient` to `UriQueryMerger` identically to HC4, strips destination-contributed params that appear verbatim in the next link, and cleans up the resulting query string.

## Definition of Done

  - note that there are no e2e tests (yet)
- [ ] Documentation updated
  - I will consider documentation after the initial code reviews 
- [x] Release notes updated


datamodel/odata-client-apache-httpclient5/src/main/java/com/sap/cloud/sdk/datamodel/odata/client/ODataResponseDeserializer.java
```diff
@@ -0,0 +1,146 @@
...
+        for( int i = 0; i < nodes.size(); i++ ) {
+            if( resultElement == null || !resultElement.isJsonObject() ) {
+                log.warn("JSON path {} could not be resolved for {} at position {}.", path, resultElement, i);
+                return null;
+            }
+            if( nodes.get(i).equals("*") ) {
+                if( !resultElement.getAsJsonObject().entrySet().isEmpty() ) {
+                    resultElement = resultElement.getAsJsonObject().entrySet().iterator().next().getValue();
+                } else {
+                    log.warn("Wildcard in JSON path {}  did not match anything for {}.", path, resultElement);
```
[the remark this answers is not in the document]
Jonas-Isr: Implemented

Jonas-Isr:
Changes after first pair review with Charles:
- cleaned up description of the _HTTP Execution Method_ part in the PR comment above
  - using `executeOpen()` in HC5 is mostly equivalent to `execute()` in HC4, see more details above
- introduced additional changes and tests from [PR 1138](https://github.com/SAP/cloud-sdk-java/pull/1138) that were still missing
- smaller fixes 
  - `assertThatCode` --> `assertThatThrownBy`
  - fixing some outcommented parts of tests
  - consistently not setting HttpVersion 1.1
- checked that no HC4 modules are used for the new module
  - `mvn dependency:tree -pl datamodel/odata-client-apache-httpclient5 '-Dincludes=org.apache.httpcomponents:*'` gives out nothing

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
```diff
@@ -0,0 +1,120 @@
...
+        void
+        process( @Nonnull final HttpRequest request, final EntityDetails entityDetails, final HttpContext context )
+            throws HttpException,
+                IOException
+    {
+        if( !MUTATING_METHODS.contains(request.getMethod().toUpperCase()) ) {
+            return;
+        }
+
+        if( request.containsHeader(X_CSRF_TOKEN_HEADER_KEY) ) {
```
rpanackal, line 46: (question)

Does this need to be a case insensitive check?

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
```diff
@@ -0,0 +1,120 @@
...
+            requestUri = request.getUri();
+        }
+        catch( final Exception e ) {
+            log.debug("Failed to determine request URI for CSRF token fetch, skipping.", e);
+            return;
+        }
+
+        final URI csrfFetchUri = deriveServiceRootUri(requestUri);
+        final HttpHead headRequest = new HttpHead(csrfFetchUri);
+        headRequest.addHeader(X_CSRF_TOKEN_HEADER_KEY, X_CSRF_TOKEN_FETCH_VALUE);
```
rpanackal, line 62: (major)

Don't we need to load other headers from incoming request eg: `Authorization` etc?

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
```diff
@@ -0,0 +1,120 @@
...
+     * segment. Example: {@code http://host/service/$batch} → {@code http://host/service/},
+     * {@code http://host/service/Entity} → {@code http://host/service/}
+     */
+    @Nonnull
+    static URI deriveServiceRootUri( @Nonnull final URI requestUri )
+    {
+        final String path = requestUri.getRawPath();
+        // Service root is everything up to and including the trailing slash before the first resource segment.
+        // Find the last '/' that is followed by at least one more character (i.e., there is a resource segment).
+        final int lastSlash = path.lastIndexOf('/');
```
rpanackal: (major/question)

Can the request path be like `/sap/opu/odata/sap/API_BP/BusinessPartner('123')/to_Address` ? If so, we will have a problem in service path resolution.

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+        catch( final Exception e ) {
+            log
+                .warn(
+                    "CSRF token retrieval failed: the HEAD request was not successful. "
+                        + "The subsequent request may fail if a CSRF token is required.",
+                    e);
+        }
+    }
```
rpanackal, line 90: (question)

Why don't we throw? Am I missing something?

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/UriQueryMerger.java
```diff
@@ -0,0 +1,24 @@
+package com.sap.cloud.sdk.cloudplatform.connectivity;
```
rpanackal, line 1: (question)

What was the motivation to keep the same namespace/package ? Why not add something to point out its hc5?

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ApacheHttpClient5Wrapper.java
```diff
@@ -127,4 +129,14 @@ public RequestConfig getConfig()
...
+
+    @Nonnull
+    @Override
+    public URI mergeRequestUri( @Nonnull final URI requestUri )
+    {
+        final UriPathMerger merger = new UriPathMerger();
+        final URI merged = merger.merge(destination.getUri(), requestUri);
+        final String queryString = String.join("&", QueryParamGetter.getQueryParameters(destination));
+        return merger.merge(merged, URI.create("/?" + queryString));
+    }
```
rpanackal, line 138: (minor)

I would like to have a test for this

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/UriQueryMerger.java
Jonas-Isr: The motivation was to make migration from hc4 to hc5 as easy as possible. We do not expect users to use both in parallel.

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
Jonas-Isr: `.containsHeader()` form `org.apache.hc.core5.http.message.HeaderGroup` is already case insensitive. Nothing we need to do.

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
Jonas-Isr: All relevant headers (including Authorization) for the CSRF call are taken from the destination object and added to the call via the `ApacheHttpClient5Wrapper` anyways (see [here](https://github.com/SAP/cloud-sdk-java/blob/9f012fb3d868f81e60c0869889720880033674a3/cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ApacheHttpClient5Wrapper.java#L108)). So we do not need to add any other headers here.

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
Jonas-Isr: This is the downside of the usage of the interceptor pattern for the CSRF call. I noted it already in the PR note above but did not really explain it.

The main problem is that the interceptor runs inside `httpClient.executeOpen()` of `ODataHttpRequest.requestResource()`. There is no clean way (that I found) to throw from the `CsrfTokenInterceptor.process()` method that surfaces through `ODataHttpRequest.requestResource()` in a way that an SDK user can understand what happened. (`CsrfTokenInterceptor.process()` is also only allowed to throw `HttpException` or `IOException` by the `HttpRequestInterceptor` interface.) 

The next-best option I found was to log the error of the CSRF token call and see if the following requests fail or not. If they fail, then the log message shows why. If they don't fail, we are happy anyways.

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ApacheHttpClient5Wrapper.java
Jonas-Isr: Added some unit tests.

cloudplatform/connectivity-apache-httpclient5/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/CsrfTokenInterceptor.java
Jonas-Isr: Adressed

Jonas-Isr:
Comments regarding pair review with @vladimir-a-sap:

**Issue 1:** Looking at `ODataResourcePath.copy`, can we make the class truly immutable?
- We probably could, but I will not tackle this as part of this PR. The main reason for this is that my current implementation stays as close as possible to the HC4 version. There, the class is also not immutable. And while having immutability would certainly be an improvement to the code quality, I do neither want to touch the HC4 code nor want to deviate from the HC4 code where not strictly necessary. This way, handling changes/feature requests/bugs in the future with the 2 parallel modules is easier as if we now start to change things only in the HC5 version. I would rather see this as a possible future improvement (that should probably be tackled for both modules in parallel) that is outside of the scope of this PR. 

**Issue 2:** `ODataHttpRequest:118` Is the exception thrown here actually actually about ConnectionPool timeout? If not, the message is misleading.
- After some investigation I am pretty certain that `ConnectionRequestTimeoutException` is only thrown when waiting for a lease from the connection pool. Wrong address or network issues should as far as I see it be subclasses of `IOException`.

**Issue 3:** In `ODataRequestResultGeneric` and elsewhere we operate on String-based methods to get parameters out of the URL instead of using higher-level API. This seems rather brittle.
- I agree. But as was the case for Issue 1, this is mostly a copy of the way it is done in the HC4 module. That's why I see it out of scope for this PR to change the way of how these URLs are handled.

vladimir-a-sap (approved):
LGTM

cassiofariasmachado (approved):

PR #1185, opened 2026-06-01
fix: [Connectivity] Change (almost) all JKS usage to PKCS12

rpanackal:
## Context
 

- https://github.com/SAP/ai-sdk-java-backlog/issues/394)

Helpful links:
- https://docs.oracle.com/en/java/javase/11/security/java-cryptography-architecture-jca-reference-guide.html#GUID-A7EEDE25-C4C0-4C28-94EA-262858AE9212
- https://csrc.nist.gov/pubs/sp/800/131/a/r2/final

To be FIPS, we need to change out of JKS usages in our SDK. JKS is a format by SUN, which used MD5 + some proprietary cipher (not FIPS approved).

We can therefore switch to PKCS12 is a fips compatible keystore type. `"PKCS12"` works in  standard JDK, BC FIPS JVM (non-approved), Red Hat FIPS OpenJDK and IBM JCEPlusFIPS.

Even better, we dont harcode the type but invoke it as `KeyStore.getDefaultType()` whose value can be overriden on jvm parameters.

Technically, our choice only affects the internal memory representation and therefore shouldn't break anything for our user.

### Feature scope:

 
- [x] Update in KeyStoreReader
- [x] Update in ZTIS 

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/KeyStoreReader.java
```diff
@@ -65,7 +65,7 @@ static Certificate[] loadCertificates( @Nonnull final Reader certReader )
             IOException
     {
         final List<Certificate> certs = new ArrayList<>();
-        final CertificateFactory factory = CertificateFactory.getInstance("X509");
```
rpanackal, line 68: (drive-by)

This is the standard notation to be used for X.509 certificates

CharlesDuboisSAP (commented):
You should e2e test this, and there might already be an e2e test for this.

rpanackal:
> You should e2e test this, and there might already be an e2e test for this.

Successful run on branch
- https://sdk-v2.cpe.c.eu-de-1.cloud.sap/job/end-to-end-tests/job/v5-scp-cf-spring-mtls/1102/

cloudplatform/pom.xml
```diff
@@ -59,6 +60,17 @@
...
+			<id>non-release</id>
+			<activation>
+				<property>
+					<name>!release</name>
+				</property>
+			</activation>
+			<modules>
+				<module>connectivity-fips-sample</module>
+			</modules>
+		</profile>
```
CharlesDuboisSAP, line 72: Isn't this useless if the module is added anyway line 45?

CharlesDuboisSAP (commented):
~LGTM~

CharlesDuboisSAP (commented):
Could you add a JVM argument to run the `connectivity-fips-sample` and make sure it passes the FIPS check

cloudplatform/pom.xml
rpanackal: thanks for catching this

rpanackal:
> Could you add a JVM argument to run the `connectivity-fips-sample` and make sure it passes the FIPS check

I already have it in `cloudplatform/connectivity-fips-sample/pom.xml` line 85. 

Or do you mean something else? If so, please clarify in detail.

CharlesDuboisSAP (approved):

PR #1187, opened 2026-06-02
test: [Connectivity] Fips profile for BC provider

rpanackal:
## Context

- BLI: https://github.com/SAP/ai-sdk-java-backlog/issues/394

Builds on top of the FIPS keystore fix PR. Adds a CI test profile that validates the PKCS12 fix works in a simulated FIPS environment using Bouncy Castle FIPS in `approved_only` mode.

This change adds the `fips-approved` Maven profile and the `FipsProviderTest` regression guard. It does **not** change any production code — all production changes are in the base PR.

### Feature scope:

- [x] Add `fips-approved` Maven profile to root `pom.xml`
  - Sets `-Dorg.bouncycastle.fips.approved_only=true` in test JVM args
  - Filters Surefire to run only `@Tag("fips-approved")` tests
- [x] Add `fips-approved` profile to `connectivity-apache-httpclient4/pom.xml`
  - Excludes non-FIPS Bouncy Castle (`bcprov-jdk18on`, `bcpkix-jdk18on`)
  - Adds BC FIPS artifacts (`bc-fips:2.1.2`, `bcpkix-fips:2.1.9`) as test-scope dependencies
  - Replicates the dependency swap users must perform in their own applications
- [x] Add `FipsProviderTest` regression test
  - Registers `BouncyCastleFipsProvider` in `@BeforeAll`, removes it in `@AfterAll` (self-contained)
  - Asserts `KeyStoreReader.createKeyStore()` produces a `PKCS12` keystore

**Run the test:**
```bash
mvn test -P fips-approved -pl cloudplatform/connectivity-apache-httpclient4 -am


cloudplatform/connectivity-apache-httpclient4/src/test-fips/java/com/sap/cloud/sdk/cloudplatform/connectivity/FipsProviderTest.java
rpanackal, line 1: This file lives outside regular test root to avoid being scanned for test-compile in default profile

PR #1142, opened 2026-04-09
feat: [Connectivity] Improve Certificate Rotation with ZTIS for OAuth2TokenService

MatKuhr:
## Context

#1134

### Feature scope:

Improves `OAuth2Service` to dynamically re-create HTTP clients for ZTIS certificates. This causes destinations loaded via `ServiceBindingDestinationLoader` to function indefinitely (via the object reference chain `OAuth2Service <- OAuth2HeaderProvider <- HttpDestination`).

However, this improvement does NOT address the KeyStore attached directly to `HttpDestination`, which is needed for the IAS App2Service communication pattern.

## Definition of Done

- [x] Functionality scope stated & covered
- [x] Tests cover the scope above
- [x] Error handling created / updated & covered by the tests above
- [x] ~Documentation updated~
- [x] Release notes updated

PR #1202, opened 2026-06-11
feat: [Connectivity] Dynamic Keystore in HttpDestination

MatKuhr:
## Context

Follow-up for #1142

Allows in-place rotation for destination keystore to fully support #1134 

### Feature scope:

- [x] Extend `DefaultHttpDestination` to support `Supplier<Keystore>` 
- [x] Extend `OAuth2ServiceBindingDestinationLoader` and related classes to use this capability 
- [x] Ensure no new public API is introduced
 
## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated


cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -582,6 +582,26 @@ public static class Builder
 
         final List<DestinationHeaderProvider> customHeaderProviders = new ArrayList<>();
 
+        /**
+         * The {@link KeyStore} to be used when communicating over HTTP.
+         */
+        @Nonnull
+        public Builder keyStore( @Nullable final KeyStore keyStore )
```
MatKuhr, line 589: no new public API, just replaces the previous public `@Setter`

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -52,8 +53,8 @@ public final class DefaultHttpDestination implements HttpDestination
     @Delegate
     private final DestinationProperties baseProperties;
 
-    @Nullable
-    private final KeyStore keyStore;
+    @Nonnull
+    private final Supplier<Option<KeyStore>> keyStore;
```
MatKuhr: `Supplier<Option<KeyStore>>` looks a bit strange, but this is because allowing a supplier to return `null` is often dangerous (especially, because our current annotations don't allow declaring `Supplier<@Nullable KeyStore>`). This solution was the least un-intuitive variant I could find that allows for the KeyStore to be null but still using a supplier.

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -568,11 +571,8 @@ public static class Builder
         private DefaultHttpDestinationBuilderProxyHandler proxyHandler =
             new DefaultHttpDestinationBuilderProxyHandler();
 
-        /**
-         * The {@link KeyStore} to be used when communicating over HTTP.
-         */
-        @Setter( onParam_ = @Nullable )
-        KeyStore keyStore = null;
+        @Nonnull
+        Supplier<Option<KeyStore>> keystoreSupplier = Option::none;
```
Jonas-Isr: Very minor, but inconsistent ;) 

```suggestion
        Supplier<Option<KeyStore>> keyStoreSupplier = Option::none;
```

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -582,6 +582,26 @@ public static class Builder
...
+        @Nonnull
+        public Builder keyStore( @Nullable final KeyStore keyStore )
+        {
+            this.keystoreSupplier = () -> Option.of(keyStore);
+            return this;
+        }
+
+        /**
+         * A {@link Supplier<KeyStore>} to allow for dynamically resolve certificates at runtime for HTTP communication.
+         */
```
Jonas-Isr, line 598: (Grammar)


```suggestion
        /**
         * A {@link Supplier<KeyStore>} to allow for dynamically resolving certificates at runtime for HTTP communication.
         */
```

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
```diff
@@ -549,7 +552,7 @@ public int hashCode()
         return new HashCodeBuilder(17, 37)
             .append(baseProperties)
             .append(customHeaders)
-            .append(resolveKeyStoreHashCode(keyStore))
+            .append(resolveKeyStoreHashCode(keyStoreSupplier.get().getOrNull()))
```
Jonas-Isr, line 555: (Question)

How often do we expect the certificate to rotate (and thus the keyStore to differ)? Could this become a problem with hashCode() returning a different value then, resulting in cache misses or similar?

Jonas-Isr (commented):
Only minor issues/questions :)

cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/DefaultHttpDestination.java
MatKuhr: For now, this should only apply for ZTIS, since the new API is internal and only used there. Certs are expected to change rotate between 12hrs or 3.5 days, so not often.

Indeed, a rotated certificate would cause a cache miss if it were cached, but currently we don't have this for the ZTIS use case. But even if we did, I think that would be okay, this would have been the old behavior / expectation anyway 🙂

Jonas-Isr (approved):

PR #1231, opened 2026-07-21
feat: Support OpenAPI 3.1.x

ZhongpinWang:
## Context

This is a generated (later manually reviewed and fixed) PoC for evaluating the effort of supporting OpenAPI 3.1.x in the generator.

The following two files coming from the initial round of generation might be outdated but you can use as a reference.
- Check the OAS31_GAPS.md for the gaps evaluated.
- Check the OAS31_CHANGES.md for the actual changes generated.

I generated all data models in AI SDK Java with the new SNAPSHOT cloud sdk generator, and there were no breaking changes.

**See https://github.com/SAP/ai-sdk-java/pull/958 for the generated result.**


datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/GenerationConfigurationConverter.java
```diff
@@ -115,6 +115,13 @@ private static void setGlobalSettings( @Nonnull final GenerationConfiguration co
             log.warn("Parsing the specification yielded the following messages: {}", spec.getMessages());
         }
         final var result = spec.getOpenAPI();
+        if( OasVersionUtil.isOas31(result) ) {
+            log
+                .info(
+                    "Detected OAS 3.1 specification (version: {}). "
+                        + "OAS 3.1 support is available with known limitations documented in OAS31_GAPS.md.",
```
CharlesDuboisSAP:
```suggestion
                        + "OAS 3.1 support is available with known limitations documented in https://github.com/SAP/cloud-sdk-java/blob/main/datamodel/openapi/OAS31_GAPS.md",
```

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
```diff
@@ -55,11 +55,14 @@ public void preprocessOpenAPI( @Nonnull final OpenAPI openAPI )
...
         if( USE_EXCLUDE_PATHS.isEnabled(config) ) {
             final String[] exclusions = USE_EXCLUDE_PATHS.getValue(config).trim().split("[,\\s]+");
-            for( final String exclusion : exclusions ) {
-                if( !openAPI.getPaths().keySet().remove(exclusion) ) {
-                    log.error("Could not remove path {}", exclusion);
+            if( openAPI.getPaths() != null ) {
+                for( final String exclusion : exclusions ) {
+                    if( !openAPI.getPaths().keySet().remove(exclusion) ) {
+                        log.error("Could not remove path {}", exclusion);
+                    }
```
CharlesDuboisSAP:
```suggestion
                    openAPI.getPaths().remove(exclusion);
```
I don't see the remove operation failing
Could you also clean up the other warnings in this class?

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
```diff
@@ -200,6 +203,12 @@ private void preprocessRemoveRedundancies( @Nonnull final OpenAPI openAPI )
         final var refs = new LinkedHashSet<String>();
         final var pattern = Pattern.compile("\\$ref: #/components/schemas/(\\w+)");
 
+        // Gap 5: OAS 3.1 documents may have no paths (webhooks-only or components-only).
+        if( openAPI.getPaths() == null || openAPI.getPaths().isEmpty() ) {
+            log.warn("No paths found in OpenAPI spec; skipping unused-component removal.");
+            return;
```
CharlesDuboisSAP, line 260: (Minor) Why not do it line 73 like the remove line 64?

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
```diff
@@ -211,6 +220,20 @@ private void preprocessRemoveRedundancies( @Nonnull final OpenAPI openAPI )
...
+        // Gap 6: OAS 3.1 adds components/pathItems — traverse them for schema references too
+        final var pathItems = openAPI.getComponents() != null ? openAPI.getComponents().getPathItems() : null;
+        if( pathItems != null ) {
+            for( final var pathItem : pathItems.values() ) {
+                final var m = pattern.matcher(pathItem.toString());
+                while( m.find() ) {
+                    final var name = m.group(1);
+                    final var schema = openAPI.getComponents().getSchemas().get(name);
+                    queue.add(schema);
+                    refs.add(m.group(0).split(" ")[1]);
```
CharlesDuboisSAP, line 283: Never tested, don't know if it works and what it does

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomOpenAPINormalizer.java
```diff
@@ -87,4 +108,41 @@ protected void normalizeReferenceSchema( final @Nonnull Schema schema )
...
+        // so that downstream type-mapping (File -> byte[]) continues to work.
+        if( schema.getFormat() == null ) {
+            if( "base64".equalsIgnoreCase(schema.getContentEncoding()) ) {
+                schema.setFormat("byte");
+            } else if( schema.getContentEncoding() != null ) {
+                // Any other content encoding (e.g., "binary") → treat as binary
+                schema.setFormat("binary");
+            } else if( schema.getContentMediaType() != null ) {
+                // contentMediaType without contentEncoding → binary stream
+                schema.setFormat("binary");
```
CharlesDuboisSAP, line 142: Very little is tested here

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/OasVersionUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+    static boolean isOas31( @Nonnull final OpenAPI openAPI )
+    {
+        return isOas31(openAPI.getOpenapi());
+    }
+
+    static boolean isOas31( @Nonnull final JsonNode rootNode )
+    {
+        return isOas31(rootNode.path("openapi").asText(null));
+    }
```
CharlesDuboisSAP, line 17: 2 functions are never used: `version` and `rootNode`

datamodel/openapi/openapi-generator/src/test/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGeneratorUnitTest.java
```diff
@@ -389,4 +395,44 @@ void testConfigOptionsArePassedToGenerator()
...
+                .outputDirectory(outputDirectory.toAbsolutePath().toString())
+                .verbose(false)
+                .apiMaturity(ApiMaturity.RELEASED)
+                .build();
+
+        final Try<GenerationResult> generationResult = new DataModelGenerator().generateDataModel(configuration);
+
+        assertThat(generationResult.isSuccess()).isTrue();
+        assertThat(generationResult.get().getGeneratedFiles()).isNotEmpty();
+    }
```
CharlesDuboisSAP: We usually test specific use cases in the `DataModelGeneratorIntegrationTest` to make sure the generated files are consistant

CharlesDuboisSAP (commented):
Some test coverage is lacking, you could add some more unit tests

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/GenerationConfigurationConverter.java
ZhongpinWang: The md will be removed anyway together with all `Gap x: ...` comments. Only for reviewer to better understand.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
ZhongpinWang: Not a must I agree. But it tells user that spec has exclusion defined which does not exist. Potentially a typo or something. No strong opinion on this and I removed it as you want :)

Also I removed the other one `log.warn("No paths found in OpenAPI spec; skipping unused-component removal.");`.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/GenerationConfigurationConverter.java
ZhongpinWang: I now only log the OAS spec version.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
ZhongpinWang: I wouldn't move the paths check outside of the function.
The function should not have the assumption that the given `openAPI` MUST contain paths. Caller does not need to know about this and the function should not fail due to NPE if paths is null.

ZhongpinWang: If we want to wrap line 64 into a function, I would also move the null check inside the function.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/OasVersionUtil.java
ZhongpinWang: Nice catch! Claude sometimes tries too hard to be perfect..

datamodel/openapi/openapi-generator/src/test/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGeneratorUnitTest.java
ZhongpinWang: Done. Thanks for explaining this.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
ZhongpinWang: I added a unit test for this. So basically, when removing unused schema, we should now also consider the new `pathItems`. If the schema is referenced in `pathItems` they should also not be removed.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomOpenAPINormalizer.java
ZhongpinWang: True. Added some unit tests.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/OasVersionUtil.java
```diff
@@ -13,18 +10,9 @@ private OasVersionUtil()
...
     static boolean isOas31( @Nonnull final OpenAPI openAPI )
     {
-        return isOas31(openAPI.getOpenapi());
-    }
-
-    static boolean isOas31( @Nonnull final JsonNode rootNode )
-    {
-        return isOas31(rootNode.path("openapi").asText(null));
+        final String version = openAPI.getOpenapi();
+        return version != null && version.startsWith("3.1");
```
CharlesDuboisSAP, line 16: Shouldn’t it be version >= 3.1 ?

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-nullable-ref/output/com/sap/cloud/sdk/datamodel/rest/test/model/SodaCategory.java
```diff
@@ -0,0 +1,173 @@
...
+  @JsonAnyGetter
+  private final Map<String, Object> cloudSdkCustomFields = new LinkedHashMap<>();
+
+  /**
+   * Set the name of this {@link SodaCategory} instance and return the same instance.
+   *
+   * @param name  The name of this {@link SodaCategory}
+   * @return The same instance of this {@link SodaCategory} class
+   */
+  @Nonnull public SodaCategory name( @Nullable final String name) {
```
CharlesDuboisSAP: Is SodaCategory.name expected too be nullable? Soda.category was already nullable

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-ref-with-sibling/output/com/sap/cloud/sdk/datamodel/rest/test/model/Soda.java
```diff
@@ -0,0 +1,208 @@
...
+   */
+  public void setName( @Nonnull final String name) {
+    this.name = name;
+  }
+
+  /**
+   * Set the description of this {@link Soda} instance and return the same instance.
+   *
+   * @param description  The description of this {@link Soda}
+   * @return The same instance of this {@link Soda} class
```
CharlesDuboisSAP: Looks like the description did not make it in the javadoc

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorIntegrationTest/oas31-nullable-ref/output/com/sap/cloud/sdk/datamodel/rest/test/model/SodaCategory.java
```diff
@@ -0,0 +1,173 @@
...
+  @JsonAnyGetter
+  private final Map<String, Object> cloudSdkCustomFields = new LinkedHashMap<>();
+
+  /**
+   * Set the name of this {@link SodaCategory} instance and return the same instance.
+   *
+   * @param name  The name of this {@link SodaCategory}
+   * @return The same instance of this {@link SodaCategory} class
+   */
+  @Nonnull public SodaCategory name( @Nullable final String name) {
```
CharlesDuboisSAP, line 57: Again not sure why this should be nullable

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorIntegrationTest/oas31-ref-with-sibling/output/com/sap/cloud/sdk/datamodel/rest/test/model/Soda.java
```diff
@@ -0,0 +1,208 @@
...
+  @Nonnull public Soda description( @Nullable final String description) {
+    this.description = description;
+    return this;
+  }
+
+  /**
+   * Get description
+   * @return description  The description of this {@link Soda} instance.
+   */
+  @Nonnull
```
CharlesDuboisSAP, line 98: Description did not make it again

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-components-path-items/input/sodastore.yaml
```diff
@@ -0,0 +1,49 @@
...
+    sodaItem:
+      get:
+        operationId: getSodaDetail
+        responses:
+          '200':
+            description: success
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/SodaDetail'
```
CharlesDuboisSAP, line 30: (Optional) why unit test only

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/OasVersionUtil.java
ZhongpinWang: It is hard to say >= 3.1 as they don't follow SemVer any more since 3.1. So there can be breaking changes in 3.2 and it should have an additional `isOas32` to control the process logic.

So what is true in 3.1 may not be true in 3.2. As for now, I would suggest we just announce support for 3.1 and say 3.2 not supported.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-ref-with-sibling/output/com/sap/cloud/sdk/datamodel/rest/test/model/Soda.java
ZhongpinWang: Yeah this was a quite complicated bug. Now AI fixed it.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorIntegrationTest/oas31-ref-with-sibling/output/com/sap/cloud/sdk/datamodel/rest/test/model/Soda.java
ZhongpinWang: Same as above, fixed.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-nullable-ref/output/com/sap/cloud/sdk/datamodel/rest/test/model/SodaCategory.java
ZhongpinWang: I personally prefer to make the property actually Nullable everywhere as it is a non-required property, which in Java can be represented as null. This behaviour aligns on the official openapi generator.

Also regenerating the AI SDK with this change introduced zero effort to add null check as these non-required properties were never called directly in our code. 

And if there were never NPE before from the customers reported, then it means not many people ever used the Java SDK generator and get / set non-required properties. So it won't be breaking for them as well.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorIntegrationTest/oas31-nullable-ref/output/com/sap/cloud/sdk/datamodel/rest/test/model/SodaCategory.java
ZhongpinWang: Same as above, I will resolve this one and track the discussion on the previous one.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-components-path-items/input/sodastore.yaml
ZhongpinWang: Make sense. Done.

datamodel/openapi/openapi-generator/src/test/resources/DataModelGeneratorApacheIntegrationTest/oas31-nullable-ref/output/com/sap/cloud/sdk/datamodel/rest/test/model/SodaCategory.java
ZhongpinWang: As wrote in the slack channel, I left the @Nonnull change for getter and fluent api return value as-is. `@Nullable` for the parameter is the current behaviour and not touched by this PR.

CharlesDuboisSAP (commented):
Overall LGTM, one small thing:
Why did you duplicate all integrations tests in `DataModelGeneratorApacheIntegrationTest`? I would remove them since they are already tested in `DataModelGeneratorIntegrationTest`, and they cover the same lines of code, apache only changes the inside of the Api class which are not the point of your tests anyway.

ZhongpinWang:
@CharlesDuboisSAP @Jonas-Isr Can you add a simple release note for this? I don't know how you do this normally in Java team.

ZhongpinWang:
I added a log warning for 3.1 to point out that this is an experimental feature. We can remove this after a while.

datamodel/openapi/openapi-generator/pom.xml.bak
```diff
@@ -0,0 +1,194 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>com.sap.cloud.sdk.datamodel</groupId>
+		<artifactId>openapi-parent</artifactId>
```
CharlesDuboisSAP: Please remove this `.bak` file

datamodel/openapi/openapi-generator/pom.xml
```diff
@@ -1,194 +1,194 @@
...
-	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>com.sap.cloud.sdk.datamodel</groupId>
-		<artifactId>openapi-parent</artifactId>
-		<version>5.33.0-SNAPSHOT</version>
-	</parent>
-	<artifactId>openapi-generator</artifactId>
-	<packaging>jar</packaging>
-	<name>Data Model - OpenAPI Services - Generator</name>
-	<description>OpenAPI Services data model (VDM) - generator classes</description>
```
CharlesDuboisSAP, line 12: Formatting?

ZhongpinWang: I guess my local formatting plugin does not have the same configuration as you people. Looks like there is no actual changes needed so I reverted it. Thanks for pointing it out.

datamodel/openapi/openapi-generator/pom.xml.bak
ZhongpinWang: I am tired with AI... Sorry for mistakenly adding this file...

CharlesDuboisSAP (commented):
LGTM, let's wait for @Jonas-Isr's review

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
```diff
@@ -38,6 +39,8 @@ class CustomJavaClientCodegen extends JavaClientCodegen
 {
     private final GenerationConfiguration config;
     private static final Predicate<String> DOUBLE_IS_PATTERN = Pattern.compile("^isIs[A-Z]").asPredicate();
+    // schemaName -> (propertyName -> sibling description) captured before normalization strips $ref context
+    private final Map<String, Map<String, String>> siblingDescriptions = new java.util.HashMap<>();
```
Jonas-Isr: (Very Minor)
 
Import once. Is also used below (l.188).

```suggestion
    private final Map<String, Map<String, String>> siblingDescriptions = new HashMap<>();
```

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
```diff
@@ -161,7 +216,7 @@ private void preprocessRemoveProperty(
         boolean removed = false;
 
         final Predicate<Schema> remove =
-            s -> s != null && s.getProperties() != null && s.getProperties().remove(propertyName) != null;
+            s -> s.getProperties() != null && s.getProperties().remove(propertyName) != null;
```
Jonas-Isr, line 219: (Question)

Why is it save to remove the null check here?

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomOpenAPINormalizer.java
```diff
@@ -46,15 +52,28 @@ protected void normalizeReferenceSchema( final @Nonnull Schema schema )
...
 
+        // warn when deprecated nullable: true is used in an OAS 3.1 spec
+        if( isOas31 && schema.getNullable() != null ) {
+            LOGGER
+                .warn(
+                    "'nullable: true' is not a valid OAS 3.1 keyword on $ref schema '{}'. "
+                        + "Use anyOf: [{{$ref: \"...\"}}, {{type: \"null\"}}] instead.",
+                    schema.get$ref());
+        }
+
```
Jonas-Isr, line 63: (Minor)

Maybe we should test this path as well?

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
```diff
@@ -137,6 +144,54 @@ protected void updateModelForComposedSchema(
         }
     }
 
+    @SuppressWarnings( { "rawtypes", "RedundantSuppression" } )
```
Jonas-Isr: (Question)

Is the SurpressWarnings actually necessary here? It looks to me like this method would be fine without it.

```suggestion
```

Jonas-Isr (commented):
Looks good to me :)

I only had some pretty minor points.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
ZhongpinWang: Good spot. Looks like AI generated code to me lol

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
ZhongpinWang: @CharlesDuboisSAP This actually comes from your commit in this PR.

But I suppose, it is because in line 230 where this predicate is actually used, there is already a null check before in line 226. So the value passed down to this is never null

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
ZhongpinWang: I removed it. Shouldn't be a problem.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomOpenAPINormalizer.java
ZhongpinWang: You are right. It was not added because testing logger is quite cumbersome... I anyway added two tests to check the warning log, by using reflection to change the default log level of slf4j-simple temporarily at runtime. If you have any other more elegant idea, let me know.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomJavaClientCodegen.java
Jonas-Isr: Ah, I missed that before. Thanks! Then it makes sense :)

Jonas-Isr (approved):
Looks good :)

PR #1233, opened 2026-07-23
chore: [DevOps] Handle CVE-2026-55760

Jonas-Isr:
## Context

Blackduck flagged `com.github.jknack:handlebars`, which is transitively used, as vulnerable so this PR fixes its version to a non-vulnerable version and aligns versions for `org.apache.commons:commons-text`.

The new version of `handlebars` introduced a dependency on `org.openjdk.nashorn` which has multiple Blackduck issues. Since `nashorn` is apparently only used by `handlebars` for JavaScript-based template helpers which are not used by the openapi-generator, it can be excluded.

I checked locally that after the exclusion
  1) all generator tests in Cloud SDK are still green,
  2) AI SDK code is generated without any changes.

[Successful Blackduck run after changes](https://github.com/SAP/cloud-sdk-java/actions/runs/30341753006).

### DIFF
--- a/.github/workflows/blackduck.yaml
+++ b/.github/workflows/blackduck.yaml
@@ -16,10 +16,24 @@ jobs:
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v7
-        with:
-          ref: cve-fixing
       - name: "Scan With Black Duck"
         uses: ./.github/actions/scan-with-blackduck
         with:
           token: ${{ secrets.BLACKDUCK_TOKEN }}
 
+  notify-job:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    needs: [ scan ]
+    if: ${{ failure() && github.ref == 'refs/heads/main' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+      - name: Notify
+        run: python .pipeline/scripts/notify.py
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          WORKFLOW: ${{ github.workflow }}
+          WORKFLOW_RUN_URL: https://github.com/SAP/cloud-sdk-java/actions/runs/${{ github.run_id }}
+          BRANCH_NAME: ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -327,6 +329,19 @@
 				<artifactId>commons-beanutils</artifactId>
 				<version>${commons-beanutils.version}</version>
 			</dependency>
+			<!-- resolve vulnerability CVE-2026-55760 (handlebars is used transitively via wiremock) -->
+			<dependency>
+				<groupId>com.github.jknack</groupId>
+				<artifactId>handlebars</artifactId>
+				<version>${handlebars.version}</version>
+				<scope>test</scope>
```
CharlesDuboisSAP: Wrong scope? Or if this is a test dependency it shouldn't be scanned by blackduck

pom.xml
```diff
@@ -787,6 +802,7 @@
 						<ignoredUnusedDeclaredDependency>org.junit.jupiter:junit-jupiter-engine</ignoredUnusedDeclaredDependency>
 						<!-- Transitive dependencies explicitly declared to fic CVEs-->
 						<ignoredUnusedDeclaredDependency>commons-beanutils:commons-beanutils</ignoredUnusedDeclaredDependency>
+						<ignoredUnusedDeclaredDependency>com.github.jknack:handlebars</ignoredUnusedDeclaredDependency>
```
CharlesDuboisSAP: This is a sign that the dependency was wrongly declared

CharlesDuboisSAP (changes requested):

.github/workflows/blackduck.yaml
```diff
@@ -16,24 +16,10 @@ jobs:
...
     steps:
       - uses: actions/checkout@v7
+        with:
+          ref: cve-fixing
       - name: "Scan With Black Duck"
         uses: ./.github/actions/scan-with-blackduck
         with:
           token: ${{ secrets.BLACKDUCK_TOKEN }}
 
-  notify-job:
```
CharlesDuboisSAP, line 24: Please revert

CharlesDuboisSAP (approved):

PR #1234, opened 2026-07-23
fix: [DevOps] Dependabot Automerge

CharlesDuboisSAP:
- `production-major group` couldn't be merged
  - https://github.com/SAP/cloud-sdk-java/pull/1229
- [Successful run](https://github.com/SAP/cloud-sdk-java/actions/runs/30000755922)

### DIFF
--- a/.github/workflows/dependabot-automerge.yaml
+++ b/.github/workflows/dependabot-automerge.yaml
@@ -29,6 +29,10 @@ jobs:
           permission-contents: write
           permission-pull-requests: write
 
+      - name: Checkout
+        uses: actions/checkout@v7
+
+
       - name: 'Approve and Merge PRs'
         run: |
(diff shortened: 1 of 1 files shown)


Jonas-Isr:
But did we not on purpose leave out major-version updates so that a human looks at them before merging?

CharlesDuboisSAP:
> But did we not on purpose leave out major-version updates so that a human looks at them before merging?

That might be true, but what should we do if it passes the tests anyway?

Jonas-Isr:
I mean, in that case it is 1 additional click only, anyways. Which is why I liked checking major version updates just to be sure.

Jonas-Isr:
We discussed the issue offline and came to the conclusion that it makes sense to try this out.

Jonas-Isr (approved):

Issue #1243, opened 2026-08-03
ZeroTrustIdentityService does not configure svidPicker, causing non-deterministic SVID selection


capalini:
### Describe the Bug

# `ZeroTrustIdentityService` does not configure `svidPicker`, causing non-deterministic SVID selection

**Library:** `com.sap.cloud.sdk.cloudplatform:connectivity-ztis`
**Affected version:** `5.32.0` (and likely earlier)
**SPIFFE library:** `io.spiffe:java-spiffe-core:0.8.17`

---

## Problem

`ZeroTrustIdentityService.initX509Source()` builds an `X509SourceOptions` without setting an `svidPicker`:

```java
X509SourceOptions.builder()
    .spiffeSocketPath(socketPath)
    .initTimeout(DEFAULT_SOCKET_TIMEOUT)
    .build();  // no .svidPicker(...) → falls back to getDefaultSvid()
```

`DefaultX509Source.setX509Context()` (in `java-spiffe-core`) then falls back to `getDefaultSvid()`, which simply returns **the first SVID in the list**:

```java
if (picker == null) {
    svidUpdate = update.getDefaultSvid();   // first in list — non-deterministic
} else {
    svidUpdate = picker.apply(update.getX509Svids());
}
```

The SPIRE agent does **not** guarantee ordering. The set of SVIDs served for a workload changes when:

- A new CF service key is created for the ZTIS service instance — the ZTIS service broker registers a new SPIRE workload entry with the same selectors (CF space/app), making it visible to the running sidecar.
- Another service on the same Diego cell has a workload entry with overlapping selectors.

When this happens, `getDefaultSvid()` silently returns a **different** SVID — one belonging to another service key or workload. The mTLS handshake then either fails or presents the wrong identity to the remote service.

---

```java
// Desired behaviour inside ZeroTrustIdentityService
String expectedSpiffeId = binding.getCredentials().get("workload.spiffeID");

X509SourceOptions.builder()
    .spiffeSocketPath(socketPath)
    .initTimeout(DEFAULT_SOCKET_TIMEOUT)
    .svidPicker(svids -> svids.stream()
        .filter(s -> s.getSpiffeId().toString().equals(expectedSpiffeId))
        .findFirst()
        .orElseThrow(() -> new IllegalStateException(
            "No SVID found for SPIFFE ID: " + expectedSpiffeId)))
    .build();
```

---

## Impact

[...]

PR #1246, opened 2026-08-04
fix: [Connectivity-ZTIS] Deterministic SVID selection

rpanackal:
Fixes 
- SAP/cloud-sdk-java#1243.

`ZeroTrustIdentityService` did not configure an `svidPicker` when building `X509SourceOptions`, causing `java-spiffe-core` to fall back to `getDefaultSvid()` which simply returns the first SVID in an unordered list. When the SPIRE agent returns multiple SVIDs to a workload (e.g., after creating a second service key for the same ZTIS instance, or when a co-located workload has overlapping selectors), this resulted in non-deterministic SVID selection that could silently present the wrong identity during mTLS handshakes.

This change configures an `svidPicker` that deterministically selects the SVID whose SPIFFE ID matches the expected ID from the ZTIS service binding (`credentials.workload.spiffeID`), as required by the SPIFFE specification.

### Feature scope:
 
- [x] Configure `svidPicker` in `ZeroTrustIdentityService.initX509Source()` to select SVID by SPIFFE ID
- [x] Add `pickSvid` helper method with proper error handling for no-match scenarios
- [x] Add unit tests covering happy path (single/multiple SVIDs) and error cases
- [x] Update release notes with bug fix entry

## Definition of Done

- [x] ~Documentation updated~
- [x] Release notes updated

### DIFF
--- a/cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
+++ b/cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
@@ -10,6 +10,7 @@ import java.security.cert.CertificateException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Date;
+import java.util.List;
 import java.util.Objects;
 
 import javax.annotation.Nonnull;
@@ -23,6 +24,7 @@ import com.sap.cloud.sdk.cloudplatform.exception.CloudPlatformException;
 
 import io.spiffe.bundle.x509bundle.X509Bundle;
 import io.spiffe.exception.X509SvidException;
+import io.spiffe.spiffeid.SpiffeId;
 import io.spiffe.spiffeid.TrustDomain;
 import io.spiffe.svid.x509svid.X509Svid;
 import io.spiffe.workloadapi.DefaultX509Source;
@@ -109,14 +111,45 @@ public class ZeroTrustIdentityService
         final String socketPath = Option.of(System.getenv(SOCKET_ENVIRONMENT_VARIABLE)).getOrElse(DEFAULT_SOCKET_PATH);
         log.info("Using socket path {} for ZTIS agent.", socketPath);
 
-        final X509SourceOptions x509SourceOptions =
-            X509SourceOptions.builder().spiffeSocketPath(socketPath).initTimeout(DEFAULT_SOCKET_TIMEOUT).build();
+        // The SPIRE agent may return multiple SVIDs when overlapping registration selectors exist
+        // (e.g. a second service key on the same ZTIS instance, or a co-located workload). We must pick
+        // the one whose SPIFFE ID matches our binding — see pickSvid.
+        final SpiffeId expectedSpiffeId = SpiffeId.parse(mapView.getMapView("workload").getString("spiffeID"));
+        final X509SourceOptions options =
+            X509SourceOptions
+                .builder()
+                .spiffeSocketPath(socketPath)
+                .initTimeout(DEFAULT_SOCKET_TIMEOUT)
+                .svidPicker(svids -> pickSvid(svids, expectedSpiffeId))
+                .build();
         try {
-            return DefaultX509Source.newSource(x509SourceOptions);
+            return DefaultX509Source.newSource(options);
         }
         catch( final Exception e ) {
             throw new CloudPlatformException("Failed to load the certificate from the unix socket: " + socketPath, e);
         }
+    }
+
+    /**
+     * Selects the {@link X509Svid} whose SPIFFE ID equals {@code expectedSpiffeId} from the given list. Used as the
+     * {@code svidPicker} for {@link DefaultX509Source} to avoid non-deterministic selection when the SPIRE agent
+     * returns multiple SVIDs (e.g. after a second service key is created for the same ZTIS instance).
+     */
+    @Nonnull
+    static X509Svid pickSvid( @Nonnull final List<X509Svid> svids, @Nonnull final SpiffeId expectedSpiffeId )
+    {
+        log.debug("SPIRE agent returned {} SVID(s); selecting the one matching '{}'.", svids.size(), expectedSpiffeId);
+        return svids
+            .stream()
+            .filter(svid -> expectedSpiffeId.equals(svid.getSpiffeId()))
+            .findFirst()
+            .orElseThrow(
+                () -> new CloudPlatformException(
+                    "No SVID matching SPIFFE ID '"
+                        + expectedSpiffeId
+                        + "' among "
+                        + svids.size()
+                        + " returned by the SPIRE agent."));
     }
 
     @Nonnull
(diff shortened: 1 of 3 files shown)


cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -109,16 +111,47 @@ X509Source initX509Source()
         final String socketPath = Option.of(System.getenv(SOCKET_ENVIRONMENT_VARIABLE)).getOrElse(DEFAULT_SOCKET_PATH);
         log.info("Using socket path {} for ZTIS agent.", socketPath);
 
-        final X509SourceOptions x509SourceOptions =
-            X509SourceOptions.builder().spiffeSocketPath(socketPath).initTimeout(DEFAULT_SOCKET_TIMEOUT).build();
+        // The SPIRE agent may return multiple SVIDs when overlapping registration selectors exist
+        // (e.g. a second service key on the same ZTIS instance, or a co-located workload). We must pick
+        // the one whose SPIFFE ID matches our binding — see pickSvid.
+        final SpiffeId expectedSpiffeId = SpiffeId.parse(mapView.getMapView("workload").getString("spiffeID"));
```
Jonas-Isr: (Minor)

Let's wrap the exception if `SpiffeId.parse()` throws.


```suggestion
        final SpiffeId expectedSpiffeId;
        try {
            expectedSpiffeId = SpiffeId.parse(mapView.getMapView("workload").getString("spiffeID"));
        } catch (Exception e) {
            throw new CloudPlatformException("Invalid SPIFFE ID in Zero Trust Identity Service binding.", e);
        }
```

cloudplatform/connectivity-ztis/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityServiceTest.java
```diff
@@ -152,6 +154,31 @@ void testCertOnFileSystem()
             .isInstanceOf(IllegalStateException.class);
     }
 
+    @Test
+    void testPickSvidSelectsMatchingSvid()
```
Jonas-Isr: (Comment)

It's a bit annoying that it is so hard to test the actual behaviour of the `.initX509Source()` method instead. I would in general prefer having a test that actually fails with the old way of doing it but it is very cumbersome to mock that... So I think the current tests are a fine compromise :)

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -109,16 +111,47 @@ X509Source initX509Source()
...
+    @Nonnull
+    static X509Svid pickSvid( @Nonnull final List<X509Svid> svids, @Nonnull final SpiffeId expectedSpiffeId )
+    {
+        log.debug("SPIRE agent returned {} SVID(s); selecting the one matching '{}'.", svids.size(), expectedSpiffeId);
+        return svids
+            .stream()
+            .filter(svid -> expectedSpiffeId.equals(svid.getSpiffeId()))
+            .findFirst()
+            .orElseThrow(
+                () -> new CloudPlatformException(
```
CharlesDuboisSAP, line 157: Why throw? If not found we could use the old behaviour of not setting a `svidPicker`?

Jonas-Isr (commented):
Only a minor point and a comment. Looks good to me otherwise :)

CharlesDuboisSAP (commented):
None of the test fail when you remove the `svidPicker`

cloudplatform/connectivity-ztis/src/test/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityServiceTest.java
rpanackal: Yeah, exactly. I (and Claude) thought these is better for testing

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
rpanackal: Falling back to the old behavior would reproduce the bug in a smaller scope — it would only behave correctly when there's exactly one SVID, or when the library happens to pick the right one from an unordered list. That's the  non-determinism we're fixing.

CharlesDuboisSAP: It only throws when it doesn’t find the svid from the list right?

rpanackal: Yes

rpanackal: Also, just improved the tests to address the other concern regarding removing the svidPicker

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
```diff
@@ -109,16 +111,57 @@ X509Source initX509Source()
...
     }
 
+    X509SourceOptions
+        buildX509SourceOptions( @Nonnull final String socketPath, @Nonnull final SpiffeId expectedSpiffeId )
+    {
+        return X509SourceOptions
+            .builder()
+            .spiffeSocketPath(socketPath)
+            .initTimeout(DEFAULT_SOCKET_TIMEOUT)
+            .svidPicker(svids -> pickSvid(svids, expectedSpiffeId))
```
CharlesDuboisSAP, line 139: If I remove the `svidPicker`, the tests are still green apart from `assertThat(picker).isNotNull();`
I don't think any test checks the actual change of this PR, the `svidPicker`

CharlesDuboisSAP (changes requested):

cloudplatform/connectivity-ztis/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/ZeroTrustIdentityService.java
Jonas-Isr: I think that is the same issue discussed [here](https://github.com/SAP/cloud-sdk-java/pull/1246#discussion_r3712472097)?

CharlesDuboisSAP: Then wdym it's a compromise, if you revert the production code the tests are still green

Jonas-Isr: They are still green because they test the newly added method directly and not the integration of the new method into the larger process.

I agree that this is not the ideal testing strategy but I also see no better way to test it.

Jonas-Isr: As discussed during the call today, we will let @cassiofariasmachado and team have final decision on this.

PR #1260, opened 2026-08-25
fix: [OpenAPI] String properties are not mistakenly generated as File anymore

CharlesDuboisSAP:
## Context
 

- https://github.com/SAP/ai-sdk-java-backlog/issues/423

A property of type String can be either a File or a strong depending on the content type

### Feature scope:

 
- [x] Content type application/json does not generate into File but rather String

## Definition of Done

- [ ] ~Documentation updated~
- [x] Release notes updated


Jonas-Isr (dismissed):
Looks good to me :)

As discussed, we can investigate this issue a bit further in the near future but this looks like a good fix to the initial problem.

datamodel/openapi/openapi-generator/src/main/java/com/sap/cloud/sdk/datamodel/openapi/generator/CustomOpenAPINormalizer.java
```diff
@@ -137,7 +137,8 @@ public Schema normalizeSchema( final @Nonnull Schema schema, final @Nonnull Set<
             } else if( schema.getContentEncoding() != null ) {
```
vladimir-a-sap, line 137: documenting our discussion here:

- this line may actually be the issue
- current plan is to make this quick fix and then investigate later separately

datamodel/openapi/openapi-generator/src/test/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGeneratorIntegrationTest.java
```diff
@@ -60,6 +60,17 @@ enum TestCase
             6,
             Map.of(),
             Map.of()),
+        DOUBLE_STRING(
```
vladimir-a-sap:
```suggestion
        MULTIPART_BINARY_STRING(
```

datamodel/openapi/openapi-generator/src/test/java/com/sap/cloud/sdk/datamodel/openapi/generator/DataModelGeneratorIntegrationTest.java
```diff
@@ -257,7 +268,7 @@ enum TestCase
     }
 
     @ParameterizedTest
-    @EnumSource( value = TestCase.class, mode = EnumSource.Mode.EXCLUDE, names = { "FILE_HANDLING" } )
+    @EnumSource( value = TestCase.class, mode = EnumSource.Mode.EXCLUDE, names = { "FILE_HANDLING", "DOUBLE_STRING" } )
```
vladimir-a-sap: is new test case exclusion intended?

CharlesDuboisSAP: Excluded from the Spring test to only be tested in Apache, because the test contains File as a type which is handled differently

vladimir-a-sap (approved):
LGTM

PR #1282, opened 2026-09-21
fix: KeyStoreReader fails to decrypt PBES2/PBKDF2-encrypted destinati…

NicoleMGomes:
# Fix PBES2/PBKDF2-encrypted PEM private keys on keystores

## Context

The BTP Destination Service now issues **PBES2/PBKDF2-encrypted** private keys by default for newly generated certificates used with `ClientCertificateAuthentication`. The SDK failed to decrypt these keys at runtime with:

```
PBKDF2with8BIT SecretKeyFactory not available
```

Root cause: `KeyStoreReader` used `JcaPEMKeyConverter` and `JceOpenSSLPKCS8DecryptorProviderBuilder` without pinning a JCE provider. BouncyCastle is a compile-time dependency but was never registered with the JVM's global JCE provider list. The built-in Sun JCE providers do not support PBKDF2withHmacSHA256+AES-256 (PBES2), so decryption failed for any destination with a newly provisioned certificate.

The fix passes a `BouncyCastleProvider` instance explicitly to each BouncyCastle builder, bypassing the global JCE registry without mutating it via `Security.addProvider()`.

### Feature scope:

- [x] Pass `BouncyCastleProvider` explicitly to `JcaPEMKeyConverter` and `JceOpenSSLPKCS8DecryptorProviderBuilder` in `KeyStoreReader.loadKey()`
- [x] Add PBES2/PBKDF2-encrypted PEM test fixtures (`pbes2-client.key`, `pbes2-client.crt`)
- [x] Add regression test `testPbes2EncryptedPem` reproducing the DINC1028726 failure before the fix

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


cassiofariasmachado (approved):
Nice work!

ricardosrib (approved):

